PR #10, opened 2023-05-23
in-memory conversation history implementation

khshah6:


src/server.rs
```diff
@@ -125,6 +143,29 @@ struct SearchRequest {
...
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+struct ConversationHistoryAddRequest {
+    /// The memory policy for storing and retrieving from conversation history.
+    memory_policy: MemoryPolicy,
+    turn: String,
+}
+
+struct ConversationHistoryAddResponse {
```
diptanu, line 153: @khshah6 The derive macro for serde - Serialize and Deserialize needs to be added since we are returning a Json<T> from the axum handler. This adds the Serde related methods into the struct during compilation.

src/server.rs
```diff
@@ -165,6 +208,7 @@ impl Server {
     /// * A result indicating success or failure of the operation.
     pub async fn run(&self) -> Result<()> {
         let embedding_router = Arc::new(EmbeddingRouter::new(self.config.clone())?);
+        let conversation_history_router = Arc::new(ConversationHistoryRouter::new(self.config.clone())?);
```
diptanu, line 211:
```
        let conv_history_router = ConversationHistoryRouter::new(self.config.clone())?;
        let conversation_history_router = Arc::new(Mutex::new(conv_history_router));
```

@khshah6  The router needs to be wrapped into Mutex as well since the handler expects that below.

I would suggest moving Mutex inside the struct so that we don't have to think about locking and such in the handler.

src/server.rs
```diff
@@ -189,6 +233,10 @@ impl Server {
             .route(
                 "/index/search",
                 get(index_search).with_state((index_manager.clone(), embedding_router.clone())),
+            )
+            .route(
+                "/conversation_history/add",
```
diptanu, line 238: What do you think of renaming this to `/memory/add`?

src/memory/lru.rs
```diff
@@ -0,0 +1,109 @@
...
+            } else {
+                self.tail = Some(node.clone());
+            }
+            self.head = Some(node.clone());
+            self.map.insert(node.key, node.clone());
+        }
+    }
+}
+
+fn main() {
```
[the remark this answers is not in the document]
diptanu: Also another option to use a crate like - https://github.com/jeromefroe/lru-rs

src/server_config.rs
```diff
@@ -72,6 +83,18 @@ pub struct EmbeddingModel {
     pub device_kind: DeviceKind,
 }
 
+/// Struct representing the configuration of a conversation history data structure.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub struct MemoryPolicy {
```
diptanu, line 89: MemoryStoragePolicy?

src/server_config.rs
```diff
@@ -48,6 +48,17 @@ pub enum EmbeddingModelKind {
     OpenAIAda02,
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, strum_macros::Display)]
+#[strum(serialize_all = "kebab-case")]
+pub enum MemoryPolicyKind {
```
diptanu, line 53: MemoryStoragePolicyKind ?

src/server.rs
```diff
@@ -323,6 +378,96 @@ async fn add_texts(
...
+    if let Err(err) = conversation_history_lock {
+        return (
+            StatusCode::INTERNAL_SERVER_ERROR,
+            Json(ConversationHistoryAddResponse {
+                errors: vec![err.to_string()],
+            }),
+        );
+    };
+
+    let result = conversation_history_lock.unwrap().add_turn(memory_policy.to_string(), payload.turn);
```
diptanu, line 405: Moving the lock out of here int the ConversationHistory struct would be cleaner.

diptanu (changes requested):
The main comment is related to how we are going to handle concurrent sessions and secondly, MemoryPolicy sounds like really MemoryStoragePolicy and since it wouldn't change for a given session of an agent we could set it when the conversation starts and return a session id for subsequent add and retrieve - 
```
session = indexify.create_session(agent_id, storage_policy, extra_metadata)
indexify.add_memory(session, "hello")
indexify.add_memory(session, "world")
....
memory_events = indexify.retreive(session)
```

This will allow us to add more features to allow developers to ask Indexify to post process memory events while creating a new session.

src/server.rs
khshah6: Do you mean creating a function to retrieve the lock?

src/server_config.rs
```diff
@@ -48,6 +48,17 @@ pub enum EmbeddingModelKind {
...
 
+#[derive(Debug, Clone, Serialize, Deserialize, strum_macros::Display)]
+#[strum(serialize_all = "kebab-case")]
+pub enum MemoryStoragePolicyKind {
+    #[serde(rename = "indefinite")]
+    Indefinite,
+    #[serde(rename = "window")]
+    Window,
+    #[serde(rename = "lru")]
+    Lru,
```
diptanu, line 59: nit: Same as above -
```
Lru{capacity: usize}
```

diptanu (changes requested):
Looking good overall, I think the main thing remaining here - 
Use RWLock or DashMap in the MemoryRouter struct so that locks are acquired and released inside the methods of the add and retrieve methods of the router.

src/server.rs
diptanu: Responded in detail in the review.

src/server.rs
```diff
@@ -125,6 +143,42 @@ struct SearchRequest {
...
+
+#[derive(Serialize, Deserialize)]
+struct MemorySessionAddResponse {
+    errors: Vec<String>,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+struct MemorySessionRetrieveRequest {
+    session_id: Uuid,
+    query: String,
```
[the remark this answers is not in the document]
khshah6: yeah, for LruCache similarity scoring and other future implementations

diptanu (approved):
LGTM

PR #11, opened 2023-05-28
Implement DPR based retrieval

diptanu:
* Inference is moved to Python.


src/vectordbs/qdrant.rs
```diff
@@ -34,10 +34,9 @@ impl QdrantDb {
         }
     }
 
-    async fn create_client(&self) -> Result<QdrantClient, VectorDbError> {
+    fn create_client(&self) -> Result<QdrantClient, VectorDbError> {
         let client_config = QdrantClientConfig::from_url(&self.qdrant_config.addr);
         let client = QdrantClient::new(Some(client_config))
-            .await
             .map_err(|e| VectorDbError::IndexCreationError(e.to_string()))?;
```
khshah6, line 40: should probably be a different error

diptanu: I changed the name, but it's not resolved since this was a comment and not a review.

PR #13, opened 2023-06-01
Add langchain memory impl.

khshah6:


sdk-py/indexify/indexify.py
```diff
@@ -54,6 +65,20 @@ class SearchResult:
     results: List[TextChunk]
 
 
+@dataclass
+class MemoryStoragePolicy:
+    policy_kind: str
+    index_name: Optional[str]
+    db_url: Optional[str]
```
diptanu, line 72: We don't need to pass a db_url here, since the server will be configured with a database URL.

sdk-py/indexify/memory/indexify_entity_store.py
```diff
@@ -0,0 +1,51 @@
+from typing import Any, Optional
+from uuid import UUID
+
+from langchain.memory.entity import BaseEntityStore
+
+from indexify.indexify import Indexify, MemoryResult
+
+
+class IndexifyEntityStore(BaseEntityStore):
```
diptanu, line 9: Looks like we want to implement https://github.com/hwchase17/langchain/blob/master/langchain/schema.py#L218 BaseChatMessageHistory? BaseEntityStore feels like is related to storing more rudimentary k/v in a database. Any thoughts?

sdk-py/indexify/tools/search.py
```diff
@@ -0,0 +1,18 @@
+"""Tool for memory search."""
+from indexify.indexify import Indexify, Metric, TextSplitter
+from pydantic import BaseModel
+
+
+class IndexifySearchTool(BaseModel):
```
diptanu, line 6: I think in addition to exposing Indexify as a tool, we should also implement the traditional retrieval interface, which is very similar - https://github.com/hwchase17/langchain/blob/master/langchain/schema.py#L273

sdk-py/indexify/tools/search.py
```diff
@@ -0,0 +1,18 @@
...
+    """Search memory embeddings for information."""
```
diptanu, line 7: Also, if we are exposing Indexify as a tool we should add some description to what the tool does so that LLMs can discover the tool when it's planning execution sequence for a query.

diptanu: Here is an example - https://github.com/hwchase17/langchain/blob/master/langchain/tools/google_search/tool.py#L17

src/server_config.rs
```diff
@@ -155,9 +164,8 @@ impl Default for ServerConfig {
             }),
             index_config: None,
             memory_policies: vec![MemoryStoragePolicy {
-                policy_kind: MemoryStoragePolicyKind::Indefinite,
-                window_size: None,
-                capacity: None,
+                policy_kind: MemoryPolicyKind::Indefinite,
```
diptanu, line 167: Can we add to the unit test, to make sure the config is parsed properly.

diptanu (changes requested):
The main comments are -
1. Implementing the Langchain BaseChatMessageHistory interface for storage of messages.
2. Adding a retriever for searching memory.

PR #20, opened 2023-06-04
persist memory

khshah6:
- Simplify memory implementation
- Expose endpoints to create memory session, add to memory, retrieve from memory, and search memory

TODO (in following PRs)
- Create memory_sessions db table to persist session_id and corresponding index_name (1:1 for now)
- Add tests
- Add endpoints in python sdk
- Create conversation_history, retrieval, search langchain implementations
- Notebook with example


src/server.rs
```diff
@@ -355,10 +348,13 @@ async fn add_texts(
 
 #[axum_macros::debug_handler]
 async fn create_memory_session(
-    State(memory_manager): State<Arc<MemorySessionRouter>>,
+    State(memory_manager): State<Arc<Option<MemoryManager>>>,
```
diptanu, line 351: Let's make MemoryManager not optional, i.e - the service will come through with MemoryManager all the time.  1. Reduces complexity 2. Makes the surface of the product more consistent.

src/server.rs
```diff
@@ -389,11 +388,12 @@ async fn add_record(
...
 #[axum_macros::debug_handler]
-async fn retrieve_records(
-    State(memory_manager): State<Arc<MemorySessionRouter>>,
+async fn get_from_memory_session(
+    State(memory_manager): State<Arc<Option<MemoryManager>>>,
     Json(payload): Json<MemorySessionRetrieveRequest>,
 ) -> Result<Json<MemorySessionRetrieveResponse>, IndexifyAPIError> {
-    let result = memory_manager.retrieve_history(payload.session_id, payload.query);
+    let memory_manager = memory_manager.as_ref().as_ref().unwrap();
+    let result = memory_manager.retrieve_messages(payload.session_id).await;
```
diptanu, line 396: .map_err(|e| IndexifyAPIErr::new(status_code, e.to_string()))?

src/server.rs
```diff
@@ -402,7 +402,29 @@ async fn retrieve_records(
         ))
     } else {
```
diptanu, line 403: The else would go away after the previous comment.

src/server.rs
```diff
@@ -402,7 +402,29 @@ async fn retrieve_records(
...
+
+#[axum_macros::debug_handler]
+async fn search_memory_session(
+    State(memory_manager): State<Arc<Option<MemoryManager>>>,
+    Json(payload): Json<MemorySessionSearchRequest>,
+) -> Result<Json<MemorySessionSearchResponse>, IndexifyAPIError> {
+    let memory_manager = memory_manager.as_ref().as_ref().unwrap();
+    let result = memory_manager
+        .search(payload.session_id, payload.query, payload.k)
+        .await;
```
diptanu, line 418: .map_err

src/memory/mod.rs
```diff
@@ -1,182 +1,134 @@
...
-        }
+    ) -> Result<Uuid, MemoryError> {
         let session_id = session_id.unwrap_or(Uuid::new_v4());
-        let session = self.create_memory_session(session_id, memory_storage_policy);
-        self.router.insert(session_id, session.unwrap());
+        let index_name = self._get_index_name(session_id)?;
+        // TODO: Persist session_id and index_name to memory_sessions DB table
+        self.index_manager
+            .create_index(
+                CreateIndexParams {
```
diptanu, line 64: Is this a default? If it is, let's 1. allow users to specify the default in server config, otherwise it might now work, because all-minilm-l12-v2 might not be loaded in the server.

diptanu: We could do this as a followup doesn't need to be in this PR.

src/memory/mod.rs
```diff
@@ -1,182 +1,134 @@
...
+        session_id: Uuid,
+        messages: Vec<Message>,
+    ) -> Result<(), MemoryError> {
+        let index_name = self._get_index_name(session_id)?;
+        let texts = get_texts_from_messages(session_id, messages);
+        let index = self
+            .index_manager
+            .load(index_name.into())
+            .await
+            .unwrap()
```
diptanu, line 89: Probably want to do map_err()?

src/memory/mod.rs
```diff
@@ -1,182 +1,134 @@
...
+            .unwrap();
```
diptanu, line 90: .ok_or()?

src/memory/mod.rs
```diff
@@ -1,182 +1,134 @@
...
-    fn get_session(
-        &self,
-        session_id: Uuid,
-    ) -> Option<RefMut<Uuid, Arc<dyn MemorySession + Send + Sync>>> {
-        if !self.does_session_exist(Some(session_id)) {
-            return None;
-        }
-        return self.router.get_mut(&session_id);
+impl MemoryManager {
+    pub async fn new(index_manager: Arc<IndexManager>) -> Result<Option<Self>, MemoryError> {
```
diptanu, line 44: Lets make it not-optional. Result<Self, MemoryError>

diptanu (changes requested):
This looks great overall, just have some minor comments. Let's add a unit test on the memory manager so that we catch any breakages while refactoring.

diptanu (approved):
LGTM, looks like the memory test is failing. Please fix it before merging.

PR #21, opened 2023-06-05
adding memory table

khshah6:
Createing memory_sessions db table to persist session_id and corresponding index_name
Minor doc fix and config cleanup

TODO:
Replace temporary hashmap with table in memory_manager impl
Add endpoints in python sdk
Add tests
Create conversation_history, retrieval, search langchain implementations
Update documentation


migration/src/lib.rs
```diff
@@ -1,12 +1,16 @@
...
 mod m20220101_000001_create_table;
+mod m20230605_200842_create_memory_table;
 
 pub struct Migrator;
 
 #[async_trait::async_trait]
 impl MigratorTrait for Migrator {
     fn migrations() -> Vec<Box<dyn MigrationTrait>> {
-        vec![Box::new(m20220101_000001_create_table::Migration)]
+        vec![
```
diptanu, line 11: We can add the new tables in the existing migration.

diptanu (approved):
The only comment here is that we can use the existing migration, everything else looks great!

diptanu (approved):
The only comment here is that we can use the existing migration, everything else looks great!

migration/src/m20220101_000001_create_table.rs
```diff
@@ -62,12 +62,35 @@ impl MigrationTrait for Migration {
...
+                    )
+                    .col(ColumnDef::new(MemorySessions::Metadata).json())
+                    .to_owned(),
+            )
             .await
     }
 
     async fn down(&self, manager: &SchemaManager) -> Result<(), DbErr> {
-        manager
+        let _ = manager
```
MscotelloDeepWalk, line 89: Is this let _ neccessary?

src/memory/mod.rs
```diff
@@ -53,17 +56,17 @@ impl MemoryManager {
...
 
     fn _get_index_name(&self, session_id: Uuid) -> Result<String, MemoryError> {
         // TODO: Create better default index name without exposing session_id
         // TODO: Retrieve index_name from memory_sessions DB table
-        let binding = self.temp_session_id_index_map
-            .lock()
-            .unwrap();
-        let index_name = binding
-            .get(&session_id);
+        let binding = self.temp_session_id_index_map.lock().unwrap();
```
MscotelloDeepWalk, line 68: Yes, `unwrap()` is unsafe because it will panic if the lock is poisoned. Consider using `expect()` for a more descriptive panic message or handle the `Err` case explicitly.

```rust
let binding = self.temp_session_id_index_map.lock().expect("Failed to acquire lock");
```

Or handle error:

```rust
match self.temp_session_id_index_map.lock() {
    Ok(binding) => { /* use binding */ },
    Err(_) => { /* handle error */ },
}
```

PR #22, opened 2023-06-06
Update mem impl with persistence

khshah6:
Updating memory manager to use persistent memory_sessions table
Updating unit tests
Minor api change (adding optional metadata to persist to capture additional heuristics)

TODO:
Add endpoints in python sdk
Add tests
Create conversation_history, retrieval, search langchain implementations
Update documentation


src/index.rs
```diff
@@ -134,6 +135,28 @@ impl IndexManager {
         .await?;
         Ok(index)
     }
+
+    pub async fn create_index_for_memory_session(
+        &self,
+        session_id: Uuid,
+        index_name: String,
+        metadata: Option<HashMap<String, String>>,
```
diptanu, line 143: metadata: HashMap<> - this will be simpler when we serialize and store it in the database, and also de-serialize. And empty hashmp will be an empty json object.

src/memory/mod.rs
```diff
@@ -92,11 +86,13 @@ impl MemoryManager {
         vectordb_params: CreateIndexParams,
         embedding_model: String,
         text_splitter: TextSplitterKind,
+        metadata: Option<HashMap<String, String>>,
     ) -> Result<Uuid, MemoryError> {
         // TODO: Persist session_id and index_name to memory_sessions DB table
         let session_id = session_id.unwrap_or(Uuid::new_v4());
         let index_name = vectordb_params.name.clone();
-        self._set_index_name(session_id, index_name);
+        self._set_index_name(session_id, index_name, metadata)
```
diptanu, line 94: How do we roll back if creating the vector db for the index fails? We should do this in a single transaction below.

diptanu (commented):
Left some comments here related to integrity of the indexes in vectordb and metadata store.

PR #23, opened 2023-06-07
wrap memory writes in transaction

khshah6:
Addressing feedback from https://github.com/diptanu/indexify/pull/22


src/persistence.rs
```diff
@@ -132,13 +132,14 @@ impl Repository {
         Self { conn: db }
     }
 
-    pub async fn create_index(
+    async fn _create_index(
         &self,
+        tx: DatabaseTransaction,
```
diptanu, line 137: tx: &DatabaseTranscation

that way the method can still return () 
I believe almost every method which uses a transaction requires a reference to it anyways

src/persistence.rs
```diff
@@ -166,6 +166,20 @@ impl Repository {
...
+    pub async fn create_index(
+        &self,
+        embedding_model: String,
+        index_params: CreateIndexParams,
+        vectordb: vectordbs::VectorDBTS,
+        text_splitter: String,
+    ) -> Result<(), RepositoryError> {
+        let tx = self.conn.begin().await?;
+        let tx = self
+            ._create_index(tx, embedding_model, index_params, vectordb, text_splitter)
```
diptanu, line 181: &tx

src/persistence.rs
khshah6: it seems like sea_orm `DatabaseTranscation` doesn't have `Copy` trait, so I'm not sure how to mitigate issues when calling `tx.rollback().await?` after using `tx` in  `insert_result`


error:
```
`*tx` moved due to this method
cannot move out of `*tx` which is behind a shared reference
move occurs because `*tx` has type `DatabaseTransaction`, which does not implement the `Copy` trait
```

khshah6: skipping this refactor for now, based on offline discussion

PR #24, opened 2023-06-08
Update public docs

khshah6:
- Few minor fixes
- Updating public docs with memory information, as well as some structure changes and minor fixes
- Noticed issue with search results not having associated metadata; will follow up to either add metadata to point in vector DB or run search on sql table to gather associated metadata


docs/docs/apis/embeddings.md
```diff
@@ -0,0 +1,26 @@
+# Indexify APIs
```
diptanu, line 1: Can we have a main APIs page - explaining how the apis are divided across memory and retrieval? From there we can have links to the individual pages.

docs/docs/apis/memory.md
```diff
@@ -0,0 +1,67 @@
...
+```
+
+### Add Memory
+```
+POST /memory/add
+```
+#### Request Body
+* `session_id` - UUID corresponding to memory session.
+* `messages` -
+    * `role` - Role (typically `Human` or `AI`).
```
diptanu, line 34: make them lower cased? `human` and `ai`

docs/docs/apis/memory.md
```diff
@@ -0,0 +1,67 @@
...
+
+Memory APIs are centered around retrieving memory from conversation history to use in context. Indexify supports retrieving all conversation history for a memory session, as well as query-based semantic search.
+
+
+### Create Memory Session
+```
+POST /memory/create
+```
+#### Request Body
+* `index_args`
```
diptanu, line 15: It's kinda weird that we ask users to specify a vector index name while creating a memory session. I think we will have the opportunity to make the UX better when we have repositories to add content and attach indexes to them in the form of "extractors" specifically for retrieval.

Just pointing it out, not requesting to change what is on the PR now.

docs/docs/apis/memory.md
```diff
@@ -0,0 +1,67 @@
...
+```
+curl -X POST http://localhost:8900/memory/add   -H "Content-Type: application/json" -d '{"session_id": "77569cf7-8f4c-4f4b-bcdb-aa54355eee13", "messages": [{"role": "Human", "text": "Indexify is amazing!", "metadata": {}}]}'
+```
+
+### Retrieve Memory
+```
+GET /memory/get
+```
+#### Request Body
+* `session_id` - UUID corresponding to memory session.
```
diptanu, line 48: Can we relax this by allowing users to specify non-UUID based identifiers as well? So basically string based unique identifiers. in the event they are not passed we can create it for them as you already have.

diptanu (changes requested):
The main request to change here is creating a parent API page to explain that there are two APIs mainly - Rest and Python. And a conceptual overview of the api, and then link to the memory and retrieval api pages.

Unrelated to the PR - We should document clearly the optional, required fields and the default values. They don't exist for the current docs as well, I have a note to do it when I add the repository docs.

src/server.rs
```diff
@@ -411,7 +413,9 @@ async fn add_to_memory_session(
         .await
         .map_err(|e| IndexifyAPIError::new(StatusCode::INTERNAL_SERVER_ERROR, e.to_string()))?;
 
-    Ok(Json(MemorySessionAddResponse {}))
+    Ok(Json(MemorySessionAddResponse {
+        successful: true
```
diptanu, line 417: Do we need this? Under what conditions would this be successful: false, and not return IndexifyAPIError?

src/persistence.rs
```diff
@@ -237,9 +250,15 @@ impl Respository {
...
         index_name: String,
-        metadata: Option<HashMap<String, String>>,
+        metadata: HashMap<String, String>,
+        vectordb_params: CreateIndexParams,
+        embedding_model: String,
+        vectordb: vectordbs::VectorDBTS,
+        text_splitter: String,
     ) -> Result<(), RespositoryError> {
-        let tx = self.conn.begin().await?;
+        let tx = self
```
diptanu, line 259: I believe this will change based on the other PR.

docs/docs/develop.md
```diff
@@ -10,7 +10,7 @@ Install various rust related tools -
 ### Python Dependencies
 Create a virtual env 
 ```
-virtualenv ve
+python3.11 -m venv ve
```
diptanu, line 13: Do we have a hard dependency on 3.11 can we use `python3` instead? I think that would alias to python3.11 or python 3.9 on most systems.

diptanu (changes requested):

docs/docs/develop.md
khshah6: thats fine, we'll just have to modify this then

`export PYTHONPATH=${PYTHONPATH}:${PWD}/ve/lib/python3.11/site-packages`

src/persistence.rs
khshah6: We decided to leave it as is for now

docs/docs/apis/embeddings.md
khshah6: sure, added

diptanu (approved):
LGTM!

PR #28, opened 2023-06-08
WIP: Repository Manager

diptanu:
* Created public endpoints for managing data-repositories.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -1,7 +1,9 @@
+use crate::data_repository_manager::DataRepositoryManager;
 use crate::index::IndexManager;
-use crate::persistence::Text;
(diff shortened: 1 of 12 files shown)


src/server.rs
```diff
@@ -295,6 +442,74 @@ async fn root() -> &'static str {
     "Indexify Server"
 }
 
+#[axum_macros::debug_handler]
+async fn sync_repository(
```
khshah6, line 446: should we change this name to `create_repository` or `initialize_repository`?

khshah6 (commented):
looks good, just left a comment about naming

src/server.rs
diptanu: @khshah6 This API is declarative - any changes to the repository after it's been created like adding/removing extractors happen via the same API.

PR #46, opened 2023-06-28
Bug Fix + Refactor python SDK to be friendly with setuptools

Vijay2win:


sdk-py/poetry.lock
diptanu, line 1: This needs to come back.

sdk-py/pyproject.toml
diptanu, line 1: This needs to come back. We are using poetry to package, distribute and push the package to pypi - https://pypi.org/project/indexify/

sdk-py/setup.py
diptanu, line 1: This is not required as pyproject is recommended over setuptools.

sdk-py/requirements.txt
diptanu, line 1: Dependencies should be in pyproject

sdk-py/indexify/memory.py
```diff
@@ -0,0 +1,40 @@
+import requests
+
+from indexify import *
+from utils import _get_payload
+
+
+class Memory:
+    DEFAULT_INDEXIFY_URL = "http://localhost:8900"
```
diptanu, line 8: Please move this in a const file or something as bunch of classes would refer to it.

sdk-py/indexify/utils.py
```diff
@@ -0,0 +1,28 @@
...
+        return self.name.lower()
+
+
+def _get_payload(response):
+    payload = {"errors": []}
+    try:
+        payload = json.loads(response.text)
+    except:
+        raise ApiException(response.text)
+    if "errors" in payload.keys() and len(payload["errors"]) > 0:
```
diptanu, line 25: We no longer have errors in payload, so not required.

sdk-py/indexify/index.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class Index:
+    DEFAULT_INDEXIFY_URL = "http://localhost:8900"
+
+    def __init__(self, url, index):
+        self._url = url
+        self._index = index
+
+    def sync(self, name: str, extractor_type):
```
diptanu, line 14: Please remove this or fix the signature of the sync method - its name and a list of extractors.

sdk-py/indexify/data_containers.py
```diff
@@ -0,0 +1,43 @@
...
+    def __init__(self, role: str, text: str, metadata: dict = {}):
+        self.role = role
+        self.text = text
+        self.metadata = metadata
+
+    def to_dict(self):
+        return {"role": self.role, "text": self.text, "metadata": self.metadata}
+
+
+class SearchChunk:
```
diptanu, line 33: Does this work? Assuming this is a data class, so probably needs a @dataclass decorator?

sdk-py/indexify/data_containers.py
```diff
@@ -0,0 +1,43 @@
...
+class TextChunk:
+    def __init__(self, text: str, metadata: Optional[dict] = {}):
+        self.text = text
+        self.metadata = metadata
+
+    def to_dict(self):
+        return {"text": self.text, "metadata": self.metadata}
+
+
+class Message:
```
diptanu, line 23: Make this a data class?

sdk-py/examples/question_answer_demo.py
diptanu, line 1: This is a great! I think it would work really well if this was a notebook, then we could have embedded it right in the docs of the website :)

diptanu (changes requested):
Thanks, looks great! Small nits in code, and let’s bring back the pyproject related files, which is currently being used to build and push to pypi.

sdk-py/Makefile
diptanu, line 1: Can we also bring this back? This has some handy scripts to lint and build the package.

diptanu (changes requested):

sdk-py/examples/simple_example.py
diptanu, line 1: There is a notebook in the docs - https://github.com/diptanu/indexify/blob/main/docs/docs/python_sdk.ipynb If you update that, it will automatically get compiled and published on the website.

sdk-py/poetry.lock
jjmachan: the poetry.lock file does not have to be added for libraries and SDKs right? pip will be able to take it from pyproject or setup.py and whatever is installed on the user's system. If we lock it that limits things at the user end I believe?

diptanu: Good point @jjmachan but the other issue is version mismatch of what we test with vs what the user has could lead to inconsistent behavior/bugs.

PR #51, opened 2023-06-29
adapter changed

kitrak-rev:
-- Changing the axios adapter to be inclusively compatible for browser extensions.

### DIFF
--- a/sdk-typescript/package.json
+++ b/sdk-typescript/package.json
@@ -93,5 +93,8 @@
     "exclude": [
       "**/*.spec.js"
     ]
+  },
+  "dependencies": {
+    "@kitrak_rev/axios-fetch-adapter": "^0.3.3"
   }
 }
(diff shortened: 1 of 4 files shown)


sdk-typescript/package.json
```diff
@@ -93,5 +93,8 @@
     "exclude": [
       "**/*.spec.js"
     ]
+  },
+  "dependencies": {
+    "@kitrak_rev/axios-fetch-adapter": "^0.3.3"
```
diptanu, line 98: Can’t we use the normal axios-fetch adapter?

sdk-typescript/src/core/request.ts
```diff
@@ -12,6 +14,8 @@ import { CancelablePromise } from './CancelablePromise';
 import type { OnCancel } from './CancelablePromise';
 import type { OpenAPIConfig } from './OpenAPI';
 
+
```
diptanu, line 17: Remove the white spaces.

diptanu (changes requested):

PR #54, opened 2023-06-30
Fix devcontainer for Github Codespaces

delip:
- devcontainer fix


diptanu:
Closed it since I needed this to fix the other remaining issues.

PR #63, opened 2023-07-04
Langchain memory

Vijay2win:


sdk-py/indexify/langchain_integration.py
```diff
@@ -0,0 +1,63 @@
...
+
+'''
+This class will initialize the Indexify class with the indexify_url your installation
+
+Example:
+memory = IndexifyMemory(indexify_url="http://10.0.0.1:8900/")
+'''
+
+
+class IndexifyMemory(BaseChatMemory):
```
diptanu, line 15: Lets move this to a top level directory indexify/langchain-integration/memory/indexify.py

Once we have that, we could have a pyproject.toml in that folder which publishes a new library `indexify-langchain` which depends on indexify package.

diptanu (approved):
Merging these changes, left a comment about how we can extract the langchain integration in a separate package.

PR #69, opened 2023-07-09
Update Document for Langchain

Vijay2win:
Add to_document method for easier Vector support in langchain, update QA examples


Vijay2win:
@diptanu the latest commit should address the comments from #64

diptanu:
Langchain is already in master!

PR #73, opened 2023-07-16
Feature/ner extractor

vidhyaarvind:


src_py/indexify_py/entity_extractor.py
```diff
@@ -0,0 +1,53 @@
...
+            "I-ORG": "Organization",
+            "B-LOC": "Location",
+            "I-LOC": "Location"
+         }   
+        return entity_type_map.get(type, "INVALID")
+    
+
+class EntityExtractor:
+   
+    def __init__(self):
```
diptanu, line 34: Can we move the model name into the constructor? with the default value as the model you used here?

diptanu (approved):
Thanks! LGTM, just a small minor comment about moving the model name into the constructor.

src_py/indexify_py/entity_extractor.py
diptanu: Depending on which model we plug in, I realize the extraction results will be slightly different. I guess we will need to make sure the model name is from a supported list, and given the model name, the output of the entity name is normalized to what we expect.

PR #77, opened 2023-07-28
Switch to stable dns for docker internal network

rylandg:
Use the stable internal network dns that docker provides instead of hand defining ip addresses for the services. I've only tested this on OSX but generally speaking docker internal host dns is cross platform. 

This change was required because docker networking overlay works poorly on OSX due to the virtualization required to run Docker itself. This greatly limits the flexibility of ip tables and network configuration which resulted in a broken docker-compose startup on my M1 machine.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -18,7 +18,7 @@ services:
       - ./sample_config.yaml:/indexify/config/indexify.yaml:ro
     environment:
       - OPENAI_API_KEY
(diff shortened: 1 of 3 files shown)


rylandg:
Alternative changes were made by @diptanu that solves this underlying issue and obviates these changes.

For posterity, if you're having issues with Docker network setup on a M1 machine, it may mean that your Docker needs to be updated or clean installed (it can't hurt at least). Please file an issue if the problem persists after that.

PR #79, opened 2023-07-31
Update TS sdk and examples

rylandg:
Fixed the getting started based on api changes. Made it simpler to run the examples. Also fixed an issue which made it impossible to actually run the generated code because we were using import and node even on V20.5.0 still doesn't like that by default.


diptanu:
Amazing, thanks!

PR #90, opened 2023-08-10
major updates to python sdk

shabani1:
- Created Indexify client, and classes for Extractor and RepoExtractor (an extractor bound to a repo)
- Updated the getting started page

### DIFF
--- a/sdk-py/indexify/extractor.py
+++ b/sdk-py/indexify/extractor.py
@@ -1,10 +1,6 @@
 import requests
 
-
-from .index import Index
-from .data_containers import TextChunk
 from .settings import DEFAULT_INDEXIFY_URL
-from .utils import _get_payload, wait_until
 
 
 def list_extractors(base_url: str = DEFAULT_INDEXIFY_URL) -> list[dict]:
@@ -14,6 +10,7 @@ def list_extractors(base_url: str = DEFAULT_INDEXIFY_URL) -> list[dict]:
 
 
 # TODO: consider naming this IndexifyExtractor
+# TODO: consider making this a dataclass
 class Extractor:
 
     def __init__(self, name: str, description: str, extractor_type : dict):
--- a/sdk-py/indexify/extractor.py
+++ b/sdk-py/indexify/extractor.py
@@ -1,0 +1,28 @@
+import requests
+
+
+from .index import Index
+from .data_containers import TextChunk
+from .settings import DEFAULT_INDEXIFY_URL
+from .utils import _get_payload, wait_until
+
+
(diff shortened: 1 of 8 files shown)


sdk-py/indexify/repository.py
```diff
@@ -20,6 +22,22 @@ def list_repositories(base_url: str = DEFAULT_INDEXIFY_URL) -> list[dict]:
     return response.json()['repositories']
 
 
+# TODO: consider tying this back to IndexifyExtractor
+class RepoExtractor:
```
diptanu, line 26: If this is just a data class, we could just use `@dataclass` - would achieve the same outcome with less boilerplate code.

sdk-py/indexify/extractor.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+def list_extractors(base_url: str = DEFAULT_INDEXIFY_URL) -> list[dict]:
+    response = requests.get(f"{base_url}/extractors")
+    response.raise_for_status()
+    return response.json()['extractors']
+
+
+# TODO: consider naming this IndexifyExtractor
+class Extractor:
```
diptanu, line 17: May be use dataclass here unless we are going to add some extra behavior in the form of methods?

sdk-py/indexify/repository.py
```diff
@@ -44,24 +62,15 @@ async def add(self, *chunks: TextChunk) -> None:
                 return await _get_payload(resp)
 
     async def add_documents(self, *documents: dict) -> None:
```
diptanu, line 64: what does this dict going to have? {"text": "....", "metadata": {...} } ? 
It's not quite clear from the signature of the method.

diptanu (commented):
Thanks and looks good overall, only have some small comments.

src/server.rs
```diff
@@ -112,6 +112,7 @@ impl Server {
                 post(run_extractors).with_state(repository_endpoint_state.clone()),
             )
             .route(
+                // FIXME: this should be /index/search
```
diptanu, line 115: Indexes are not global so I wonder if they should be a top level API such as this? for ex, index `foo` could be in one or many repositories.

shabani1: Discussed. Diptanu will update paths to read `/repository/search`

sdk-py/indexify/repository.py
shabani1: Removed the extra `add()` method as discussed

sdk-py/indexify/extractor.py
shabani1: Put a TODO to consider converting to dataclass if no methods

PR #94, opened 2023-08-11
Api refactor

diptanu:
* Existing APIs were RPC methods over HTTP
* Refactored them to look more RESTful.

### DIFF
--- a/docs/docs/getting_started.md
+++ b/docs/docs/getting_started.md
@@ -48,6 +48,7 @@ Data Repositories are logical buckets that store content. Indexify starts with a
             {"text": "Kevin Durant is the best basketball player in the world.", "metadata": {"topic": "nba"}}
         ]}' 
     ```
+<<<<<<< HEAD
 === "python"
 
     ```python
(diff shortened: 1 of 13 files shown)


src/server.rs
```diff
@@ -89,44 +90,43 @@ impl Server {
...
                 post(add_events).with_state(repository_endpoint_state.clone()),
             )
             .route(
-                "/events",
+                "/events/:repository_name",
                 get(list_events).with_state(repository_endpoint_state.clone()),
             )
             .route(
-                "/repository/create",
+                "/repositories",
```
shabani1, line 121: I think these next 3 should be `repository` also instead of `repositories`? And they'll have the same base as the actions above

src/server.rs
```diff
@@ -89,44 +90,43 @@ impl Server {
...
                 get(index_search).with_state(repository_endpoint_state.clone()),
             )
             .route(
-                "/repository/attribute_lookup",
+                "/repository/:repository_name/attributes",
                 get(attribute_lookup).with_state(repository_endpoint_state.clone()),
             )
             .route(
-                "/events",
+                "/events/:repository_name",
```
shabani1, line 113: Is this getting events from a specific repository? If so, the url should probably be `/repository/:repository_name/events`

shabani1 (commented):
This is awesome, thank you! ❤️

docs/docs/getting_started.md
```diff
@@ -48,6 +48,7 @@ Data Repositories are logical buckets that store content. Indexify starts with a
             {"text": "Kevin Durant is the best basketball player in the world.", "metadata": {"topic": "nba"}}
         ]}' 
     ```
+<<<<<<< HEAD
```
shabani1, line 51: Need to delete merge conflict markers in this file

src/server.rs
shabani1: Actually now I'm not sure about this one 🤔 cause then everything will end up under repository. So maybe /events/ but then repository name is an arg? Not sure the best way to do it.

src/server.rs
diptanu: `repositories` is the resource to create a new repository or list them, everything else which is related to a specific repository is on `/repository<action>`

shabani1: Gotcha. This is the way I'm used to seeing it, but open to other approaches.

List repos: `GET /repository`
Get a specific repo: `GET /repository/:repo_name`
Create repo: `PUT /repository/:repo_name`
Some action on a repo: `GET/POST/PUT /repository/:repo_name/<action>`

PR #95, opened 2023-08-16
Updates to python sdk for new server api

shabani1:
There are a number of things still broken that we need to implement on the server side, and I added new FIXME comments here for each of them.


shabani1:
This is a good stopping point for this PR since I can't make more changes without changing the server side. Want to get these in first and then do the server fixes in tandem.

shabani1:
Pushed an update, I realized what the query attribute endpoint needs so updated that call so it's now working.

docs/docs/getting_started.md
```diff
@@ -105,8 +105,8 @@ Every extractor we bind results in a corresponding index being created in Indexi
 === "python"
 
     ```python
-    repo.add_extractor("EntityExtractor", index_name="entityindex", filter={"content_type": "text"})
-    repo.add_extractor("MiniLML6", index_name="embeddingindex", filter={"content_type": "text"})
+    repo.bind_extractor("EntityExtractor", index_name="entityindex")
```
diptanu, line 108: Add filters attribute here?

sdk-py/README.md
```diff
@@ -0,0 +1,21 @@
+# Indexify Python Client
+
+## Development
```
diptanu, line 3: This is the README that a developer will see on pypi. So we want a different set of instructions here like -

```
pip install indexify
```

And then some examples of how to use the client.

sdk-py/examples/benchmark.py
```diff
@@ -1,15 +1,15 @@
...
+from indexify import AIndex, ARepository, AMemory, TextChunk, DEFAULT_SERVICE_URL, wait_until, Message
 from datasets import load_dataset
 
 
 class BulkUploadRepository:
 
     def __init__(self, url):
         self.url = url
-        self.repository = ARepository(url)
+        self.repository = ARepository(, url
```
diptanu, line 12: Is this a typo?

src/server.rs
```diff
@@ -273,6 +273,9 @@ async fn get_repository(
 )]
 #[axum_macros::debug_handler]
 async fn bind_extractor(
+    // TODO: shouldn't index_name be required here?
```
diptanu, line 276: Extractor Bindings create indexes, not the other way round. The payload contains the index_name that the binding will populate.

src/vector_index.rs
```diff
@@ -51,6 +51,7 @@ impl VectorIndexManager {
     ) -> Result<()> {
         let mut index_params: Option<CreateIndexParams> = None;
         if let ExtractorType::Embedding { dim, distance } = &extractor_config.extractor_type {
+            // FIXME: "{}/{}" should be "{}.{}" or some other separator that's not url encoded
```
diptanu, line 54: I agree, but does it cause any issues?

diptanu (changes requested):

docs/docs/develop.md
```diff
@@ -59,15 +59,10 @@ Build the server in development mode
 cargo build
 ```
 
-### Create a development database
-```
-make migrate-dev
```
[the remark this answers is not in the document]
shabani1: Oh yea, good point. I had removed it because the command was above, but we need it here too.

docs/docs/getting_started.md
```diff
@@ -53,9 +53,9 @@ Data Repositories are logical buckets that store content. Indexify starts with a
 
     repo = Repository()
     repo.add_documents([
-        {"text": "Indexify is amazing!", "metadata": {"topic": "llm"}},
-        {"text": "Indexify is a retrieval service for LLM agents!", "metadata": {"topic": "ai"}},
-        {"text": "Kevin Durant is the best basketball player in the world.", "metadata": {"topic": "nba"}}
+        {"text": "Indexify is amazing!"},
```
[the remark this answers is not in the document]
shabani1: For the getting started, I think it's much more clear this way. This was some of the feedback that I provided the first time I looked at this, which is "as a developer, it's not clear to me where those metadata topics came from, and most of my docs won't have that. They look like things that would get created by an extractor."

I agree that the filter feature for the bindings is a great one. Would it be okay if we kept this and then demonstrated that example as part of a different tutorial? Like a separate extractors tutorial or something? I just want this part to be crystal clear to someone seeing indexify for the first time.

docs/docs/getting_started.md
shabani1: It's there as a default argument right now. Python method has the value here as the default.

sdk-py/README.md
shabani1: Oh crap, you're right! I was writing notes for myself for which steps to repeat each time, and I shouldn't have put them here. Will update.

sdk-py/examples/benchmark.py
shabani1: Must've screwed this up in the refactor 🤔

src/server.rs
shabani1: Right, that's what I'm saying. But currently in the payload, the index_name is [optional](https://github.com/diptanu/indexify/blob/e34cf94e018d8bbe19c8bdc23d922e65f3500a34/src/api.rs#L79), shouldn't it be required?

src/vector_index.rs
shabani1: It creates ambiguity for the qdrant API, since collection name has a slash in it. For example, I can't visit `http://localhost:6333/collections/default/embeddingindex` to view my embedding index stats, I have to visit `http://localhost:6333/collections/default%2Fembeddingindex` instead. If it's not difficult to change might be worth doing it now.

shabani1:
Made requested changes.

docs/docs/getting_started.md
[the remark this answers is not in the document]
diptanu: Got it. Can you please add some examples with filters?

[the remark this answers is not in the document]
shabani1: As discussed, will add new examples with filters in a future PR. For now, keeping the metadata in CURL examples.

PR #98, opened 2023-08-18
updates to filters with extractor bindings

shabani1:
Implemented `filter` and `exclude` arguments in `repo.bind_extractors` method. Also added a docstring with a couple of examples.

### DIFF
--- a/sdk-py/indexify/repository.py
+++ b/sdk-py/indexify/repository.py
@@ -74,9 +74,41 @@ class Repository(ARepository):
     def add_documents(self, *documents: dict) -> None:
         return wait_until(ARepository.add_documents(self, *documents))
 
-    # FIXME: filter should be filters to be consistent with extractor_bindings
-    def bind_extractor(self, name, index_name, filter={"content_type": "text"}) -> dict:
-        req = {"extractor_name": name, "index_name": index_name, "filter": filter}
+    # TODO: what happened to content_type argument?
+    def bind_extractor(self, extractor_name: str, index_name: str,
+                       filters: dict | None = None,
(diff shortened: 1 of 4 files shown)


sdk-py/indexify/repository.py
```diff
@@ -74,9 +74,41 @@ def __init__(self, name: str = "default", service_url: str = DEFAULT_SERVICE_URL
     def add_documents(self, *documents: dict) -> None:
         return wait_until(ARepository.add_documents(self, *documents))
 
-    # FIXME: filter should be filters to be consistent with extractor_bindings
-    def bind_extractor(self, name, index_name, filter={"content_type": "text"}) -> dict:
-        req = {"extractor_name": name, "index_name": index_name, "filter": filter}
+    # TODO: what happened to content_type argument?
+    def bind_extractor(self, extractor_name: str, index_name: str,
```
diptanu, line 78: Its not very clear what filters and exclude mean here, here is a suggestion -

We use a Filter object to contain a list of filters - could be equality or un-equality filters. A FilterBuilder can help in building the filters.


```
class FilterBuilder:
     def add_eq(self, key, value) -> FilterBuilder:
       ....
     def add_neq(self, key, value) -> FilterBuilder:
      ....
      def build(self) -> Filter:
        ....
        
        
def bind_extractor(self, extractor_name: str, index_name: str, predicates: Filter): 
     ...
```

shabani1: As discussed: we should build this filter builder, and I'll do so as part of the next PR. For now, changing `filters` to `include`.

PR #103, opened 2023-08-22
updating the python sdk to support server features

diptanu:
WIP 

Need to add the async versions to APIs that search, lookup indexes and to the api that adds text. 

Control plane apis, which are called infrequently can be left synchronous for now.


sdk-py/indexify/client.py
```diff
@@ -8,23 +8,22 @@ class IndexifyClient:
     def __init__(self, service_url: str = DEFAULT_SERVICE_URL):
         self._service_url = service_url
 
-    def create_repository(self, name: str, extractors: list = [], metadata: dict = {}) -> dict:
-        return create_repository(name, extractors, metadata, self._service_url)
-
-    @property
```
shabani1, line 14: Why remove property decorator? Won't the new approach not update the set of extractors when called?

sdk-py/indexify/repository.py
```diff
@@ -1,55 +1,53 @@
...
+        return ExtractorBinding(**json)
 
+class Repository:
 
-class ARepository:
-
-    def __init__(self, name: str, service_url: str):
-        self.name = name
+    def __init__(self, name: str, service_url:str):
+        self._name = name
```
shabani1, line 33: I had previously changed this from `_name` to `name`. I think it's nice to have that as a public property that a dev could easily use for logging, for example.

sdk-py/indexify/repository.py
```diff
@@ -1,55 +1,53 @@
...
+@dataclass
 class ExtractorBinding:
-
-    def __init__(self, extractor_name: str, index_name: str, filters: dict, input_params: dict):
-        self.extractor_name = extractor_name
-        self.index_name = index_name
-        self.filters = filters
-        self.input_params = input_params
+    index_name: str
+    extractor_name: str
```
shabani1, line 17: Nit: I'd move this up before `index_name`

sdk-py/indexify/repository.py
```diff
@@ -1,55 +1,53 @@
...
+    filters: list[Filter] 
+    input_params: dict
 
     def __repr__(self) -> str:
         return f"ExtractorBinding(extractor_name={self.extractor_name}, index_name={self.index_name})"
 
     def __str__(self) -> str:
         return self.__repr__()
+    
+    def from_dict(self, json: dict):
```
shabani1, line 27: Love this 👍

sdk-py/indexify/repository.py
```diff
@@ -1,55 +1,53 @@
...
+class Repository:
 
-class ARepository:
-
-    def __init__(self, name: str, service_url: str):
-        self.name = name
+    def __init__(self, name: str, service_url:str):
+        self._name = name
         self._service_url = service_url
-        self.url = f"{self._service_url}/repositories/{self.name}"
```
shabani1, line 45: Why'd we get rid of this? Is it added elsewhere?

sdk-py/indexify/repository.py
```diff
@@ -1,55 +1,53 @@
...
+
+        response = httpx.get(f"{self._service_url}/repositories/{self._name}")
+        response.raise_for_status()
+        repository_json = response.json()['repository']
+        extractor_bindings = []
+        for eb in repository_json['extractor_bindings']:
```
shabani1, line 40: Don't we want to do these (`extractor_bindings` and `metadata`) as properties the way I have them below? That way they're updated whenever called.

sdk-py/indexify/repository.py
```diff
@@ -1,55 +1,53 @@
...
+    def from_dict(self, json: dict):
+        return ExtractorBinding(**json)
 
+class Repository:
 
-class ARepository:
-
-    def __init__(self, name: str, service_url: str):
-        self.name = name
+    def __init__(self, name: str, service_url:str):
```
shabani1, line 32: Nit: PEP8

sdk-py/indexify/repository.py
```diff
@@ -1,55 +1,53 @@
...
+        self._name = name
         self._service_url = service_url
-        self.url = f"{self._service_url}/repositories/{self.name}"
+
+        response = httpx.get(f"{self._service_url}/repositories/{self._name}")
```
shabani1, line 36: Does instantiating a repo no longer create a new repo then? Won't this line fail if the repo doesn't exist on the backend?

PR #114, opened 2023-09-26
Postgres Embedding Store: pg_embedding from neon tech

yenicelik:


src/vectordbs/pg_embedding.rs
```diff
@@ -0,0 +1,227 @@
...
+                DbBackend::Postgres,
+                query,
+                parameters,
+            ))
+            .await
+            .map_err(|e| {
+                VectorDbError::IndexDeletionError(index.to_string(), format!("{:?}", e))
+            })?;
+        if exec_res.rows_affected() == 0 {
+            Err(VectorDbError::IndexDeletionError(
```
diptanu, line 126: If rows affected is 0, does that mean the PK exists ?

diptanu (changes requested):

src/vectordbs/pg_embedding.rs
```diff
@@ -0,0 +1,227 @@
...
+                                .collect(),
+                        )),
+                    ),
+                ]
+            })
+            .collect::<Vec<sea_orm::Value>>();
+        parameters.insert(0, sea_orm::Value::String(Some(Box::new(index.to_string()))));
+        let exec_res: ExecResult = self
+            .db_conn
+            .execute(Statement::from_sql_and_values(
```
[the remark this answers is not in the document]
yenicelik: Yes that would work. I was thinking about the case that people may try to do SQL-injection from the frontend. I don't know enough about the whole pipeline (from API input until this stage), it may not be needed. Let me know if SQL injection is irrelevant for these, then i'll change it to from_string

src/vectordbs/mod.rs
```diff
@@ -3,12 +3,14 @@ use std::sync::Arc;
 use anyhow::Result;
 use async_trait::async_trait;
 
+use sea_orm::FromQueryResult;
```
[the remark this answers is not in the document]
yenicelik: It is actually used here: https://github.com/diptanu/indexify/blob/ea5fbee6ab964676af946b53201b621215d66dea/src/vectordbs/mod.rs#L38 for easier querying with sea-orm. 

You are right that there's a lot of unused items, I will remove them once the PR becomes final

src/vectordbs/pg_embedding.rs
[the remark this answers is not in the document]
yenicelik: also for embedding not serializing the embeddings into the string will likely be faster, although idk enough about how sea-orm works internally here

src/vectordbs/pg_embedding.rs
yenicelik: hmm I think this can only happen if the there are no rows inserted. 
I just realized, instead of INSERT, we should probably do an UPSERT, if you want insertion to also be idempotent?

yenicelik: INSERT will throw an error on duplicate PKs

yenicelik: I just see one of the tests is called "insert idempotent", so I will change it to an UPSERT now.

yenicelik:
Includes https://github.com/diptanu/indexify/pull/117 , which was required to run it locally

PR #117, opened 2023-10-02
Maintenance: Make Devcontainer work again

yenicelik:
Tested in VSCode devcontainer. Some modifications to make the tests run. 2 out of the 9 tests are failing, this may or may not be related to points specific to the devcontainer.


diptanu:
LGTM but we need to get the code which pumps in version into the binary working again before making a release

PR #119, opened 2023-10-03
Postgres Embedding Store: pg_vector instead of pg_embedding

yenicelik:
pg_embedding is deprecated https://neon.tech/docs/extensions/pg_embedding#migrate-from-pg_embedding-to-pgvector


src/vectordbs/pg_embedding.rs
```diff
@@ -84,29 +86,36 @@ impl VectorDb for PgEmbedding {
...
-        let query = r#"SET enable_seqscan = off;"#;
-        // The "enable seqscan = off" should not be set for the whole server. Will need to put this out
+
+        // Create index if dimensions is below 2000. It will be much slower for larger vectors, we should be explicit about this in the docs
+        // This is a limitation of pg_vector
+        if vector_dim <= 2000 {
+            warn!("Parameters are: {:?}", self.config);
+            let query = format!(
+                r#"
+                    CREATE INDEX ON {INDEX_TABLE_PREFIX}{index_name} USING hnsw(embedding {distance_extension}) WITH (m = {}, ef_construction = {});
```
diptanu, line 96: Let's add IF NOT EXISTS also.

src/vectordbs/pg_embedding.rs
```diff
@@ -84,29 +86,36 @@ impl VectorDb for PgEmbedding {
...
+            warn!("Running query: {:?}", query);
+            let exec_res: ExecResult = self
+                .db_conn
+                .execute(Statement::from_string(DbBackend::Postgres, query))
+                .await
+                .map_err(|e| {
+                    VectorDbError::IndexCreationError(format!("{:?}: {:?}", index_name.clone(), e))
+                })?;
+            // If more than one row is affected, something funky has occurred
+            if exec_res.rows_affected() > 1 {
```
diptanu, line 109: Do we need this? If the above queries don't work, the await should fail, and the exception would be thrown upwards. When will this case ever happen?

diptanu (changes requested):

run_tests.sh
```diff
@@ -32,8 +33,8 @@ until curl --output /dev/null --silent --get --fail http://$QDRANT_HOST/collecti
   sleep 5
 done
 
-
 timeout 90s bash -c "until docker exec postgres_test pg_isready ; do sleep 5 ; done"
+docker exec postgres_test psql -U postgres -c 'create database indexify;'
```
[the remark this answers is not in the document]
yenicelik: I am creating the extension whenever a new index is inserted (assuming the extension is not activated when the first index is added). 

https://github.com/diptanu/indexify/blob/9a21d102fd46d23922fafa65443a0633e249045d/src/vectordbs/pg_embedding.rs#L72 

Do you want to keep it like this, or do you want to CREATE EXTENSION as a preparation step? Both sound viable to me, especially since index tables will not be added frequently

src/vectordbs/pg_embedding.rs
yenicelik: The "query did not work" error is this one here: https://github.com/diptanu/indexify/blob/9a21d102fd46d23922fafa65443a0633e249045d/src/vectordbs/pg_embedding.rs#L107 and it is thrown upward as you mention. this if statement checks how many rows were affected. in the case of CREATE INDEX, no more than 1 row should ever be affected, hence the if-statement. we can remove this if you want, this case should never occur. but there is also no harm in keeping it in case anything "funky" ever occurs (and more than 1 table is affected), it will be easier to debug.

tldr: this should never happen

run_tests.sh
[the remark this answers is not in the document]
diptanu: Let's remove it if we don't expect it to happen ever.

src/vectordbs/pg_embedding.rs
```diff
@@ -40,12 +42,12 @@ impl VectorDb for PgEmbedding {
         let index_name = Self::escape_to_valid_indexname(index_name);
```
diptanu, line 42: Not part of this PR but I think we can refactor this in a upcoming pR 

pub struct IndexName(String);

impl IndexName {
   pub new(name: &str) -> Self {
     let name =  // logic to escape index name
     Self {name}
   }
}

run_tests.sh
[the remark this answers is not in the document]
diptanu: Ah I see, what you are doing is better. Let's keep it the way it is. Otherwise we won't be able to create indexes dynamically.

src/vectordbs/pg_embedding.rs
```diff
@@ -84,29 +86,37 @@ impl VectorDb for PgEmbedding {
...
-                index_name.clone(),
-                exec_res.rows_affected()
-            )));
-        };
-        let query = r#"SET enable_seqscan = off;"#;
-        // The "enable seqscan = off" should not be set for the whole server. Will need to put this out
+
+        // Create index if dimensions is below 2000. It will be much slower for larger vectors, we should be explicit about this in the docs
+        // This is a limitation of pg_vector
+        if vector_dim <= 2000 {
```
diptanu, line 92: Do this check at the top, so that if the dimensions are more than what we can handle we just throw an exception and the tables are never created. The user will then never be able to use the extractor unless they use a different storage. 

```
if vector_dim > 2000 { 

   // return Err 
}
```

src/vectordbs/pg_embedding.rs
```diff
@@ -84,29 +86,37 @@ impl VectorDb for PgEmbedding {
...
+            warn!("Parameters are: {:?}", self.config);
+            // IF NOT EXISTS requires an index-name in postgres. "_{INDEX_TABLE_PREFIX}{index_name}_hnsw" is automatically set as the index-name
+            let query = format!(
+                r#"
+                    CREATE INDEX IF NOT EXISTS _{INDEX_TABLE_PREFIX}{index_name}_hnsw ON {INDEX_TABLE_PREFIX}{index_name} USING hnsw(embedding {distance_extension}) WITH (m = {}, ef_construction = {});
+                "#,
+                self.config.m, self.config.efconstruction
+            );
+            warn!("Running query: {:?}", query);
```
diptanu, line 101: If we don't want to log these by default and meant for debugging lets use debug!()

src/vectordbs/pg_embedding.rs
diptanu: Let's remove it if we don't expect it to happen ever. Less code is better :D

diptanu (changes requested):
Some more changes please! But looking great.

src/vectordbs/pg_embedding.rs
yenicelik: i think with vector_dim > 2000 it still works, it will just be much slower because indexing is not on. do you want to throw an exception for postgres? it is a question of taste perhaps? can also throw an error as you like

src/vectordbs/pg_embedding.rs
yenicelik: in fact all queries are logged in debug!() i'll remove these

PR #123, opened 2023-10-06
Integrating OpenTelemetry

yenicelik:
- Tracing is also exported as part of the OpenTelemetry (otel) feed, using the OTLP protocol which is vendor-agnostic for metrics and traces. That means it is not outputted to stdout at the moment. grpc (tonic) is used to export the logs under the hood. 
- Docs added on how to read logs on jaeger
- Docs added on how to read metrics with prometheus and grafana
- Modified Dockerfile and added instructions on how to compile Opentelemetry with OpenSSL and g++
- Axum Opentelemetry middleware is added for (1) metrics (requests, errors, durations) and (2) logs
- Added #[instrument] decorators wherever a new span (for tracing) should be added. This helps us create flamegraphs and understand where time is most spent in the application. 
- #[derive(Debug)] is added to a number of structs to automatically display them in the logs when axum requests and responses are generated.
- a number of (minimally required) otel packages were added. `tracing-opentelemetry` and `tracing-subscriber` were added to output traces using OTLP
- Removed some unused packages, and cleaned up Cargo.toml to be more alphabetical.

Not covered by this PR: 
- common CPU, disk, network usage metrics. We can use the [metrics package](https://docs.rs/opentelemetry-system-metrics/latest/opentelemetry_system_metrics/), but the codebase does not seem to be extensive / super high quality


docs/docs/logs_and_metrics.md
```diff
@@ -0,0 +1,96 @@
...
+
+Indexify uses [OpenTelemetry](https://opentelemetry.io/) to bundle logs and metrics.
+All logs are exported using the unified [OTLP](https://opentelemetry.io/docs/specs/otel/protocol/). You can plug in any downstream data-collectors and visualizers
+
+## Logs
+
+Indexify exposes logs (traces) and metrics to manage the system better.
+
+It uses the [tracing](https://docs.rs/tracing/latest/tracing/) module which asynchronously (using tokio) produces logs.
+You can modify the log-leve using the RUST_LOG environment variable.
```
diptanu, line 11: log-level

docs/docs/logs_and_metrics.md
```diff
@@ -0,0 +1,96 @@
+# Logs and Metrics
+
+Indexify uses [OpenTelemetry](https://opentelemetry.io/) to bundle logs and metrics.
+All logs are exported using the unified [OTLP](https://opentelemetry.io/docs/specs/otel/protocol/). You can plug in any downstream data-collectors and visualizers
+
+## Logs
+
+Indexify exposes logs (traces) and metrics to manage the system better.
+
+It uses the [tracing](https://docs.rs/tracing/latest/tracing/) module which asynchronously (using tokio) produces logs.
```
diptanu, line 10: Let's remove this line, it's not very useful to operators reading the manual to learn about where and what is being logged.

docs/docs/logs_and_metrics.md
```diff
@@ -0,0 +1,96 @@
...
+You can modify the log-leve using the RUST_LOG environment variable.
+<!-- TODO: this should be set before Docker is run -->
+These traces are automatically displayed in stdout.
```
diptanu, line 13: Put this under a section #### Visualizing Logs on Jaeger or something similar.

docs/docs/logs_and_metrics.md
```diff
@@ -0,0 +1,96 @@
...
+As you make specific API calls to your indexify application (such as "localhost:8900/repositories"), traces are populated, which you can then view spans and details in Jaeger.
+
+![Traces](docs/docs/images/jaeger/traces.png)
+![Detailed Spans & Logs](docs/docs/images/jaeger/traces.png)
+![Statistics](docs/docs/images/jaeger/stats.png)
+![Flamegraph](docs/docs/images/jaeger/flamegraph.png)
+
+## Metrics
+
+The coordinator, executor and server individually expose metrics using [this package](https://github.com/ttys3/axum-otel-metrics).
```
diptanu, line 34: Why not use the open telemetry metrics package? https://docs.rs/opentelemetry_api/0.20.0/opentelemetry_api/metrics/index.html

docs/docs/logs_and_metrics.md
```diff
@@ -0,0 +1,96 @@
...
+
+To visualize these metrics, you can use any visualization library that can collect and parse OLTP data. In this example, we will use prometheus and grafana.
```
diptanu, line 36: Put this in a separate sub section

docs/docs/logs_and_metrics.md
```diff
@@ -0,0 +1,96 @@
+# Logs and Metrics
+
+Indexify uses [OpenTelemetry](https://opentelemetry.io/) to bundle logs and metrics.
+All logs are exported using the unified [OTLP](https://opentelemetry.io/docs/specs/otel/protocol/). You can plug in any downstream data-collectors and visualizers
+
+## Logs
+
+Indexify exposes logs (traces) and metrics to manage the system better.
```
diptanu, line 8: What is the otel service name we are using? Might be worth mentioning it here.

src/data_repository_manager.rs
```diff
@@ -41,6 +41,15 @@ pub struct DataRepositoryManager {
     blob_storage: BlobStorageTS,
 }
 
+impl fmt::Debug for DataRepositoryManager {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("DataRepositoryManager")
+            .field("repository", &self.repository)
```
diptanu, line 47: Remove the repository and vector_index_manager attributes from here. They are just manager classes and not really a 1-1 mapping between a data repository and a underlying index.

src/executor.rs
```diff
@@ -68,7 +72,21 @@ pub struct ExtractorExecutor {
...
 
+impl fmt::Debug for ExtractorExecutor {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("ExtractorExecutor")
+            .field("repository", &self.repository)
+            .field("config", &self.config)
+            .field("executor_id", &self.executor_id)
+            .field("extractor_info_list", &self.extractor_info_list)
+            .field("attribute_index_manager", &self.attribute_index_manager)
+            .field("work_store", &self.work_store)
```
diptanu, line 83: Remove the `work_store`, `attribute_index_manager` and `repository`

src/extractors.rs
```diff
@@ -122,6 +124,12 @@ pub struct PythonDriver {
     module_object: PyObject,
```
diptanu, line 124: We might be able to inject the name of the object into this while creating it. Can do in a followup pr if it's not stragihtforward.

src/main.rs
```diff
@@ -37,19 +44,44 @@ enum Commands {
...
+    let otlp_exporter = opentelemetry_otlp::new_exporter().tonic();
+    let tracer = opentelemetry_otlp::new_pipeline()
+        .tracing()
+        .with_exporter(otlp_exporter)
+        .with_trace_config(
+            opentelemetry::sdk::trace::config()
+                .with_resource(Resource::new(vec![KeyValue::new(
+                    "service.name",
+                    // TODO: @diptanu, we should probably set a service-name here, could strumify the Commands enum (i.e. make each option also return a string)
+                    "indexify-service",
```
diptanu, line 67: This should be in the switch statements below. So for server it should be indexify.api_server, for coordinator indexify.coordinator, and for the executors, indexify.executor

src/main.rs
```diff
@@ -37,19 +44,44 @@ enum Commands {
...
+                )]))
+                // TODO: In production, we can change this config
+                .with_sampler(opentelemetry::sdk::trace::Sampler::AlwaysOn),
```
diptanu, line 70: Yeah this should probably be in config.

src/main.rs
```diff
@@ -37,19 +44,44 @@ enum Commands {
...
+        )
+        .with_batch_config(opentelemetry::sdk::trace::BatchConfig::default())
+        .install_batch(opentelemetry_sdk::runtime::Tokio)?;
+    let otlp_layer = tracing_opentelemetry::layer().with_tracer(tracer);
+
+    // Hook it up to tracing
+    let subscriber = tracing_subscriber::registry().with(filter).with(otlp_layer);
+    tracing::subscriber::set_global_default(subscriber)?;
+
+    // TODO: We can add another global span here, before we call the individual functions
```
diptanu, line 80: Lets do it?

src/main.rs
```diff
@@ -37,19 +44,44 @@ enum Commands {
...
+    // This will be required to trace functions that are not explicity marked as #[instrument]
+
+    // Start application
+    info!("Spinning up");
```
diptanu, line 84: Is this intentional?

src/main.rs
```diff
@@ -99,5 +133,8 @@ async fn main() -> Result<(), Error> {
             executor_server.run().await?
         }
     }
+
+    // Is shutdown here? Should probably also be configured on drop?
+    global::shutdown_tracer_provider();
```
diptanu, line 138: Yeah this would be at the end. Individual await blocks the servers.

src/vector_index.rs
```diff
@@ -15,6 +16,14 @@ pub struct VectorIndexManager {
     vector_db: VectorDBTS,
 }
 
+impl fmt::Debug for VectorIndexManager {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("VectorIndexManager")
+            .field("repository", &self.repository)
```
diptanu, line 22: Remove the repository from here.

diptanu (changes requested):
Looking great, requesting for some small changes. By the way why are we not using this package for metrics? - https://docs.rs/opentelemetry_api/0.20.0/opentelemetry_api/metrics/index.html

docs/docs/logs_and_metrics.md
yenicelik: Mentioned it, maybe it's easiest if you reproduce the PR locally and modify the .md a bit more yourself

src/executor.rs
yenicelik: I will also remove the Debug decorators again to keep the codebase smaller (I had added them for the sole purpose of adding these here)

src/extractors.rs
yenicelik: hmm, seems like PyObject implements Debug. I don't see an easy way to make the debug only show the name of the object though (i dont know what will be displayed honestly). 
source code: https://docs.rs/pyo3/latest/src/pyo3/instance.rs.html#1033

we can also leave it like this for now and revisit as you mentioned

docs/docs/logs_and_metrics.md
yenicelik: Is this resolved? 

this package basically uses opentelemetry metrics to compute stats about rates, errors and durations. https://github.com/ttys3/axum-otel-metrics/blob/main/src/lib.rs#L88 so we would basically be re-implementing parts of this package. it may be worth re-implementing when we want more flexibility and more metrics, but right now this is probably overkill?

src/main.rs
yenicelik: turns out we should probably just control the sample-rate for now, we shouldn't make the config too complex: 
```
pub enum Sampler {
    /// Always sample the trace
    AlwaysOn,
    /// Never sample the trace
    AlwaysOff,
    /// Respects the parent span's sampling decision or delegates a delegate sampler for root spans.
    ParentBased(Box<dyn ShouldSample>),
    /// Sample a given fraction of traces. Fractions >= 1 will always sample. If the parent span is
    /// sampled, then it's child spans will automatically be sampled. Fractions < 0 are treated as
    /// zero, but spans may still be sampled if their parent is.
    /// *Note:* If this is used then all Spans in a trace will become sampled assuming that the
    /// first span is sampled as it is based on the `trace_id` not the `span_id`
    TraceIdRatioBased(f64),
    /// Jaeger remote sampler supports any remote service that implemented the jaeger remote sampler protocol.
    /// The proto definition can be found [here](https://github.com/jaegertracing/jaeger-idl/blob/main/proto/api_v2/sampling.proto)
    ///
    /// Jaeger remote sampler allows remotely controlling the sampling configuration for the SDKs.
    /// The sampling is typically configured at the collector and the SDKs actively poll for changes.
    /// The sampler uses TraceIdRatioBased or rate-limited sampler under the hood.
    /// These samplers can be configured per whole service (a.k.a default), or per span name in a
    /// given service (a.k.a per operation).
    #[cfg(feature = "jaeger_remote_sampler")]
    JaegerRemote(JaegerRemoteSampler),
}
```
TraceIdRatioBased(1.0) will correspond to AlwaysOn, so we can just accept a float in the config

src/main.rs
yenicelik: This may already be implemented, need to check this in the spans!

diptanu (approved):
LGTM. This is amazing, a great start towards an observable system!

PR #146, opened 2023-11-22
Add support for devcontainer (updated)

yenicelik:
Update devcontainer requirements to allow for a robust development environment. This should be accessible from github workspaces now too. This was tested on VSCode.


yenicelik:
@diptanu not the cleanest and shorted commit history (I could squash them if you think it makes sense) but I think this is fine for now. We can prob merge this first then the extractors (or implement extractors straight into https://github.com/tensorlakeai/indexify-extractors

PR #155, opened 2023-11-28
Fixes for sdk-py and update integration tests

lucasjacks0n:
Updated python sdk along with integration tests
- tests can be ran multiple times without failing, using uuid to avoid duplicate bindings
- updated sdk to comply with new schema
- added missing Extractor from_dict function

### DIFF
--- a/sdk-py/indexify/repository.py
+++ b/sdk-py/indexify/repository.py
@@ -111,7 +111,7 @@ class Repository:
 
         Args:
             - name (str): Name of the extractor.
-            - outputs (dict): Dictionary containing index names of outputs.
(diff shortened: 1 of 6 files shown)


sdk-py/indexify/repository.py
```diff
@@ -126,9 +128,10 @@ def bind_extractor(
         """
         req = {
             "extractor_name": extractor_name,
-            "index_name": index_name,
+            "index_names": {"embedding": index_name},
```
diptanu, line 131: Make the method take a dictionary of output name and index name

diptanu (changes requested):

diptanu (approved):

PR #160, opened 2023-12-01
Add langchain RAG example

lucasjacks0n:
This pr adds a langchain example imenting Retrieval-Augmented Generation with indexify retriever 
Based on https://towardsdatascience.com/retrieval-augmented-generation-rag-from-theory-to-langchain-implementation-4e9bd5f6a4f2

### DIFF
--- a/langchain/indexify_langchain/langchain_example.py
+++ b/langchain/indexify_langchain/langchain_example.py
@@ -1,0 +1,89 @@
+from langchain.document_loaders import TextLoader
+from langchain.text_splitter import CharacterTextSplitter
+import requests
+import dotenv
+
+# specify your OPENAI_API_KEY in .env file
+dotenv.load_dotenv()
+
+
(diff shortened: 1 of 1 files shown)


langchain/indexify_langchain/langchain_example.py
```diff
@@ -0,0 +1,89 @@
...
+
+with open("state_of_the_union.txt", "w") as f:
+    f.write(res.text)
+
+loader = TextLoader("./state_of_the_union.txt")
+documents = loader.load()
+
+
+# chunk text
+text_splitter = CharacterTextSplitter(chunk_size=500, chunk_overlap=50)
```
diptanu, line 22: We can move this to the extractor. When binding the extractor you can pass the splitter name, chunk size and such. Look at the base_embedding extractor class. So the extractor will chunk things for you!

diptanu (changes requested):

PR #161, opened 2023-12-01
Remove extractors from this repo

lucasjacks0n:
Moving forward our extractors will live inside of indexify-extractors


diptanu:
Looks good, remove this also https://github.com/diptanu/indexify/blob/main/run_tests.sh#L27

PR #162, opened 2023-12-01
BaseEmbeddingExtractor - Use params for overlap + chunk size

lucasjacks0n:
This pr updates BaseEmbeddingExtractor to use EmbeddingInputParams for chunk overlap size

### DIFF
--- a/indexify_extractor_sdk/base_embedding.py
+++ b/indexify_extractor_sdk/base_embedding.py
@@ -23,9 +23,7 @@ class BaseEmbeddingExtractor(Extractor):
     def extract(
         self, content_list: List[Content], params: EmbeddingInputParams
     ) -> List[List[Content]]:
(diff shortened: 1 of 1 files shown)


diptanu:
Looks good, just remove this also - https://github.com/diptanu/indexify/blob/main/run_tests.sh#L27

indexify_extractor_sdk/base_embedding.py
```diff
@@ -9,6 +9,8 @@
     Extractor,
     Feature,
 )
+
+
 class EmbeddingInputParams(BaseModel):
     overlap: int = 0
```
diptanu, line 15: Can we add a field chunk_size also? Right now we are hardcoding to the model context length. Lets use a default chunksize of whatever is the model context length. You can set this in the schema so users will know about it. And they can still override it by specifying a different chunk_size.

diptanu (changes requested):

PR #175, opened 2023-12-12
Updated version dependencies to be more specific

rakshith-ravi:
A lot of versions were using vague dependencies, which means `cargo update` would break a lot of things.

Made things a bit more specific, as well as moved all dependencies to the workspace level to make it easier to upgrade them all together.

Also, the migrations crate was using `async-std` for some reason, I moved it to tokio so that our build times go down as well

### DIFF
--- a/src/vectordbs/qdrant.rs
+++ b/src/vectordbs/qdrant.rs
@@ -119,12 +119,7 @@ impl VectorDb for QdrantDb {
         }
         let _result = self
             .create_client()?
-            .upsert_points(
-                &index,
-                None /* TODO: Somebody please check if this parameter is correct. Version upgrades required an extra parameter */,
-                points,
-                None
-            )
+            .upsert_points(&index, None, points, None)
             .await
             .map_err(|e| VectorDbError::IndexNotCreated(e.to_string()))?;
         Ok(())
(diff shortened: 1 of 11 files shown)


src/vectordbs/qdrant.rs
```diff
@@ -119,7 +119,12 @@ impl VectorDb for QdrantDb {
         }
         let _result = self
             .create_client()?
-            .upsert_points(&index, points, None)
+            .upsert_points(
+                &index,
+                None /* TODO: Somebody please check if this parameter is correct. Version upgrades required an extra parameter */,
```
rakshith-ravi, line 124: Need an extra set of eyes on this

src/vectordbs/qdrant.rs
```diff
@@ -181,7 +186,9 @@ impl VectorDb for QdrantDb {
         let collection_info = result
             .result
             .ok_or(VectorDbError::IndexNotRead("index not found".into()))?;
-        Ok(collection_info.points_count)
+        // TODO: check if this is correct. The new version returns an optional value instead of a u64
```
rakshith-ravi, line 189: And this as well

rakshith-ravi (commented):
The version upgrade required these two specific code changes that I'm not sure is the right approach.

Would like to bring this to your attention so that somebody can check me on this

PR #176, opened 2023-12-12
Refactor the way the CLI works

rakshith-ravi:
Added nightly rustfmt for formatting.
A lot of rustfmt changes and clippy changes for a more readable code.

The biggest change would be the way the CLI works and how the execution is done. Most of the changes brings in separation of concerns so that each individual part of the CLI can be changed without affecting anything else

As always, I've run `./run_tests.sh` along with rebuilding the devcontainer without cache (since I touched the devcontainer stuff as well) and they all run without any issues.


rakshith-ravi:
PS - in some cases, you'd notice that I've removed `help = ""` from clap and added docs. Clap automatically considers docs as the help text

diptanu (approved):
LGTM! Thanks for the PR, please separate functional and structural/formatting changes in the future :)

PR #177, opened 2023-12-12
Fixed a small typo

rakshith-ravi:
That note should've been for devs. Adding it as a doc would mean it'll show up in the CLI. I've changed it to a comment, so that it doesn't show up in the CLI

### DIFF
--- a/src/cmd/mod.rs
+++ b/src/cmd/mod.rs
@@ -10,9 +10,8 @@ mod server;
 #[derive(Debug, Args)]
 pub struct GlobalArgs {
     /// how verbose the logging should be
-    ///
-    /// Currently each command has it's own stuff. I'm just keeping this here as
-    /// an example. We can iterate on this as and when we need to.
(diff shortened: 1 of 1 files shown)


src/cmd/mod.rs
```diff
@@ -10,9 +10,8 @@ mod server;
 #[derive(Debug, Args)]
 pub struct GlobalArgs {
     /// how verbose the logging should be
-    ///
-    /// Currently each command has it's own stuff. I'm just keeping this here as
-    /// an example. We can iterate on this as and when we need to.
+    // Currently each command has it's own stuff. I'm just keeping this here as
```
diptanu, line 13: Understood, let's remove the comment.

diptanu (changes requested):

src/cmd/mod.rs
rakshith-ravi: Done

PR #187, opened 2023-12-19
Return index name after creating an extractor.

stangirala:
Return index name after creating an extractor.

For https://github.com/diptanu/indexify/issues/185


diptanu (approved):
Love it! Thanks <3

PR #193, opened 2023-12-21
makefile install

lucasjacks0n:
This pr adds install target which depends on build-release

to run:
```bash
sudo make install
```

this will build-release and install to /usr/local/bin/$(APPLICATION_NAME)


diptanu (approved):
This is great! No more ./debug/target/indexify :D

PR #194, opened 2023-12-21
WIP add extractor cache path on host

stangirala:


src/cmd/extractor/extract.rs
```diff
@@ -7,6 +7,9 @@ pub struct Args {
     #[arg(short = 'e', long)]
     extractor_path: Option<String>,
 
+    #[arg(short = 'c', long)]
+    extractor_cache_path: Option<String>,
```
diptanu, line 11: cache_dir

src/cmd/extractor/extract.rs
```diff
@@ -7,6 +7,9 @@ pub struct Args {
     #[arg(short = 'e', long)]
     extractor_path: Option<String>,
 
+    #[arg(short = 'c', long)]
```
diptanu, line 10: Lets use only long? I think we use -c as global flag for config

src/cmd/extractor/extract.rs
```diff
@@ -21,13 +24,14 @@ impl Args {
...
             extractor_path,
+            extractor_cache_path,
             name,
             text,
             file,
         } = self;
 
         if let Some(name) = name {
-            let _ = crate::extractor::run_docker_extractor(name, text, file)
+            let _ = crate::extractor::run_docker_extractor(name, extractor_path, text, file)
```
diptanu, line 34: Shouldn't you be passing the cache path here?

src/extractor/mod.rs
```diff
@@ -97,22 +98,36 @@ pub async fn run_docker_extractor(
...
             target: Some(format!("/indexify/{}", file_name)),
             source: Some(file_path.display().to_string()),
             typ: Some(bollard::service::MountTypeEnum::BIND),
             ..Default::default()
         }];
+
+        let extractor_cache_path = Path::new(&extractor_cache_path).canonicalize().unwrap();
+        let extractor_cache_path = extractor_cache_path.file_name().unwrap().to_str().unwrap();
+        args.push("--extractor-cache-path".to_string());
+        args.push(extractor_cache_path.to_string());
```
diptanu, line 123: Do we assume there is always a cache dir being passed? I think it should be optional.

diptanu (changes requested):
Looks great overall, have some questions and comments.

src/cmd/extractor/extract.rs
stangirala: yes, will fix.

src/extractor/mod.rs
stangirala: I guess we could make it optional, but should we create a cache path for all extractors on the host by default?

src/extractor/mod.rs
```diff
@@ -97,33 +98,59 @@ pub async fn run_docker_extractor(
...
             typ: Some(bollard::service::MountTypeEnum::BIND),
             ..Default::default()
-        }];
-        host_config.replace(HostConfig {
-            mounts: Some(mounts),
+        });
+    }
+
+    if let Some(cache_dir) = cache_dir {
+        let cache_dir = Path::new(&cache_dir).canonicalize().unwrap();
```
diptanu, line 125: We need to remove unwraps() from everywhere in code but we can do that in a subsequent PR. This is not just here but I have written a bunch of unwraps in other places which needs to be gone.

diptanu (approved):
LGTM! One small comment and that's about it. Also if you can rename CACHE_DIR to INDEXIFY_CACHE_DIR that would be great before merging.

diptanu:
Looks great!

PR #200, opened 2024-01-03
Add pytorch base image for extractors via indexify.yaml

stangirala:
Add pytorch base image for extractors via indexify.yaml. Extractor config has to specify `torch`, `cuda` and `cudann` version in the yaml, see tests for example.

### DIFF
--- a/src/package.rs
+++ b/src/package.rs
@@ -54,7 +54,7 @@ impl Packager {
 
     pub async fn package(&self, verbose: bool) -> Result<()> {
         let docker_file = self.create_docker_file()?;
-        println!("{}", docker_file);
+        info!("{}", docker_file);
         let mut header = tar::Header::new_gnu();
         header.set_path("Dockerfile").unwrap();
         header.set_size(docker_file.len() as u64);
@@ -157,18 +157,17 @@ impl Packager {
 
         match (pytorch_version, gpu) {
             (Some(version), true) => {
-                if version == "latest" || cuda_version == "" || cuda_nn == "" {
+                if version == "latest" || cuda_version.is_none() || cuda_nn.is_none() {
                     image_name = "pytorch/pytorch:latest".to_string();
                 } else {
-                    image_name = format!("pytorch/pytorch:{}-cuda{}-cudnn{}-runtime", version, cuda_version, cuda_nn);
+                    image_name = format!("pytorch/pytorch:{}-cuda{}-cudnn{}-runtime", version, cuda_version.unwrap(), cuda_nn.unwrap());
                 }
             },
             (Some(_version), false) => {
                 image_name = "tensorlake/indexify-extractor-base".to_string();
                 additional_pip_flags = "--index-url https://download.pytorch.org/whl/cpu";
             },
-            (None, true) => { image_name = "tensorlake/indexify-extractor-base".to_string(); },
-            (None, false) => { image_name = "tensorlake/indexify-extractor-base".to_string(); },
+            (None, _) => { image_name = "tensorlake/indexify-extractor-base".to_string(); },
         }
 
         let system_dependencies = self.config.system_dependencies.join(" ");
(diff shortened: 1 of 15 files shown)


src/package.rs
```diff
@@ -52,6 +54,7 @@ impl Packager {
 
     pub async fn package(&self, verbose: bool) -> Result<()> {
         let docker_file = self.create_docker_file()?;
+        println!("{}", docker_file);
```
diptanu, line 57: Remove this ?

src/server_config.rs
```diff
@@ -131,6 +131,12 @@ pub struct ExtractorConfig {
     pub gpu: bool,
     pub system_dependencies: Vec<String>,
     pub python_dependencies: Vec<String>,
+
+    #[serde(default)]
+    pub cuda_version: String,
+
+    #[serde(default)]
+    pub cudann_version: String,
```
diptanu, line 139: Probably best if you used Option<String> instead of denoting "" as unspecified.

src/package.rs
```diff
@@ -130,6 +133,44 @@ impl Packager {
...
+                    image_name = "pytorch/pytorch:latest".to_string();
+                } else {
+                    image_name = format!("pytorch/pytorch:{}-cuda{}-cudnn{}-runtime", version, cuda_version, cuda_nn);
+                }
+            },
+            (Some(_version), false) => {
+                image_name = "tensorlake/indexify-extractor-base".to_string();
+                additional_pip_flags = "--index-url https://download.pytorch.org/whl/cpu";
+            },
+            (None, true) => { image_name = "tensorlake/indexify-extractor-base".to_string(); },
```
diptanu, line 170: I think you could club the two statements here by something like this but I am not sure - 

```
(None, _) => {image_name = "....";},

File: src/package.rs [hunk 4]
```diff
@@ -130,6 +133,44 @@ impl Packager {
...
+            None => {}
+        }
+
+        let mut additional_pip_flags = "";
+        let image_name:String;
+
+        match (pytorch_version, gpu) {
+            (Some(version), true) => {
+                if version == "latest" || cuda_version == "" || cuda_nn == "" {
+                    image_name = "pytorch/pytorch:latest".to_string();
```
diptanu, line 161: How does the indexify binary get into this PyTorch base image? The indexify extractor base image has the indexify binary on top of a ubuntu base image.

diptanu (changes requested):
Looks great! Some minor style related comments and a question about how does the indexify binary make it's way into the PyTorch base image?

src/package.rs
stangirala: Changed to info, just for logging.

src/package.rs
```diff
@@ -130,6 +133,43 @@ impl Packager {
...
+                    pytorch_version = Option::Some("latest");
+                }
+            },
+            None => {}
+        }
+
+        let mut additional_pip_flags = "";
+        let image_name:String;
+
+        match (pytorch_version, gpu) {
```
jackbackes, line 158: The logic for creating the image name is fairly complex. I'm imagining a scenario where an invalid image name is being generated but it's hard to debug. At minimum I think this logic should be extracted and separately tested. You may also want to consider a more structured approach.

src/package.rs
```diff
@@ -130,6 +133,43 @@ impl Packager {
...
+            (Some(version), true) => {
+                if version == "latest" || cuda_version.is_none() || cuda_nn.is_none() {
+                    image_name = "pytorch/pytorch:latest".to_string();
+                } else {
+                    image_name = format!("pytorch/pytorch:{}-cuda{}-cudnn{}-runtime", version, cuda_version.unwrap(), cuda_nn.unwrap());
+                }
+            },
+            (Some(_version), false) => {
+                image_name = "tensorlake/indexify-extractor-base".to_string();
```
jackbackes, line 167: Hardcoded values like "pytorch/pytorch:latest" and "tensorlake/indexify-extractor-base" could limit flexibility. You might consider external configuration here. Also it might make sense to generate these as lazy_static globals that can be reused throughout the application.

src/package.rs
```diff
@@ -52,6 +54,7 @@ impl Packager {
 
     pub async fn package(&self, verbose: bool) -> Result<()> {
         let docker_file = self.create_docker_file()?;
+        info!("{}", docker_file);
```
jackbackes, line 57: this could potentially log sensitive information.

src/package.rs
stangirala: Left a TODO comment here, will need to follow up

diptanu:
Thanks @stangirala for this!

Issue #201, opened 2024-01-03
Missing step in setup docs


jackbackes:
In [initial dev setup](https://getindexify.ai/develop/#rust-compiler), `cargo build` fails if the repo is not setup with the rust "nightly" toolchain.

jackbackes:
Can't reproduce.

PR #202, opened 2024-01-03
Added nightly toolchain step to setup docs

jackbackes:
Minor change to address #201 by adding a step to developer setup to install the rust nightly toolchain.

### DIFF
--- a/docs/docs/develop.md
+++ b/docs/docs/develop.md
@@ -11,6 +11,13 @@ Install various rust related tools -
 * Rust Compiler - <http://rustup.rs>
 * Cargo tools - clippy, rustfmt is very helpful for formating and fixing warnings.
 
+Install the "nightly" toolchain -
+
+```shell
+rustup update nightly
+rustup override set nightly
+```
+
 ### Python Dependencies
 
 Create a virtual env


docs/docs/develop.md
```diff
@@ -11,6 +11,13 @@ Install various rust related tools -
 * Rust Compiler - <http://rustup.rs>
 * Cargo tools - clippy, rustfmt is very helpful for formating and fixing warnings.
 
+Install the "nightly" toolchain -
+
+```shell
+rustup update nightly
+rustup override set nightly
```
diptanu, line 18: Do we need this? We don't depend on anything from the nightly other than formatting stuff. 

I have a make target in the Makefile which does cargo @nightly fmt or something like that which triggers the nightly toolchain only formatting.

diptanu (changes requested):
One small comment.

docs/docs/develop.md
jackbackes: I ran into this issue when doing a fresh build for development setup. There were unsupported dependencies that required the nightly toolchain. IIRC it had to do with sea-orm.

I'm rebuilding it fresh so I'll see if I run into the issue again.

jackbackes: I'm going to close this as "can't reproduce." I reinstalled and ran Indexify in a fresh environment and it worked fine. Maybe I had an out of date version of rustc. That will teach me to not use the dev container. ¯\_(ツ)_/¯

jackbackes:
Can't reproduce.

PR #214, opened 2024-01-11
Add caching modules and dependencies

jackbackes:
- Added Moka and Redis dependencies in Cargo.toml
- Created caching module with submodules for Moka, Redis, and NoOp caches
- Implemented caching traits and asynchronous operations for cache types


src/extractor_router.rs
```diff
@@ -13,6 +13,8 @@ pub struct ExtractorRouter {
     coordinator_client: Arc<CoordinatorClient>,
 }
 
+
```
diptanu, line 16: Are these from formatting?

src/caching/redis_cache.rs
```diff
@@ -0,0 +1,105 @@
...
+        let key = serde_json::to_string(key)?;
+        redis::cmd("DEL")
+            .arg(key)
+            .query_async(&mut conn)
+            .await?;
+        Ok(())
+    }
+}
+
+// TODO: integrate Redis into local dev setup so we can run this test
```
diptanu, line 69: Should be super easy, can we just add the dependency in Makefile and uncomment these before merging?

diptanu (changes requested):
Minor comments, looks good overall. I am not very happy with encoding the key as json but I understand why it had to be done. I wonder if there is a faster serialization/de-serilazation. I wonder if Avro would be a better fit than json for this use case, we can still use serde to move back and forth between the structs and serialized forms - https://docs.rs/avro-rs/latest/avro_rs/

src/extractor_router.rs
jackbackes: Ran make fmt so it has standard formatting now.

src/caching/redis_cache.rs
jackbackes: Added Redis. Test now passes if Redis is running!

PR #221, opened 2024-01-13
Leader election

diptanu:
* Hooked up scheduler with Raft's leader election process 
* Scheduler runs on the leader now now, but we can run this in parallel across all nodes in future for scale out.
* Scheduling is not yet reactive

### DIFF
--- a/src/state/mod.rs
+++ b/src/state/mod.rs
@@ -2,6 +2,7 @@
 #![deny(unused_qualifications)]
 
 use std::{
+    cell::RefCell,
     collections::{BTreeMap, HashMap},
     io::Cursor,
     sync::Arc,
@@ -13,7 +14,10 @@ use network::Network;
 use openraft::{self, storage::Adaptor, BasicNode};
 use store::{Request, Response, Store};
 use tokio::{
-    sync::{watch, Mutex},
+    sync::{
+        watch::{self, Receiver, Sender},
+        Mutex,
+    },
     task::JoinHandle,
 };
 use tracing::{error, info};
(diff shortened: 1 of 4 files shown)


src/state/mod.rs
```diff
@@ -534,3 +552,46 @@ impl App {
...
+    let mut rx = raft.metrics();
+    let prev_server_state = RefCell::new(openraft::ServerState::Learner);
+
+    loop {
+        tokio::select! {
+                    _ = shutdown_rx.changed() => {
+                        info!("shutting down leader change watcher");
+                        return Ok(());
+                    }
+                    _ = rx.changed() => {
```
drmingdrmer, line 571: `rx.change()` returns an error when the sender is dropped. You need to handle such errors:

https://github.com/datafuselabs/openraft/blob/9f5a6955d69aedb5f6635807e274f83955acdc83/openraft/src/metrics/wait.rs#L92-L110

And in this way, you do not need to watch the `shutdown_rx` any more: when your application shutdown, the sender is dropped and `rx.changed()` will be notified.

src/state/mod.rs
```diff
@@ -534,3 +552,46 @@ impl App {
...
+                        let server_state = rx.borrow_and_update().state.clone();
+                        let mut prev_srvr_state = prev_server_state.borrow_mut();
+                        if !(prev_srvr_state).eq(&server_state) {
+                            info!("raft change metrics prev {:?} current {:?}", prev_srvr_state, server_state);
+                            // If the prev state was leader and the current state is non leader send a
+                            // change notification to indicate that we lost leadership
+
+                            if prev_srvr_state.is_leader() && !server_state.is_leader() {
+                                info!("node {} lost leadership", node_id);
+                            leader_change_tx.send(false.into()).unwrap();
```
drmingdrmer, line 581: The indent is incorrect.

`unwrap()` should be avoided in a production codebase, unless it is proved to be safe. In this case, if the `Receiver` end is closed, this line will panic.

PR #223, opened 2024-01-15
Add metrics to coordinator

stangirala:
Add metrics using autometrics and expose a prometheus format `/metrics` endpoint.


maxkozlovsky:
Added metrics through a service layer.

PR #227, opened 2024-01-17
Updated links to new repo name

rakshith-ravi:


rakshith-ravi:
Looks like vercel will also have to be reconfigured to use the new repo name. Sucks that they don't pick it up automatically. @diptanu can you update that?

PR #230, opened 2024-01-17
Udate k8 deployments 

lucasjacks0n:
This pr adds coordinator deployment and minilm-l6 extractor


diptanu:
Looks good, please remove the static port definitions

PR #243, opened 2024-01-21
Made all reads parallel instead of sequential

rakshith-ravi:
The way we're doing `BlobStorageBuilder` currently restricts things to bring it into the trait itself, so this is a temporary fix for now. Once we have more time I'll restructure the code to include it into the trait itself

### DIFF
--- a/src/blob_storage/mod.rs
+++ b/src/blob_storage/mod.rs
@@ -1,100 +1,180 @@
-use std::sync::Arc;
+use std::fmt::{self, Debug, Formatter};
 
-use anyhow::{anyhow, Result};
+use anyhow::{Context, Result};
 use async_trait::async_trait;
 use bytes::Bytes;
 use object_store::aws::AmazonS3Builder;
-
-use self::s3::{S3Reader, S3Storage};
-use crate::server_config::BlobStorageConfig;
+use serde::{Deserialize, Serialize};
 
 pub mod disk;
 pub mod s3;
 
-pub type BlobStorageTS = Arc<dyn BlobStorage + Sync + Send>;
+#[derive(Debug, Clone, Serialize, Deserialize)]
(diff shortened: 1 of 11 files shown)


src/data_repository_manager.rs
```diff
@@ -290,16 +291,24 @@ impl DataRepositoryManager {
...
+                    content_type: content.mime,
+                    bytes,
+                    labels: content.labels.clone(),
+                    feature: None,
+                };
+                Ok(content)
+            }));
+        }
+        for task in tasks {
+            let content = task.await.unwrap()?;
```
diptanu, line 311: Let's remove the unwraps here, and use futures::future::join_all to await for completion of all the tasks.

diptanu (changes requested):

src/data_repository_manager.rs
rakshith-ravi: These unwraps are for when the task panics, which it never will. The `?` after that is for the error from task output itself

src/blob_storage/mod.rs
```diff
@@ -1,100 +1,180 @@
...
+            return Ok(vec![]);
+        };
+
+        if key.starts_with("s3://") {
+            let (bucket, key) = parse_s3_url(key)
+                .map_err(|err| anyhow::anyhow!("unable to parse s3 url: {}", err))?;
+            return s3::S3Storage::new(
+                key,
+                AmazonS3Builder::from_env()
+                    .with_bucket_name(bucket)
```
diptanu, line 83: Add .with_region(..) here since the control plane or executors could be running in regions which is not where the bucket is at.

diptanu (changes requested):

rakshith-ravi:
Rebased with main

PR #245, opened 2024-01-22
Added labels dict to Content

mohit-raghavendra:
Formatted the python files according to pep8 using black


mohit-raghavendra:
Formatted the python files according to pep8 using [black]([https://github.com/psf/black])

PR #247, opened 2024-01-23
Labels sent to rust + test

mohit-raghavendra:
Updated to take labels from python object and send it to rust.

### DIFF
--- a/src/extractor/py_extractors.rs
+++ b/src/extractor/py_extractors.rs
@@ -31,6 +31,9 @@ pub struct PyContent {
     pub data: Vec<u8>,
     #[pyo3(get, set)]
     pub feature: Option<PyFeature>,
+    #[pyo3(get, set)]
+    pub labels: Option<HashMap<String, String>>,
+
 }
 
 impl TryFrom<PyContent> for internal_api::Content {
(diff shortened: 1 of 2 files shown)


src/extractor/py_extractors.rs
```diff
@@ -244,10 +260,21 @@ mod tests {
...
+
+        let test_labels1 = Some(HashMap::from([
+            ("url".to_string(), "test.com".to_string()),
+        ]));
+        let test_labels2 = Some(HashMap::from([
+            ("url".to_string(), "test.com".to_string()),
+        ]));
+        let content1 = PyContent::new(
+            "My name is Donald and I live in Seattle".to_string(),
+            test_labels1.clone())
```
diptanu, line 272: The labels in this test should come from the extractor output, no point in passing labels into the extractor during extraction.

diptanu (changes requested):

src/extractor/py_extractors.rs
```diff
@@ -75,16 +78,18 @@ impl TryFrom<internal_api::Content> for PyContent {
             content_type,
             data,
             feature: None,
+            labels: None,
```
diptanu, line 81:
```
labels: Some(content.labels)
```

diptanu (changes requested):

PR #266, opened 2024-01-26
Add MIME Type Validation in Extractor

jackbackes:
Implemented MIME type validation in the extractor module. This update introduces a flag in the extractor command to choose whether to throw an error for unmatched MIME types. The default behavior is to error on unmatched types. The extractor runner now filters content by MIME type, ensuring only supported types are processed. This enhancement improves the reliability of the extraction process by preventing unexpected behavior due to unsupported content types.

Key Changes:

- Added `err_for_unmatched_mimetype` flag in the `extract.rs` cli command, for opting into MIME type validation.
- Implemented `matches_mime_type` method in `Extractor` trait to check if the extractor supports the content's MIME type.
- Enhanced `ExtractorRunner` to filter content based on MIME type compatibility.
- Included MIME type validation logic in relevant modules (`executor.rs`, `executor_server.rs`, and `extractor_runner.rs`).

Note:
- This change ensures better alignment of content types with extractor capabilities, so that extractors aren't run on unsupported content.
- Tests were added to include scenarios for MIME type validation.

### DIFF
--- a/docs/docs/getting_started.md
+++ b/docs/docs/getting_started.md
@@ -271,7 +271,7 @@ Now let's rerun our query for documents related to "sports":
     {'text': 'Indexify is a retrieval service for LLM agents!', 'confidence_score': -0.012608046, 'metadata': {}}
     ```
 
-We can see the new document we added about Steph Curry is now included in the search results. Indexify automatically ran our extractors when we added the new document and updated the relevant indexes.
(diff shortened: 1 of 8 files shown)


diptanu (changes requested):
As discussed let's do this on the coordinator and not on the data plane to make scheduling centralized and not needing to change the data plane.

diptanu (approved):

PR #280, opened 2024-01-30
Added workflow to publish to dockerhub for x86 and Arm64

rakshith-ravi:
@lucasjacks0n can you make sure I'm not doing something ignorant here?


rakshith-ravi:
Just want another pair of eyes on this

PR #282, opened 2024-01-31
Add TaskAllocator and integrate with Coordinator

jackbackes:
This commit introduces the TaskAllocator, a dedicated module for task allocation in the Coordinator. Key aspects of this integration include:

- **Centralizing Task Allocation**: The TaskAllocator encapsulates task allocation logic, which was previously part of the Coordinator. This separation of concerns allows for a more focused and manageable approach to task distribution.

- **TaskAllocator Design Rationale**: The TaskAllocator was designed to provide a clear and isolated domain for managing task allocation. This isolation makes it easier to modify and extend task allocation strategies without impacting the broader system.

- **Load-Balancing Allocation**: Introduced load-balancing allocation, which will assign tasks to executors that rank lower for load.


src/coordinator.rs
```diff
@@ -204,6 +188,20 @@ impl Coordinator {
     }
 
     pub async fn remove_executor(&self, executor_id: &str) -> Result<()> {
+        // before removing the executor, evacuate and reassign its tasks
+        if let Err(err) = self
+            .get_task_allocator()
+            .evacuate_executor_and_reallocate_tasks(&executor_id)
```
diptanu, line 194: This will run the scheduler in the middle of the health check handler breaking because an executor dropped out. I think this makes the process of finishing the request handler in-deterministic since we don't know how long the scheduler will take to return. I would suggest removing the executor from shared state and create some event which triggers the scheduler asynchronously.

src/state/mod.rs
```diff
@@ -333,6 +334,23 @@ impl App {
         Ok(matched_content_list)
     }
 
+    pub async fn all_tasks(&self) -> Result<Vec<internal_api::Task>> {
+        let store = self.store.state_machine.read().await;
```
diptanu, line 338: TODO - We should look into having a copy on write reference here to avoid copies

src/state/mod.rs
```diff
@@ -333,6 +334,23 @@ impl App {
...
+
+    pub async fn get_tasks_by_ids(
+        &self,
+        task_ids: Vec<TaskId>,
+    ) -> Result<Vec<internal_api::Task>, anyhow::Error> {
+        Ok(self
+            .all_tasks()
+            .await?
+            .into_iter()
+            .filter(|task| task_ids.contains(&task.id))
```
diptanu, line 350: Convert task_ids to hash set if you are going to do contains so that it's an O(1) lookup vs O(n)

src/state/store/mod.rs
```diff
@@ -55,7 +57,10 @@ pub enum Request {
         tasks: Vec<internal_api::Task>,
     },
     AssignTask {
-        assignments: HashMap<TaskId, ExecutorId>,
+        assignments: TaskToExecutorAllocations,
+    },
+    EvacuateExecutor {
```
diptanu, line 62: What is this for?

diptanu (changes requested):
Looked at the first couple of files in the PR, have some minor comments and some questions.

src/task_allocator/mod.rs
```diff
@@ -0,0 +1,227 @@
+use std::{
+    borrow::BorrowMut,
+    collections::{BTreeSet, HashMap, HashSet},
+};
+
+use indexify_internal_api as internal_api;
+
+use self::work_planner::ExecutorIdRef;
```
diptanu, line 8: Rename to planner

src/coordinator.rs
jackbackes: Reallocations are now reactive to state changes.

src/state/mod.rs
jackbackes: I played with adding Cow in the ChangeType::NewBinding handler. It's a little clunky, but there may be ways to improve it.

src/state/mod.rs
jackbackes: fixed

src/state/store/mod.rs
jackbackes: removed.

src/task_allocator/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+            .commit_distribution_plan_to_state(plan)
+            .await?;
+        Ok(())
+    }
+}
+
+/// Facade pattern for accessing the shared state, to prevent too much
+/// complexity from leaking between the scheduler and the state machine.
+#[derive(Clone)]
+pub struct SchedulerStateFacade {
```
diptanu, line 106: I don't think this would be very useful. We are not building a general purpose database here on top of the shared state for a separation of application level query layer which does filtering and such. If you find doing some filtering in the scheduler, I would suggest just adding a reverse index in the storage layer to avoid that.

src/task_allocator/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+        self.shared_state
+            .store
+            .state_machine
+            .read()
+            .await
+            .extractor_executors_table
+            .clone()
+    }
+
+    async fn get_task_ids_for_executor(
```
diptanu, line 159: This should be in the shared state api.

src/task_allocator/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+            .shared_state
+            .tasks_for_executor(executor_id)
+            .await?
+            .into_iter()
+            .map(|task: internal_api::Task| task.id)
+            .collect::<Vec<TaskId>>();
+        Ok(tasks)
+    }
+
+    async fn get_unassigned_task_ids(&self) -> Result<Vec<TaskId>, anyhow::Error> {
```
diptanu, line 173: Same here and all the methods below.

src/task_allocator/planner/round_robin.rs
```diff
@@ -0,0 +1,75 @@
...
+        Self { shared_state }
+    }
+}
+
+#[async_trait::async_trait]
+impl AllocationPlanner for RoundRobinDistributor {
+    async fn plan_allocations(
+        &self,
+        task_ids: Vec<TaskId>,
+        ignore_executors: Vec<ExecutorIdRef<'_>>,
```
diptanu, line 26: What is this for?

src/task_allocator/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+    coordinator_filters::matches_mime_type,
+    state::{self, store::TaskId},
+};
+
+pub mod planner;
+
+pub type ExtractorExecutorsMap = HashMap<String, HashSet<String>>;
+
+pub struct TaskAllocator {
+    work_distributor: Box<dyn planner::AllocationPlanner + Send + Sync>,
```
diptanu, line 16: placement_planner or just task_planner.

src/state/store/mod.rs
```diff
@@ -120,7 +122,10 @@ pub type ExtractorName = String;
 #[derive(Clone)]
 pub enum ChangeType {
     NewContent,
-    NewBinding,
+    /// record repository_id because we need it to find the binding
+    NewBinding {
```
diptanu, line 126: Bindings should have unique id but it's fine to capture the repository id.

src/state/store/mod.rs
```diff
@@ -489,6 +498,16 @@ impl RaftStorage<TypeConfig> for Arc<SledStore> {
                                 .or_default();
                             executors.remove(executor_meta.extractor.name.as_str());
                         }
+                        // unassign any tasks assigned to executor, and add them to unassigned tasks
+                        if let Some(assigned_tasks) = sm.task_assignments.remove(executor_id) {
```
diptanu, line 502: Don't do this here, just mark that the executor is removed as a state change and remove the executor from all indexes. Change the task assignments in the scheduler

src/state/mod.rs
diptanu: Yeah Cow is not very ergonomic in Rust. Let's think about Cow in subsequent PRs, there are some ways to go. After this, we have to revisit how state is stored in disk and then get to the Cow work.

src/task_allocator/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    /// Reschedule all tasks that match the content types, even if
+    /// they are already assigned to a different executor.
+    /// This is useful for example when a new binding is added to an executor
+    pub async fn reallocate_all_tasks_matching_content_types(
+        &self,
+        content_types: HashSet<String>,
+    ) -> Result<(), anyhow::Error> {
+        let all_tasks = self.shared_state.get_all_task_ids().await?;
```
diptanu, line 84: Couple of things - 1. If we need all the tasks, a method which does that is more efficient than reading task ids first and then getting all the tasks. 2. Second this would be much better if there was a reverse index of content type and HashSet of Tasks vs reading all the tasks and filtering them here.

src/task_allocator/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+            .plan_allocations(unassigned_tasks, vec![])
+            .await?;
+        self.shared_state
+            .commit_distribution_plan_to_state(plan)
+            .await?;
+        Ok(())
+    }
+
+    /// Reschedule all tasks, even if they are already assigned to an executor.
+    pub async fn rebalance_all_task_allocations(&self) -> Result<(), anyhow::Error> {
```
diptanu, line 45: This looks not that efficient, since  there is no point in rebalancing tasks when executors with extractors not matching them are added. I would suggest making this method accept an extractor name, and then find all the tasks for that extractor and rebalancing only them.

src/task_allocator/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+        Self {
+            work_distributor: Box::new(planner::round_robin::RoundRobinDistributor::new(
+                state_facade.clone(),
+            )),
+            shared_state: state_facade.clone(),
+        }
+    }
+
+    /// Schedule all tasks that are not currently allocated to an executor
+    pub async fn schedule_unallocated_tasks(&self) -> Result<(), anyhow::Error> {
```
diptanu, line 32: Same here as below, take an extractor as an argument and only allocate tasks that can be allocated vs getting all the unassigned tasks.

src/task_allocator/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+        &self,
+        executor_id: ExecutorIdRef<'_>,
+    ) -> Result<(), anyhow::Error> {
+        let task_ids = self
+            .shared_state
+            .get_task_ids_for_executor(executor_id)
+            .await?;
+        let plan = self
+            .work_distributor
+            .plan_allocations(task_ids, vec![executor_id])
```
diptanu, line 69: There is no need for this vector here, if the executor is just removed from the shared state before the scheduler is invoked.

diptanu (changes requested):
Left some comments related to the task allocator.

src/task_allocator/mod.rs
jackbackes: this was actually older code, now removed.

src/task_allocator/mod.rs
jackbackes: makes sense - adding a reverse lookup by content type.

src/task_allocator/mod.rs
jackbackes: hmm... the purpose of this method is to allocate tasks after content is added or after an executor has been removed (and tasks have been de-allocated). We have a separate method for allocating tasks matching an executor, which is used when a new binding or executor is added.

src/task_allocator/mod.rs
jackbackes: ack

src/coordinator.rs
```diff
@@ -335,6 +312,100 @@ impl Coordinator {
...
+                        executor_id, err
+                    );
+                }
+            }
+            ChangeType::ExecutorRemoved => {
+                // tasks were unallocated from the executor when it was removed
+                // now we need to reassign them to other executors
+                let executor_id = state_change.id;
+                // unassign any tasks assigned to the executor
+                if let Some(unassigned_tasks) = sm.task_assignments.remove(executor_id.as_str()) {
```
diptanu, line 358: These mutations on state machine directly outside of the state store shouldn't be here. Because the writes are happening to the Raft state under the hood, they needs to be requests to Raft for mutations. This will simply not work. The comment is for your information.

diptanu (changes requested):

Cargo.toml
```diff
@@ -140,6 +141,7 @@ itertools = { workspace = true }
 jsonschema = { workspace = true }
 mime = { workspace = true }
 mime_guess = { workspace = true }
+mockall = { workspace = true }
```
diptanu, line 144: Can this be included only for tests?

src/state/mod.rs
```diff
@@ -371,6 +373,24 @@ impl App {
...
+        Ok(store.tasks.values().cloned().collect_vec())
+    }
+
+    pub async fn get_tasks_by_ids(
+        &self,
+        task_ids: HashSet<TaskId>,
+    ) -> Result<Vec<internal_api::Task>, anyhow::Error> {
+        let task_ids = task_ids.into_iter().collect::<HashSet<_>>();
+        Ok(self
+            .all_tasks()
```
diptanu, line 387: We already have an index of task_id -> Task in the state machine, use that to get the tasks vs getting all the tasks and then filtering them like this.

src/state/mod.rs
```diff
@@ -405,6 +442,28 @@ impl App {
...
+
+    pub async fn get_task_ids_for_executor(
+        &self,
+        executor_id: ExecutorIdRef<'_>,
+    ) -> Result<HashSet<TaskId>, anyhow::Error> {
+        // look up the executor
+        let executor = self.get_executor_by_id(executor_id).await?;
+        // look up tasks matching the mimetype of the executor using the
+        // unfinished_tasks_by_content_type index
+        self.get_unfinished_task_ids_by_content_type(
```
diptanu, line 461: This is not going to work because there might be unfinished tasks of the same mime type on other executors. Ether - 1. Filter the unfinished tasks by tasks with the executor id, or create a reverse index of unfinished tasks by executor id( executor_id -> HashSet<Task>).

diptanu (changes requested):

src/task_allocator/mod.rs
```diff
@@ -0,0 +1,433 @@
...
+        content_types: HashSet<String>,
+    ) -> Result<(), anyhow::Error>;
+}
+
+#[async_trait::async_trait]
+impl IsTaskAllocator for TaskAllocator {
+    async fn allocate_tasks(&self, task_ids: HashSet<TaskId>) -> Result<(), anyhow::Error> {
+        let plan = self.planner.plan_allocations(task_ids).await?;
+        self.committer
+            .commit_task_assignments(plan.into_task_executor_allocations())
```
diptanu, line 64: Everywhere you get a plan, you are changing it from ExecutorId -> HashSet<TaskID> to TaskID->Executor, change the planner to return the task to executor mapping instead to prevent this.

src/task_allocator/mod.rs
```diff
@@ -0,0 +1,433 @@
...
+            .await?;
+        Ok(())
+    }
+
+    /// Reschedule all tasks that match the executor's input mime types, even if
```
diptanu, line 69: We can't simply match extractor's mime type, it has to match the extractor name also. There could be several extractors which work off PDF documents for example, but if an invoice extractor dies, we need to reschedule the tasks on that executor to others which are running the same extractor.

src/task_allocator/mod.rs
```diff
@@ -0,0 +1,433 @@
...
+        self.committer
+            .commit_task_assignments(plan.into_task_executor_allocations())
+            .await?;
+        Ok(())
+    }
+
+    /// Reschedule all tasks that match the content types, even if
+    /// they are already assigned to a different executor.
+    /// This is useful for example when a new binding is added to an executor
+    async fn reallocate_all_tasks_matching_content_types(
```
diptanu, line 89: Matching extractors, content type would be matched automatically.

src/task_allocator/mod.rs
```diff
@@ -0,0 +1,433 @@
...
+            .await?;
+        Ok(())
+    }
+}
+
+struct RaftCommitter {
+    shared_state: SharedState,
+}
+
+/// Separated out so it can be mocked, and the entire state doesn't need to be
```
diptanu, line 109: Do we need this if we setup a single node in-memory shared state? It's super fast anyways, based on the test in coordinator.

src/task_allocator/planner/plan.rs
```diff
@@ -0,0 +1,29 @@
...
+impl Deref for TaskAllocationPlan {
+    type Target = HashMap<ExecutorId, HashSet<TaskId>>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl TaskAllocationPlan {
+    pub fn into_task_executor_allocations(&self) -> HashMap<TaskId, ExecutorId> {
```
diptanu, line 20: Let's make the planner return TaskId -> ExecutorID if that's what is most commonly used.

src/task_allocator/planner/round_robin.rs
```diff
@@ -0,0 +1,408 @@
...
+        if tasks.is_empty() {
+            return Ok(TaskAllocationPlan(HashMap::new()));
+        }
+        // if no executors, return
+        if extractor_executors_map.is_empty() {
+            return Ok(TaskAllocationPlan(HashMap::new()));
+        }
+
+        let tasks_len = tasks.len();
+        let mut plan = TaskAllocationPlan(HashMap::with_capacity(extractor_executors_map.len()));
```
diptanu, line 71: Hmm, returning TaskId -> ExecutorID would be better for this as well, we won't have to allocate hashes as the number of executors in the cluster.

src/task_allocator/planner/round_robin.rs
```diff
@@ -0,0 +1,408 @@
...
+
+        let avg_tasks_per_executor = tasks_len / extractor_executors_map.len() / 100;
+        let increment_size = if avg_tasks_per_executor > 0 {
+            avg_tasks_per_executor
+        } else {
+            1
+        };
+
+        // Initialize a min-heap for each extractor
+        for (extractor, executors) in extractor_executors_map.iter() {
```
diptanu, line 83: Instead of doing this everytime, just cache it. And adjust it everytime new allocations are being done.

diptanu (changes requested):

src/state/mod.rs
```diff
@@ -371,6 +374,22 @@ impl App {
         Ok(matched_content_list)
     }
 
+    pub async fn all_tasks(&self) -> Result<Vec<internal_api::Task>> {
```
diptanu, line 377: This is no longer needed.

src/state/mod.rs
```diff
@@ -371,6 +374,22 @@ impl App {
...
+        let store = self.indexify_state.read().await;
+        Ok(store.tasks.values().cloned().collect_vec())
+    }
+
+    pub async fn get_tasks_by_ids(
+        &self,
+        task_ids: HashSet<TaskId>,
+    ) -> Result<Vec<internal_api::Task>, anyhow::Error> {
+        let task_ids = task_ids.into_iter().collect::<HashSet<_>>();
+        future::join_all(task_ids.iter().map(|id| self.task_with_id(id)))
```
diptanu, line 387: There is no point of doing these in parallel as these are objects in memory, just serially going over them and doing hash lookups is enough. Unless these operations are blocking in some way or compute bound doing parallel copy from memory location A to B doesn't speed things up.

src/state/mod.rs
```diff
@@ -405,6 +436,43 @@ impl App {
...
+    }
+
+    pub async fn get_task_ids_for_executor(
+        &self,
+        executor_id: ExecutorIdRef<'_>,
+    ) -> Result<HashSet<TaskId>, anyhow::Error> {
+        // look up the executor
+        let executor = self.get_executor_by_id(executor_id).await?;
+        // look up unfinished tasks matching the executor's extractor
+        self.get_task_ids_for_extractor(&executor.extractor.name.as_str())
```
diptanu, line 458: Looks like you are getting all the task ids for a given extractor. The name doesn't feel right. Or are you getting all the task ids for a given executor id?

src/state/mod.rs
```diff
@@ -405,6 +436,43 @@ impl App {
...
+        &self,
+        extractor: &str,
+    ) -> Result<HashSet<TaskId>, anyhow::Error> {
+        let sm = self.indexify_state.read().await;
+        let task_ids = sm
+            .unfinished_tasks_by_extractor
+            .get(extractor)
+            .cloned()
+            .unwrap_or_default();
+        drop(sm);
```
diptanu, line 472: Why are you dropping the sm? Once the method exits, the read lock is going to dropped.

src/state/mod.rs
```diff
@@ -405,6 +436,43 @@ impl App {
...
+            .await
+    }
+
+    pub async fn get_task_ids_for_extractor(
+        &self,
+        extractor: &str,
+    ) -> Result<HashSet<TaskId>, anyhow::Error> {
+        let sm = self.indexify_state.read().await;
+        let task_ids = sm
+            .unfinished_tasks_by_extractor
```
diptanu, line 468: Rename this method to say unfished_tasks_by_executor

src/state/store/mod.rs
```diff
@@ -34,8 +41,11 @@ pub type TaskId = String;
 pub type StateChangeId = String;
 pub type ContentId = String;
 pub type ExecutorId = String;
+pub type ExecutorIdRef<'a> = &'a str;
 pub type ExtractionEventId = String;
 pub type ExtractorName = String;
+pub type ContentType = String;
+pub type ExtractorExecutorsMap = HashMap<String, HashSet<String>>;
```
diptanu, line 48: HashMap<ExecutorId, HashSet<ExecutorId>>

src/state/store/state_machine_objects.rs
```diff
@@ -143,11 +157,19 @@ impl IndexifyState {
...
+                        .remove(&task.id);
                     if let Some(executor_id) = executor_id {
+                        // remove the task from the executor's task assignments
                         self.task_assignments
                             .entry(executor_id.clone())
                             .or_default()
                             .remove(&task.id);
+
+                        // decrement the load
+                        *self.executor_load.entry(executor_id.clone()).or_default() -= 1;
```
diptanu, line 172: The type for this is usize, we need a check over here to make sure we are not decrementing beyond zero.

src/task_allocator/planner/load_aware_distributor.rs
```diff
@@ -0,0 +1,881 @@
...
+        }
+
+        let executor_load = self.shared_state.get_executor_load().await;
+        let tasks_by_extractor = self.group_tasks_by_extractor(&task_ids).await;
+
+        let mut plan = TaskAllocationPlan(HashMap::with_capacity(extractor_executors_map.len()));
+        let mut executors_load_heap: HashMap<String, BinaryHeap<Reverse<ExecutorLoad>>> =
+            HashMap::new();
+
+        for extractor in tasks_by_extractor.keys() {
```
diptanu, line 90: Are we not using the data from the state machine?

diptanu (changes requested):

src/state/mod.rs
jackbackes: I'm going to drop this method and the corresponding "reallocate_all_tasks_matching_executor" - the idea here was to find all the tasks that are compatible with the executor (i.e. that share the same extractor with the executor) but it's simpler to just have a "reallocate_all_tasks_matching_extractor" method and have the calling function get the extractor name.

src/state/mod.rs
jackbackes: I was running into an issue where tests were hanging because of a deadlock, so I got a little overzealous with explicitly dropping the lock. But yeah, it's not needed here.

src/state/mod.rs
jackbackes: this is a mapping of extractor name to task, so renaming to unfinished_tasks_by_extractor

src/state/store/state_machine_objects.rs
jackbackes: Added.

src/task_allocator/planner/load_aware_distributor.rs
jackbackes: updated to iterate over executor load, since we can assume it's guaranteed that any executor exists in the index.

src/state/mod.rs
jackbackes: Method isn't necessary anymore. Dropping.

src/task_allocator/planner/load_aware_distributor.rs
```diff
@@ -0,0 +1,891 @@
...
+use crate::state::{
+    store::{ExtractorName, TaskId},
+    SharedState,
+};
+
+/// Represents the load of an executor
+#[derive(Debug, Clone)]
+struct ExecutorLoad {
+    executor_id: String,
+    load: usize,
```
diptanu, line 18: `num_running_tasks`

src/task_allocator/planner/load_aware_distributor.rs
```diff
@@ -0,0 +1,891 @@
...
+}
+
+#[async_trait::async_trait]
+impl AllocationPlanner for LoadAwareDistributor {
+    async fn plan_allocations(&self, task_ids: HashSet<TaskId>) -> AllocationPlannerResult {
+        // if no tasks, return
+        if task_ids.is_empty() {
+            return Ok(TaskAllocationPlan(HashMap::new()));
+        }
+        let extractor_executors_map = self.shared_state.get_extractor_executors_map().await;
```
diptanu, line 78: Please remove this since we are not using this

diptanu (changes requested):

PR #292, opened 2024-02-01
Add index for unfinished tasks by content type

jackbackes:
Implemented an index to track unfinished tasks by their content type in the StateMachine struct. This index, `unfinished_tasks_by_content_type`, enhances the efficiency of task scheduling by allowing quick lookups.

Key Changes:

- Created a new struct, `TaskByContentTypeIndex`, to store the mapping between content types and task IDs.
- Integrated `TaskByContentTypeIndex` into `StateMachine`, ensuring that task additions and completions update this new index.
- Updated relevant SledStorable and SledStorableTestFactory implementations to accommodate the new index.
- Modified `RaftStorage` methods to maintain the index during task creation and completion.

This change lays the groundwork for optimizing the task scheduling process. Future refactoring will focus on adopting Sled's native data structures, moving away from the current approach of cloning the entire table.

Testing:
- Added tests for `TaskByContentTypeIndex` to ensure its correct functionality within the Sled storage framework.
- Ensured existing unit tests pass with the introduction of the new index.

### DIFF
--- a/src/state/store/mod.rs
+++ b/src/state/store/mod.rs
@@ -38,7 +38,7 @@ use serde::{Deserialize, Serialize};
 pub use sled_store::SledStore;
 use sled_store::*;
 
-use self::impl_sled_storable::TaskByContentTypeIndex;
+use self::impl_sled_storable::UnfinishedTasksByContentTypeIndex;
 
 use super::{NodeId, TypeConfig};
 
@@ -182,7 +182,7 @@ pub struct StateMachine {
 
     pub index_table: HashMap<String, internal_api::Index>,
 
-    pub unfinished_tasks_by_content_type: TaskByContentTypeIndex,
+    pub unfinished_tasks_by_content_type: UnfinishedTasksByContentTypeIndex,
 }
 
 #[async_trait]
--- a/src/state/store/mod.rs
+++ b/src/state/store/mod.rs
@@ -38,6 +38,8 @@ use serde::{Deserialize, Serialize};
 pub use sled_store::SledStore;
 use sled_store::*;
 
+use self::impl_sled_storable::TaskByContentTypeIndex;
+
 use super::{NodeId, TypeConfig};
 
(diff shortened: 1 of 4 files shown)


src/state/store/mod.rs
```diff
@@ -505,9 +508,24 @@ impl RaftStorage<TypeConfig> for Arc<SledStore> {
                     Request::CreateTasks { tasks } => {
                         for task in tasks {
                             sm.tasks.insert(task.id.clone(), task.clone());
+                            sm.unfinished_tasks_by_content_type
```
diptanu, line 511: Since you made this a type already why not create a method inside it which does the following, `.add(task_id)`

src/state/store/mod.rs
```diff
@@ -505,9 +508,24 @@ impl RaftStorage<TypeConfig> for Arc<SledStore> {
...
+                                .insert(task.id.clone());
                             sm.unassigned_tasks.insert(task.id.clone());
                         }
                         sm.overwrite_sled_kv(&state_machine_tree, "tasks", sm.tasks.clone())?;
+                        sm.overwrite_sled_kv(
+                            &state_machine_tree,
+                            "unfinished_tasks_by_content_type",
+                            sm.unfinished_tasks_by_content_type.clone(),
+                        )?;
+                        sm.overwrite_sled_kv(
```
diptanu, line 524: This was not being written already?

src/state/store/mod.rs
```diff
@@ -686,6 +704,11 @@ impl RaftStorage<TypeConfig> for Arc<SledStore> {
                         sm.tasks.insert(task.id.clone(), task.clone());
                         if *mark_finished {
                             sm.unassigned_tasks.remove(&task.id);
+                            sm.unfinished_tasks_by_content_type
+                                .0
+                                .entry(task.content_metadata.content_type.clone())
```
diptanu, line 709: .`map` is probably better since we want the side effect of removal only if the entry is present.

diptanu (changes requested):

PR #297, opened 2024-02-03
Changed storage for raft 

diptanu:
* Using RocksDB because it's more mature than sled
* Storing and restoring state machine  from snapshots


drmingdrmer:
For production use, the snapshot should be saved in a on-disk file, instead of in a rocksdb record.

PR #309, opened 2024-02-06
Fix sample_config.yaml test

lucasjacks0n:
this pr fixes tests


diptanu:
Don't merge this, we will still support qdrant as one of the supported vector datastore. The getting started experience will use pg_vector for the time being, but qdrant is still supported if a user choses to use it.

diptanu:
Looks great

PR #323, opened 2024-02-07
WIP: Snapshot version migration for backward compatibility

jackbackes:


diptanu:
Don't need this for the time being

PR #329, opened 2024-02-08
Using task allocator

diptanu:


src/coordinator.rs
```diff
@@ -7,28 +7,35 @@ use std::{
...
 use jsonschema::JSONSchema;
 use tokio::sync::watch::Receiver;
 use tracing::info;
 
-use crate::{coordinator_filters::*, state::SharedState};
+use crate::{coordinator_filters::*, state::SharedState, task_allocator::TaskAllocator};
 
 pub struct Coordinator {
     shared_state: SharedState,
+    task_allocator: TaskAllocator,
```
jackbackes, line 19: the task allocator is stateless so it could be instantiated as needed. A `task_allocator()` method might be better rather than having a persistent allocator.

src/coordinator.rs
```diff
@@ -319,12 +322,18 @@ impl Coordinator {
 
     #[tracing::instrument(skip(self))]
     pub async fn process_and_distribute_work(&self) -> Result<(), anyhow::Error> {
-        self.process_extraction_events().await?;
-
-        let task_assignments = self.distribute_work().await?;
+        let state_changes = self.shared_state.unprocessed_state_change_events().await?;
+        println!("state_changes: {:?}", state_changes);
```
jackbackes, line 326: Remove printlns

src/coordinator.rs
```diff
@@ -7,28 +7,35 @@ use std::{
...
+        Arc::new(Self {
+            shared_state,
+            task_allocator,
+        })
     }
 
     #[tracing::instrument(skip(self))]
-    pub async fn process_extraction_events(&self) -> Result<(), anyhow::Error> {
-        let state_changes = self.shared_state.unprocessed_state_change_events().await?;
+    pub async fn process_extraction_events(
```
jackbackes, line 32: maybe `create_tasks_from_extraction_events` ?

src/coordinator.rs
```diff
@@ -319,12 +322,18 @@ impl Coordinator {
...
     #[tracing::instrument(skip(self))]
     pub async fn process_and_distribute_work(&self) -> Result<(), anyhow::Error> {
-        self.process_extraction_events().await?;
-
-        let task_assignments = self.distribute_work().await?;
+        let state_changes = self.shared_state.unprocessed_state_change_events().await?;
+        println!("state_changes: {:?}", state_changes);
+        let tasks = self.process_extraction_events(&state_changes).await?;
+        println!("tasks: {:?}", tasks);
+        self.shared_state.create_tasks(tasks.clone()).await?;
```
jackbackes, line 329: I think this will allocate tasks twice. shared_state.create_tasks calls raft.client_write(). This is already done within the task allocator in `commit_task_assignments()` call.

jackbackes: actually I think it's fine. The tasks get created here but then get assigned by the allocator.

PR #332, opened 2024-02-09
Added Indexify Extractor Packager

jackbackes:
This change adds a Python-native extractor packager.

WIP - Tests pass and output Dockerfile is built successfully. Haven't run the Dockerfile yet though, and there is some additional work to do.

To Do:

- Add CLI
- Remove Rust code


diptanu:
This is probably not needed anymore

PR #338, opened 2024-02-10
Removing executor from indexify binary

diptanu:
Removing executors and packaging related code from Indexify binary, as they have moved to indexify_extractor_sdk


jackbackes:
LFG! Shipit!

PR #343, opened 2024-02-13
Added raft forwarding to leader - Attempt 2

rakshith-ravi:
Had to dig through the source code for why `is_leader()` was depreciated. Let me know if there's a better solution


diptanu:
This won't work as discussed offline.

PR #350, opened 2024-02-15
Added extractors for init compose

rakshith-ravi:
Also added network aliases instead of hard-coded IP address to compose files, since they're much easier to work with.

You can pass as many `--extractor` arguments as you want and they'd all become a unique service in the resulting compose file

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,9 +1,6 @@
 version: "3"
 networks:
   server:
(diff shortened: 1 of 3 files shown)


rakshith-ravi:
Oh also, I should mention. The template is embedded. Tested this with a release build and it's able to generate the compose file as per the required spec

PR #356, opened 2024-02-19
Add ui build step to build.rs

aahmed-se:
* Add command to build ui artifacts with npm in build.rs

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,15 +1,22 @@
 use std::error::Error;
+use std::path::Path;
 use std::process::Command;
 use vergen::EmitBuilder;
 
 fn main() -> Result<(), Box<dyn Error>> {
-    // Execute the npm commands in the 'ui' directory
-    if !Command::new("sh")
-        .arg("-c")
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -1,8 +1,25 @@
 use std::error::Error;
-
+use std::path::Path;
+use std::process::Command;
 use vergen::EmitBuilder;
 
 fn main() -> Result<(), Box<dyn Error>> {
+    // Check if the npm build is needed by verifying the existence of the 'ui/build' directory
+    if !Path::new("ui/build").exists() {
```
diptanu, line 8: I think it needs to be a little bit more sophisticated, we should build the ui if anything in that folder changed. Otherwise this will only build once and never again even if the UI is changed unless the build folder is nuked.

diptanu (changes requested):

build.rs
aahmed-se: Updated it with timestamp check

PR #389, opened 2024-03-01
Cluster Membership Changes Via Seed Node

redixhumayun:
This PR aims to allow any nodes that start up in the cluster to get the current state of cluster membership by contacting the seed node.

## Approach

* A seed node has been added to the .yaml file.
* A new RPC called `GetClusterMembership` has been defined in the `proto` file and an implementation provided. This implementation uses the methods `add_learner` and `change_membership` provided by `OpenRaft` to update cluster configuration. It follows [this example](https://github.com/datafuselabs/openraft/blob/d041202a9f30b704116c324a6adc4f2ec28029fa/examples/raft-kv-memstore/tests/cluster/test_cluster.rs#L75-L103).
* The `initialize` method will be called only on the seed node. Every other node will contact the seed node to get added as a learner and get the latest view of the cluster.
* A node will start a recurring `tokio` task to contact the seed node regularly and check cluster membership. Upon receiving a request from a node not currently in the cluster, the seed node will update the cluster configuration for all the nodes.

## Questions
* How do we handle seed node failure? Can we do something simple like having a list of alternate seed nodes?


src/state/mod.rs
```diff
@@ -793,6 +809,45 @@ impl App {
         let _resp = self.raft.client_write(req).await?;
         Ok(())
     }
+
+    pub fn start_periodic_membership_check(self: &Arc<Self>) {
+        let app_clone = Arc::clone(&self);
+        tokio::spawn(async move {
+            let mut interval = tokio::time::interval(tokio::time::Duration::from_secs(3));
```
diptanu, line 816: Move this to a constant

src/state/mod.rs
```diff
@@ -793,6 +809,45 @@ impl App {
         let _resp = self.raft.client_write(req).await?;
         Ok(())
     }
+
+    pub fn start_periodic_membership_check(self: &Arc<Self>) {
```
diptanu, line 813: Pass in a shutdown. Pass shutdown_rx into this, and do select! on either the ticker or the shutdown_rx to unblock sleeping the loop, and if shutdown is unblocked, write logs that the membership check is stopping, stop the ticket and return our the task.

src/state/mod.rs
```diff
@@ -793,6 +809,45 @@ impl App {
...
+                if app_clone.is_seed_node() {
+                    continue;
+                }
+                app_clone.check_cluster_membership().await;
+            }
+        });
+    }
+
+    async fn check_cluster_membership(&self) {
+        let addr = format!("localhost:{}", self.raft_port);
```
diptanu, line 828: This feels wrong, use the seed_addr from the server_config for this.

src/state/mod.rs
```diff
@@ -793,6 +809,45 @@ impl App {
...
+        match self
+            .network
+            .get_cluster_membership(self.id, &addr, &self.seed_node)
+            .await
+        {
+            Ok(resp) => {
+                info!("Cluster membership response: {:?}", resp);
```
diptanu, line 835: Remove the logs

src/state/mod.rs
```diff
@@ -793,6 +809,45 @@ impl App {
...
+                info!("Checking own cluster");
+                let metrics = self.raft.metrics().borrow().clone();
```
diptanu, line 837: No need to clone the whole metrics, just get reference and only copy out the node names.

src/state/mod.rs
```diff
@@ -793,6 +809,45 @@ impl App {
...
+            }
+        });
+    }
+
+    async fn check_cluster_membership(&self) {
+        let addr = format!("localhost:{}", self.raft_port);
+        match self
+            .network
+            .get_cluster_membership(self.id, &addr, &self.seed_node)
+            .await
```
diptanu, line 832: .ok_or(|e| anyhow!(...))

src/state/mod.rs
```diff
@@ -201,13 +203,27 @@ impl App {
                 .map_err(|e| anyhow!("grpc server error: {}", e))
         });
         app.join_handles.lock().await.push(h);
+        app.start_periodic_membership_check();
 
         Ok(app)
     }
 
+    /// This function checks whether this node is the seed node
+    fn is_seed_node(&self) -> bool {
```
diptanu, line 212: Let's change this logic a little and make things a little flexible - let any node be the seed node for another node. All we need from the seed node is the list of other peers it knows in the cluster. And the seed node, needs to be able to forward the request to the leader.

src/state/mod.rs
```diff
@@ -133,35 +132,35 @@ impl App {
...
-                peer.node_id,
-                BasicNode {
-                    addr: peer.addr.clone(),
-                },
-            );
-        }
+        nodes.insert(
+            server_config.node_id,
+            BasicNode {
+                addr: format!("localhost:{}", server_config.raft_port),
```
diptanu, line 152: What are we doing here?

src/state/mod.rs
```diff
@@ -105,6 +102,8 @@ pub struct App {
     pub indexify_state: Arc<RwLock<IndexifyState>>,
     pub config: Arc<openraft::Config>,
     state_change_rx: Receiver<StateChange>,
+    network: Network,
```
diptanu, line 105: Do we need the whole network implementation here or do you need the gRPC client? I would suggest not having the network implementation here if you need only the client implementation.

src/state/grpc_server.rs
```diff
@@ -71,4 +75,58 @@ impl RaftApi for RaftGrpcServer {
         }
         .await
     }
+
+    async fn get_cluster_membership(
+        &self,
+        request: Request<GetClusterMembershipRequest>,
+    ) -> Result<Response<RaftReply>, Status> {
+        let req = request.into_inner();
+        info!(
```
diptanu, line 84: Log only if there is a new node discovered.

src/state/grpc_server.rs
```diff
@@ -71,4 +75,58 @@ impl RaftApi for RaftGrpcServer {
...
+            "Received get_cluster_membership request from Node ID: {}, Address: {}",
+            req.node_id, req.address
+        );
+
+        let metrics = self.raft.metrics().borrow().clone();
```
diptanu, line 89: No need to clone the metrics object, just the nodes are enough.

src/state/grpc_server.rs
```diff
@@ -71,4 +75,58 @@ impl RaftApi for RaftGrpcServer {
...
+        let nodes_in_cluster = metrics
+            .membership_config
+            .nodes()
+            .map(|(node_id, node)| (*node_id, node.clone()))
+            .collect::<BTreeMap<_, _>>();
+        let mut node_ids: Vec<u64> = nodes_in_cluster.keys().cloned().collect();
+        if nodes_in_cluster.contains_key(&req.node_id) {
+            info!("This node is already present in the cluster, not modifying the cluster");
+            let response = RaftReply {
```
diptanu, line 98: Let the response always be the known list of nodes in the cluster.

src/state/grpc_server.rs
```diff
@@ -71,4 +75,58 @@ impl RaftApi for RaftGrpcServer {
...
+                data: "".to_string(),
+                error: "".to_string(),
+            };
+
+            return Ok(Response::new(response));
+        }
+
+        let node_to_add = BasicNode { addr: req.address };
+        match self
```
diptanu, line 107: .ok_or(|e| ...)

diptanu (changes requested):
Looking good overall but we need to change a few things.

src/state/mod.rs
redixhumayun: @diptanu I'm not using the seed_node address here. I'm using the from node's address which isn't provided in the config file.

src/state/network.rs
```diff
@@ -27,18 +25,53 @@ pub struct Network {
...
+
+    pub async fn get_cluster_membership(
+        &self,
+        node_id: NodeId,
+        node_addr: &str,
+        target_addr: &str,
+    ) -> Result<RaftReply, anyhow::Error> {
+        let client_result = self.raft_client.clone().get(target_addr).await;
+
+        let mut client = match client_result {
```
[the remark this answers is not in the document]
redixhumayun: Assuming you mean `map_err` here

src/state/mod.rs
redixhumayun: assuming you mean `map_err` here since its a `Result` not an `Option`

src/state/grpc_server.rs
redixhumayun: Do we want to do anything with this response for now? Because OpenRaft is already handling cluster membership changes for us

diptanu: A shell response object would be useful in the future if we want to add some information from the seed node about the state of the cluster/leader at the time the new node is trying to join. The response metadata could be logged for example. So lets return an empty object `ClusterMembershipResponse`

src/state/network.rs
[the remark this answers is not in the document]
diptanu: Yeah

src/state/mod.rs
diptanu: Oh I see, you are passing along the address of the current machine, in that case it would be the advertise addr, which is the network routable address of the current node. We try to probe the public address of the node first, or let operators configure it via the config. Use the `listen_if` attribute of `ServerConfig` instead of localhost here.

src/state/mod.rs
redixhumayun: @diptanu This is creating the list of members that a cluster will be initialized with. The `initialize_raft` function has the following lines to show how its being used

```
if !self.is_seed_node() {
            return Ok(());
        }
        match self.raft.initialize(self.nodes.clone()).await {
//  removed for brevity
```

src/state/mod.rs
redixhumayun: @diptanu I added the `get_cluster_membership` function into the `Network` impl because it seemed like every other network call followed the same process

`app calls network -> network calls client to get handle -> network makes request -> network returns response` 

It seems odd to bypass the network implementation when making a network request. 

Happy to change it though. Instead of storing the network here I would just store the `RaftClient` instead. Let me know.

src/state/mod.rs
redixhumayun: @diptanu Can I take care of this in the next PR since that's we plan to tackle leader forwarding?

src/state/mod.rs
diptanu: It's fine, yeah.

protos/raft.proto
```diff
@@ -13,10 +13,21 @@ message RaftReply {
     string error = 2;
 }
 
+message GetClusterMembershipRequest {
+  uint64 node_id = 1;
+  string address = 2;
+}
+
+message GetClusterMembershipResponse {
+    string data = 1;
```
diptanu, line 22: No need for data and error here, leave them empty as we are not using them right now.

src/coordinator_service.rs
```diff
@@ -493,6 +468,7 @@ pub struct CoordinatorServer {
 impl CoordinatorServer {
     pub async fn new(config: Arc<ServerConfig>) -> Result<Self, anyhow::Error> {
         let addr: SocketAddr = config.coordinator_lis_addr_sock()?;
+        info!("Addr is {}", addr);
```
diptanu, line 471: Remove this

src/state/grpc_server.rs
```diff
@@ -71,4 +76,66 @@ impl RaftApi for RaftGrpcServer {
...
+            req.node_id, req.address
+        );
+        let node_to_add = BasicNode { addr: req.address };
+
+        async {
+            let add_learner_result = self
+                .raft
+                .add_learner(req.node_id, node_to_add.clone(), false)
+                .await
+                .map_err(|e| Status::internal(format!("Error adding learner: {}", e)));
```
diptanu, line 115: .map_err(..)?;

src/state/grpc_server.rs
```diff
@@ -71,4 +76,66 @@ impl RaftApi for RaftGrpcServer {
...
+            if let Err(e) = add_learner_result {
+                return Err(e);
+            }
+
+            node_ids.push(req.node_id);
+            let change_membership_result = self
+                .raft
+                .change_membership(node_ids, false)
+                .await
+                .map_err(|e| Status::internal(format!("Error changing membership: {}", e)));
```
diptanu, line 126: Same as above

diptanu: We also have grpc_helper::internal_err to create errors

src/state/mod.rs
diptanu: Cool. I hope you have seen my comment about advertise addr (using listen_if from server config)

src/state/mod.rs
diptanu: Cool

diptanu (changes requested):

bshashikant:
The approach looks good . 
A couple of approaches to handle the seed node unavailability due to seed node failure/network partition etc :
1. Retry with certain sleep for a certain timeout optimistically waiting for the dead node to become available (network partitions may resolve on their own if any)
2. Built Service discovery mechanisms using some centralized persistent  state store like Zookeeper/Etcd/objectStore
3. Deploy Seed Node in HA mode

src/state/mod.rs
redixhumayun: @diptanu Took care of this.

local_server_config_2.yaml
```diff
@@ -0,0 +1,87 @@
+# Address on which the server listens
+listen_if: 0.0.0.0
```
diptanu, line 2: Remove this file and the 3rd file. Lets add a docker compose file in a subsequent PR which spins up a 3 node control plane cluster for testing.

src/state/mod.rs
```diff
@@ -133,35 +135,35 @@ impl App {
...
         let (tx, rx) = watch::channel::<()>(());
 
         let addr = server_config
             .raft_addr_sock()
             .map_err(|e| anyhow!("unable to create raft address : {}", e.to_string()))?;
 
         info!("starting raft server at {}", addr.to_string());
-        let raft_srvr = RaftApiServer::new(RaftGrpcServer::new(Arc::new(raft.clone())));
+        let rpc_server = RaftGrpcServer::new(Arc::new(raft.clone()));
+        let raft_srvr = RaftApiServer::new(rpc_server);
```
diptanu, line 166: Inline the two lines.

src/state/mod.rs
```diff
@@ -793,6 +818,36 @@ impl App {
...
+                            error!("failed to check cluster membership: {}", e);
+                        }
+                    }
+                }
+            }
+        });
+    }
+
+    async fn check_cluster_membership(&self) -> Result<ClusterMembershipResponse, anyhow::Error> {
+        let addr = format!("{}:{}", self.listen_if, self.raft_port);
```
diptanu, line 846: Just build cache the addr while initializing the app. no need to copy listen_if and port

diptanu (changes requested):
Looks good, some more changes and then we can check it in.

PR #391, opened 2024-03-05
Request Forwarding To Leader

redixhumayun:
## What This PR Does

* Renames `get_cluster_membership` to `join_cluster_membership` to make it more semantically meaningful
* Uses `RaftRequest` and `RaftReply` for all `gRPC` requests
* Adds leader redirection and associated tests

## Leader Forwarding

The approach is simple. All forwardable requests are defined in the `src/state/store/requests.rs` file. Some of them are executed directly on the node and currently one (`JoinClusterMembership`) on the gRPC server. 

On the gRPC server, if a function needs to execute on the leader the node will first check if it is the leader via `ensure_leader`. If it isn't, it forwards the request to the node it believes is the leader. If the node that is supposed to be the leader receives the request and is not the leader, an error is thrown. Is this correct?

If the node that receives the request is the leader, it fulfils the request and returns the response.

## Testing

I've added a simple test to ensure that leader forwarding occurs but to truly check that the message goes to the leader, I imagine we'd need some form of a simulated network which will allow us to determine at any point what messages have been delivered to which node. Is there a simpler way?

The test I currently have cannot do that since it can't inspect the network.


protos/raft.proto
```diff
@@ -18,6 +18,18 @@ message GetClusterMembershipRequest {
   string address = 2;
 }
 
+message ForwardableRequest {
+    oneof request {
+        GetClusterMembershipRequest getClusterMembershipRequest = 1;
```
diptanu, line 23: This approach is fine, add all the other types that can be forwarded.

diptanu (commented):
Request forwarding should happen at the server transparently.

diptanu (changes requested):

PR #397, opened 2024-03-07
Forwardable Raft

redixhumayun:
## What This PR Does

* Creates a `ForwardableRaft` struct which wraps the `Raft` instance of `OpenRaft`. `ForwardableRaft` allows access to the underlying `Raft` instance so it doesn't have to wrap every single method.
* Adds a `client_write` method to `ForwardableRaft` which wraps the `client_write` method of `openraft::Raft` so as to include functionality for leader forwarding.
* Adds a test harness which includes the ability to spin up a raft cluster and helper methods around it to make testing easier.
* Renames `Request` and `Response` from `state/store/requests.rs` to `StateMachineUpdateRequest` and `StateMachineUpdateResponse`. Doing this to clearly differentiate between `tonic::Request` and `tonic::Response` and these types which was an issue thus far.
* Adds test to assert that `ForwardableRaft` actually works


src/coordinator_service.rs
```diff
@@ -556,8 +557,6 @@ async fn run_scheduler(
     mut state_watcher_rx: Receiver<StateChange>,
     coordinator: Arc<Coordinator>,
 ) -> Result<()> {
-    //let mut interval =
```
diptanu, line 559: Remove

src/state/forwardable_raft.rs
```diff
@@ -0,0 +1,74 @@
+use std::{collections::BTreeMap, fmt::Debug};
+
+use anyhow::{anyhow, Result as AnyhowResult};
```
diptanu, line 3: Why import this as AnyhowResult?

src/state/grpc_server.rs
```diff
@@ -83,25 +83,35 @@ impl RaftGrpcServer {
             },
         }
     }
+
+    async fn handle_client_write(
+        &self,
+        request: state::store::requests::Request,
+    ) -> Result<Response<RaftReply>, Status> {
+        match self.raft.client_write(request).await {
```
diptanu, line 91: let _ =  self.raft.client_write(request).map_err(|e| GrpcHelper::internal_err(e.to_string()));
GrpcHelper::ok_response("")

src/state/forwardable_raft.rs
```diff
@@ -0,0 +1,74 @@
...
+    pub async fn initialize(
+        &self,
+        members: BTreeMap<NodeId, BasicNode>,
+    ) -> Result<(), RaftError<InitializeError>> {
+        self.raft.initialize(members).await
+    }
+
+    /// Use this to detect whether the current node is the leader
+    async fn ensure_leader(&self) -> AnyhowResult<Option<ForwardToLeader>> {
+        match self.raft.ensure_linearizable().await {
```
diptanu, line 66: Remove the match and make the lines read linearly.

diptanu (changes requested):

src/state/forwardable_raft.rs
redixhumayun: @diptanu Because the `AnyhowResult` is just a placeholder until I figure out exactly what needs to be returned from here and there might be places I choose to use `std::Result` in the same file. 

It makes it more readable exactly which `Result` I am using - whether the `std` or `anyhow` version.

redixhumayun:
@diptanu Overall approach works?

src/state/forwardable_raft.rs
redixhumayun: @diptanu nvm, i've just used `anyhow::Result` everywhere in this file now

src/state/network.rs
```diff
@@ -14,8 +15,9 @@ use openraft::{
     },
     BasicNode,
 };
-use tonic::IntoRequest;
+use tonic::{IntoRequest, Response};
 
+use super::store::requests::Response as RequestResponse;
```
diptanu, line 20: Rename the Response to StateMachineUpdateResponse perhaps?

src/state/network.rs
```diff
@@ -51,12 +53,29 @@ impl Network {
         Self { raft_client }
     }
 
+    pub async fn forward(&self, target_addr: &str, request: Request) -> Result<(), anyhow::Error> {
```
diptanu, line 56: Shouldn't this also return the response? We would like every raft response have the node id which handled it.

src/state/mod.rs
```diff
@@ -894,3 +896,19 @@ async fn watch_for_leader_change(
...
+    use std::time::Duration;
+
+    use crate::test_utils::RaftTestCluster;
+
+    #[tokio::test]
+    #[tracing_test::traced_test]
+    async fn test_basic_read_own_write() -> Result<(), anyhow::Error> {
+        let cluster = RaftTestCluster::new(3).await?;
+        cluster.initialize(Duration::from_secs(2)).await?;
+        cluster.read_own_write(true).await?;
```
diptanu, line 911: What is this testing?

diptanu (changes requested):

src/state/mod.rs
redixhumayun: Currently it's an implementation which checks that any write made to the cluster can be read. The `read_own_write` method also accepts a boolean which can be used to change whether write goes to leader or follower to check that raft forwarding works.

src/state/network.rs
redixhumayun: Taken care of

src/state/network.rs
redixhumayun: Renamed `Request` and `Response` to `StateMachineUpdateRequest` and `StateMachineUpdateResponse`. This helps separate naming scheme from `tonic::Request` and `tonic::Response` so its slightly easier to read. 

One problem is that `JoinCluster` falls under `StateMachineUpdateRequest` which it strictly isn't. Have added a note next to it indicating the same.

PR #416, opened 2024-03-17
Adds Serialization Trait

redixhumayun:
## What This PR Does

* Adds a trait to hide implementation details of serialization and deseriliazation
* Uses the trait instead of `serde_json`

### DIFF
--- a/src/state/store/serializer.rs
+++ b/src/state/store/serializer.rs
@@ -1,0 +1,20 @@
+use serde::de::DeserializeOwned;
+
+use super::StateMachineError;
+
+pub struct Serializer;
+
+pub trait Serialize {
+    fn serialize<T: serde::Serialize>(value: &T) -> Result<Vec<u8>, StateMachineError>;
+    fn deserialize<T: DeserializeOwned>(bytes: &[u8]) -> Result<T, StateMachineError>;
+}
+
+impl Serialize for Serializer {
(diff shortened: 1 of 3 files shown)


src/state/store/serializer.rs
```diff
@@ -0,0 +1,20 @@
+use serde::de::DeserializeOwned;
+
+use super::StateMachineError;
+
+pub struct Serializer;
+
+pub trait Serialize {
```
diptanu, line 7: Call it something else like JsonEncoder so you won't clash with serge::Serialize

src/state/store/serializer.rs
```diff
@@ -0,0 +1,20 @@
...
+    fn serialize<T: serde::Serialize>(value: &T) -> Result<Vec<u8>, StateMachineError>;
```
diptanu, line 8: Make this a derive macro so that they get attached to the objects which implement serde::Serialize and serde::Deserialize

diptanu (changes requested):

src/state/store/serializer.rs
redixhumayun: @diptanu if we call it JsonEncoder, aren't we tying ourselves to always encoding to JSON bytes semantically? Why not just give it a more generic name? 

That way instead of swapping out JsonEncoder for a different implementation throughout the codebase, we can change the implementation internally in the trait and it will semantically still make sense

PR #417, opened 2024-03-18
Raft Metrics

redixhumayun:
## This PR

* Creates a global struct to maintain Raft metrics and ensures the struct is initialized lazily with a Mutex wrapper
* The `RaftMetrics` struct uses `String` instead of `&str` to avoid any issues with lifetimes that might crop up. This will probably require regular export of the metrics to avoid memory issues. 
* Creates helper methods to modify the metrics
* Uses helper methods when internal requests are being made to track various properties of the cluster
* Adds utilities like `TimedFuture` and `CounterGuard` to help keep track of metrics
* Adds gRPC endpoint to get Raft metrics

### DIFF
--- a/src/metrics/mod.rs
+++ b/src/metrics/mod.rs
@@ -101,6 +101,25 @@ pub mod raft_metrics {
         };
 
         use once_cell::sync::Lazy;
+        use serde::Serialize;
+
+        #[derive(Serialize)]
+        pub struct MetricsSnapshot {
+            pub fail_connect_to_peer: HashMap<String, u64>,
+            pub sent_bytes: HashMap<String, u64>,
+            pub recv_bytes: HashMap<String, u64>,
+            pub sent_failures: HashMap<String, u64>,
+            pub snapshot_send_success: HashMap<String, u64>,
+            pub snapshot_send_failure: HashMap<String, u64>,
+            pub snapshot_recv_success: HashMap<String, u64>,
+            pub snapshot_recv_failure: HashMap<String, u64>,
+            pub snapshot_send_inflights: HashMap<String, u64>,
+            pub snapshot_recv_inflights: HashMap<String, u64>,
+            pub snapshot_sent_seconds: HashMap<String, Vec<Duration>>,
+            pub snapshot_recv_seconds: HashMap<String, Vec<Duration>>,
+            pub snapshot_size: Vec<u64>,
+            pub last_snapshot_creation_time: Duration,
+        }
 
         struct RaftMetrics {
             fail_connect_to_peer: HashMap<String, u64>,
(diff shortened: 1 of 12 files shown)


diptanu:
@redixhumayun Add a server endpoint to return the metrics in JSON for the UI

src/metrics/mod.rs
```diff
@@ -0,0 +1,224 @@
...
+    }
+}
+
+pub mod raft_metrics {
+    pub mod network {
+        use std::{collections::HashMap, sync::Mutex, time::Duration};
+
+        use once_cell::sync::Lazy;
+
+        struct RaftMetrics {
```
diptanu, line 101: snapshot_size,
last_snapshot_creation_time

src/metrics/mod.rs
```diff
@@ -0,0 +1,224 @@
...
+    node_addr: String,
+    func: F,
+}
+
+impl<F> CounterGuard<F>
+where
+    F: Fn(String, i64),
+{
+    pub fn new(node_addr: String, func: F) -> Self {
+        func(node_addr.clone(), 1);
```
diptanu, line 81: Make this take a reference

src/metrics/mod.rs
```diff
@@ -0,0 +1,224 @@
...
+    C: FnOnce(Duration),
+{
+    TimedFuture::new(future, callback)
+}
+
+pub struct CounterGuard<F>
+where
+    F: Fn(String, i64),
+{
+    node_addr: String,
```
diptanu, line 72: &str

src/metrics/mod.rs
```diff
@@ -0,0 +1,224 @@
...
+                    snapshot_send_inflights: HashMap::new(),
+                    snapshot_recv_inflights: HashMap::new(),
+                    snapshot_sent_seconds: HashMap::new(),
+                    snapshot_recv_seconds: HashMap::new(),
+                }
+            }
+        }
+
+        static RAFT_METRICS: Lazy<Mutex<RaftMetrics>> =
+            Lazy::new(|| Mutex::new(RaftMetrics::new()));
```
diptanu, line 136: RWMutex

diptanu: Add a TODO here to make the locks more atomic

src/state/grpc_server.rs
```diff
@@ -134,29 +151,41 @@ impl RaftApi for RaftGrpcServer {
...
+        let remote_addr = self.get_request_addr(&request);
         let is_req = GrpcHelper::parse_req(request)?;
-        let resp = self
-            .raft
-            .install_snapshot(is_req)
-            .await
-            .map_err(|e| GrpcHelper::internal_err(e.to_string()));
+        let resp = self.raft.install_snapshot(is_req).await.map_err(|e| {
+            raft_metrics::network::incr_snapshot_recv_failure(remote_addr.clone());
+            GrpcHelper::internal_err(e.to_string())
```
diptanu, line 181: Make this take a reference

src/state/grpc_server.rs
```diff
@@ -134,29 +151,41 @@ impl RaftApi for RaftGrpcServer {
...
+
+        let remote_addr = self.get_request_addr(&request);
         let is_req = GrpcHelper::parse_req(request)?;
-        let resp = self
-            .raft
-            .install_snapshot(is_req)
-            .await
-            .map_err(|e| GrpcHelper::internal_err(e.to_string()));
+        let resp = self.raft.install_snapshot(is_req).await.map_err(|e| {
+            raft_metrics::network::incr_snapshot_recv_failure(remote_addr.clone());
```
diptanu, line 180: reference

src/state/network.rs
```diff
@@ -67,6 +72,9 @@ impl Network {
 
         let tonic_request = GrpcHelper::encode_raft_request(&request)?.into_request();
 
+        let bytes_sent = tonic_request.get_ref().data.len() as u64;
```
diptanu, line 75: What does this return - `usize` ?

diptanu (changes requested):

src/state/network.rs
redixhumayun: Yeah, `len()` returns `usize` by default

src/metrics/mod.rs
redixhumayun: Note - This is for the node itself not the nodes its sending a message to. So for node 1 it will be snapshot_size and last_snapshot_creation_time of node 1

src/metrics/mod.rs
redixhumayun: @diptanu i've added the TODO and replaced the Mutex with a RWLock.

For now, even if a thread panics and I have a chance of inconsistent data I am proceeding with updating the metrics anyway because I'm assuming that it's okay to have slightly inconsistent values for the metrics. 

If this is not okay, let me know and I will return a `Result<(), Box<dyn std::error::Error>>` from here and handle the error in the consumer.

PR #429, opened 2024-03-20
functions for doing stream writes to disk and s3

maxkozlovsky:
add traits and implementations for accepting stream of Bytes for writes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -90,7 +90,6 @@ sqlx = { version = "0.7.3", features = [
     "json",
 ] }
 strum = { version = "0.26.1", features = ["derive"] }
-tempfile = "3.2.0"
 thiserror = "1"
 tonic = { version = "0.11.0", features = ["prost"] }
 tokio = { version = "1", features = ["full"] }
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -90,6 +90,7 @@ sqlx = { version = "0.7.3", features = [
     "json",
 ] }
 strum = { version = "0.26.1", features = ["derive"] }
+tempfile = "3.2.0"
```
diptanu, line 93: Add this to [dev-dependencies] since you are using this only in tests.

src/blob_storage/disk.rs
```diff
@@ -28,6 +28,24 @@ impl BlobStorageWriter for DiskStorage {
...
+        Ok(path)
+    }
+
+    async fn put_stream(
+        &self,
+        key: &str,
+        data: BoxStream<'static, Result<Bytes>>,
+    ) -> Result<String, anyhow::Error> {
+        let path = format!("{}/{}", self.config.path, key);
+        let mut file = File::create(&path).await?;
```
diptanu, line 42: Should we use temp files here and then move the file if it's successful?

maxkozlovsky: In principle yes, I was following the original implementation. It also has the problem if the write fails/partially succeeds. I can add a todo for now.

Cargo.toml
maxkozlovsky: ok

PR #441, opened 2024-03-22
Content id policy mapping

redixhumayun:


src/state/store/state_machine_reader.rs
```diff
@@ -15,6 +15,27 @@ use super::{
 pub struct StateMachineReader {}
```
diptanu, line 15: In another pr, initialize this with a db so that we can store this in sate/mod.rs where all the methods to access raft lives.

src/state/store/state_machine_reader.rs
```diff
@@ -15,6 +15,27 @@ use super::{
...
 
 impl StateMachineReader {
+    pub async fn get_content_extraction_policy_mappings_for_content_id(
+        &self,
+        content_id: &str,
+        db: &Arc<OptimisticTransactionDB>,
+    ) -> Result<indexify_internal_api::ContentExtractionPolicyMapping, StateMachineError> {
```
diptanu, line 22: Return Some< indexify_internal_api::ContentExtractionPolicyMapping>

src/state/store/state_machine_reader.rs
```diff
@@ -15,6 +15,27 @@ use super::{
...
+        let txn = db.transaction();
```
diptanu, line 23: Why is there a transaction for doing reads from a single cf ?

src/state/store/state_machine_objects.rs
```diff
@@ -413,6 +415,121 @@ impl IndexifyState {
...
+            return Err(StateMachineError::DatabaseError(format!(
+                "Extraction policy {} not applied on content {} because extraction policy was not registered against the content",
+                extraction_policy_name, content_id
+            )));
+        }
+
+        //  Mark the time the content was processed against the extraction policy and
+        // store it back
+        let mut time_of_policy_completion = content_policy_mappings.time_of_policy_completion;
+        time_of_policy_completion.insert(extraction_policy_name.into(), SystemTime::now());
```
diptanu, line 516: This can't be done here, replicated state machines shouldn't have side effects since the clock would vary in every machine when the log is applied and the time will be different. Can you please make sure there is no code like this anywhere else? pass the time in the request, so the time will be in the log.

src/state/store/state_machine_objects.rs
```diff
@@ -413,6 +415,121 @@ impl IndexifyState {
...
     }
 
+    fn set_content_policies_applied_on_content(
+        &self,
+        db: &Arc<OptimisticTransactionDB>,
+        txn: &rocksdb::Transaction<OptimisticTransactionDB>,
+        mappings: &[internal_api::ContentExtractionPolicyMapping],
+    ) -> Result<(), StateMachineError> {
+        //  Fetch all values at once
+        let mapping_cf = StateMachineColumns::ExtractionPoliciesAppliedOnContent.cf(db);
```
diptanu, line 425: Not following how we are doing the key here, there will be exactly one ContentID->PolicyMapping in a given Namespace, why don't you create the Id of the key by hashing ContentId, Namespace and PolicyName ? Then you won't have to iterate through all the keys and such.

diptanu: And you can also look the mapping when you have to update by rehashing stuff from the update request.

diptanu (changes requested):

crates/indexify_internal_api/src/lib.rs
```diff
@@ -374,6 +375,28 @@ impl From<ExtractionPolicy> for indexify_coordinator::ExtractionPolicy {
...
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct ContentExtractionPolicyMapping {
+    pub content_id: String,
+    pub extraction_policy_ids: HashSet<String>, //  NOTE: This is a hash set because the extraction policy should only be applied to a piece of content once
+    pub time_of_policy_completion: HashMap<String, SystemTime>, // policy name -> time instant. This will be written to in the server.rs file when a task is completed
+}
+
+impl Default for ContentExtractionPolicyMapping {
```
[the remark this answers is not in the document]
redixhumayun: This is for testing

src/state/store/state_machine_objects.rs
redixhumayun: Yeah, my bad didn't think of that. Will take care of it

src/state/store/state_machine_objects.rs
redixhumayun: I'm not considering the namespace here. I'm assuming that the content id keys are unique across namespaces, so when multiple tasks are created at once for different content id's, i need to insert mappings for all of them.

redixhumayun: Or are you suggesting storing `Hash(content_id, policy_name, namespace) -> Time of policy application`?

PR #466, opened 2024-03-28
Garbage collection tasks

redixhumayun:
## This PR:

* sets up content deletion
* uses a bi-directional gRPC stream between ingestion server and coordinator
* uses a broadcast channel instead of a watch channel since watch channel results in race condition issue when two tasks are written in quick succession because only latest value is stored on the channel
* walks content tree and removes data from vector stores, blob stores and metadata stores


src/state/store/requests.rs
```diff
@@ -33,6 +36,15 @@ pub enum RequestPayload {
         extractor: internal_api::ExtractorDescription,
         ts_secs: u64,
     },
+    RemoveExecutor {
+        executor_id: String,
+    },
+    RegisterIngestionServer {
```
diptanu, line 42: Do we need this? Just keep the ingestion servers in memory on the leader, and use request forwarding to register and remove ingestion servers and get allocations from leader.

diptanu (changes requested):

src/vectordbs/qdrant.rs
```diff
@@ -127,6 +127,13 @@ impl VectorDb for QdrantDb {
         Ok(())
     }
 
+    #[tracing::instrument]
+    async fn remove_embedding(&self, index: &str, content_id: &str) -> Result<()> {
```
diptanu, line 131: Add implementation?

src/vectordbs/lancedb.rs
```diff
@@ -107,6 +107,28 @@ impl VectorDb for LanceDb {
             .map_err(|e| anyhow!("unable to add to table {}", e))
     }
 
+    #[tracing::instrument]
+    #[tracing::instrument]
+    async fn remove_embedding(&self, index: &str, content_id: &str) -> Result<()> {
```
diptanu, line 112: Test?

src/test_utils.rs
```diff
@@ -164,10 +165,16 @@ impl RaftTestCluster {
...
         let seed_node_id = server_configs.first().unwrap().node_id; //  the seed node will always be the first node in the list
+        let garbage_collector = GarbageCollector::new();
         let mut nodes = BTreeMap::new();
         for config in server_configs {
             let _ = fs::remove_dir_all(config.state_store.clone().path.unwrap());
-            let shared_state = App::new(config.clone(), overrides.clone()).await?;
+            let shared_state = App::new(
+                config.clone(),
+                overrides.clone(),
+                Arc::clone(&garbage_collector),
```
diptanu, line 175: Why are you cloning the reference?

src/task_allocator/planner/load_aware_distributor.rs
```diff
@@ -724,16 +743,11 @@ mod tests {
         let start = Instant::now();
         let result = distributor.plan_allocations(task_ids).await?;
         // stop the timer
-        let duration = start.elapsed();
+        let _duration = start.elapsed();
 
         // Verify that the tasks are allocated
         assert_eq!(result.clone().0.len(), total_tasks);
 
-        println!(
```
diptanu, line 732: Keep it, might be useful to see how long this has taken.

src/task_allocator/planner/load_aware_distributor.rs
```diff
@@ -431,8 +440,13 @@ mod tests {
     async fn test_round_robin_distribution() -> Result<(), anyhow::Error> {
         let config = Arc::new(ServerConfig::default());
         std::fs::remove_dir_all(config.state_store.clone().path.unwrap()).unwrap();
-        let shared_state = App::new(config, None).await.unwrap();
+        let garbage_collector = crate::garbage_collector::GarbageCollector::new();
+        let shared_state = App::new(config, None, Arc::clone(&garbage_collector))
+            .await
+            .unwrap();
         shared_state.initialize_raft().await.unwrap();
+        let _coordinator =
```
diptanu, line 448: why are you creating the coordinator?

src/state/store/state_machine_objects.rs
```diff
@@ -1345,6 +1478,60 @@ impl IndexifyState {
...
     }
 
+    /// This method will fetch all pieces of content metadata for the tree
+    /// rooted at content_id
+    pub fn get_content_tree_metadata(
+        &self,
+        content_id: &str,
+        db: &Arc<OptimisticTransactionDB>,
+    ) -> Result<Vec<indexify_internal_api::ContentMetadata>, StateMachineError> {
+        let txn = db.transaction();
```
diptanu, line 1488: Why is there a transaction? Isn't this just a read?

src/state/store/state_machine_objects.rs
```diff
@@ -1345,6 +1478,60 @@ impl IndexifyState {
...
+                StateMachineError::DatabaseError(format!("Content {} not found", content_id))
+            })?;
+        let root_content =
+            JsonEncoder::decode::<indexify_internal_api::ContentMetadata>(&root_content_bytes)?;
+        collected_content_metadata.push(root_content);
+
+        let mut children_map: HashMap<String, Vec<indexify_internal_api::ContentMetadata>> =
+            HashMap::new();
+
+        let iterator = txn.iterator_cf(
```
diptanu, line 1504: Are we iterating through all the content in rocks to find the children of the root? If we are might be worth recording the tree in which content appear

src/state/store/state_machine_objects.rs
```diff
@@ -1345,6 +1478,60 @@ impl IndexifyState {
...
+                serde_json::from_slice::<indexify_internal_api::ContentMetadata>(&value)
+                    .map_err(StateMachineError::from)?;
+
+            children_map
+                .entry(content_metadata.parent_id.clone())
+                .or_default()
+                .push(content_metadata);
+        }
+
+        let mut queue = VecDeque::new();
```
diptanu, line 1521: What is happening here?

src/state/store/state_machine_objects.rs
```diff
@@ -1005,6 +1138,16 @@ impl IndexifyState {
             RequestPayload::CreateContent { content_metadata } => {
                 self.set_content(db, &txn, content_metadata)?;
             }
+            RequestPayload::TombstoneContent {
+                namespace: _,
+                content_ids,
+            } => {
+                //  TODO: Insert tombstone marker on content here
```
diptanu, line 1145: Whats the todo?

diptanu (changes requested):

src/state/store/state_machine_objects.rs
redixhumayun: Mistake. Already took care of the todo

src/state/store/state_machine_objects.rs
redixhumayun: Walking the tree using a deque

src/state/store/state_machine_objects.rs
redixhumayun: Yeah, plan to bring that in as part of another PR. Will build an in-memory index of `parent_content_id -> Vec<children_content_id>`

src/test_utils.rs
redixhumayun: `Arc::clone` is always called on a reference to an `Arc`. The constructor for `GarbageCollector` returns an `Arc`

src/state/store/state_machine_objects.rs
redixhumayun: Yes, but I'm doing two separate reads - one to get the root content and the other to create an iterator. I want a consistent read

src/coordinator.rs
```diff
@@ -241,19 +280,147 @@ impl Coordinator {
...
+
+            //  If no extraction policies have been applied to the parent of this node, then
+            // it cannot have output features and embeddings
+            if content_extraction_policy_mappings.is_none() {
+                let mut hasher = DefaultHasher::new();
+                namespace.hash(&mut hasher);
+                content_metadata.id.hash(&mut hasher);
+                let id = format!("{:x}", hasher.finish());
+
+                let gc_task = indexify_internal_api::GarbageCollectionTask {
```
diptanu, line 305: Move this to an new() function which creates the id and returns an object.

src/coordinator.rs
```diff
@@ -241,19 +280,147 @@ impl Coordinator {
...
+                output_tables.extend(
+                    applied_extraction_policy
+                        .index_name_table_mapping
+                        .values()
+                        .cloned(),
+                );
+            }
+
+            //  Create the garbage collection task
+            let mut hasher = DefaultHasher::new();
```
diptanu, line 358: These are repeated multiple times. Moving them into the new type will eliminate them.

src/coordinator.rs
```diff
@@ -241,19 +280,147 @@ impl Coordinator {
...
         let state_changes = self.shared_state.unprocessed_state_change_events().await?;
         for change in state_changes {
             info!(
                 "processing change event: {}, type: {}, id: {}",
                 change.id, change.change_type, change.object_id
             );
-            self.scheduler.handle_change_event(change).await?;
+
+            match change.change_type {
+                indexify_internal_api::ChangeType::IngestionServerAdded => {
```
diptanu, line 403: Remove this completely, make the coordinator register unseen ingestion servers when they first send a gc acknowledgment stream

diptanu: Also why are we writing this to the state store? Let's keep these ingestion server related metadata in memory on the leader.

src/coordinator_service.rs
```diff
@@ -279,6 +307,109 @@ impl CoordinatorService for CoordinatorServiceServer {
         }))
     }
 
+    async fn register_ingestion_server(
```
diptanu, line 310: Lets remove this and use the task acknowledgment to register these.

src/state/store/requests.rs
diptanu, line 42: Why is this here?

src/state/store/requests.rs
```diff
@@ -33,6 +36,15 @@ pub enum RequestPayload {
...
+        ingestion_server_metadata: internal_api::IngestionServerMetadata,
+    },
+    RemoveIngestionServer {
```
diptanu, line 45: Why is this here?

src/garbage_collector/mod.rs
```diff
@@ -0,0 +1,207 @@
...
+        }
+    }
+
+    async fn choose_server(&self) -> Option<String> {
+        let servers = self.ingestion_servers.read().await;
+        let mut rng = rand::thread_rng();
+        servers.iter().choose(&mut rng).cloned()
+    }
+
+    async fn watch_deletion_events(
```
diptanu, line 79: How are we ensuring that this happens in the leader? Every coordinator will have state machine changes and thus fire an event.

PR #475, opened 2024-04-03
Update Content

redixhumayun:
* Sets up ability to update root content
* Creates content versioning
* Creates explicit methods that will be used to either fetch latest version of content or explicit version of content
* Checks hash on server when receiving content to determine if the data needs to be updated
* Continuously updates content trees as new content is added


crates/indexify_internal_api/src/lib.rs
```diff
@@ -465,10 +485,81 @@ impl From<ExtractionPolicy> for indexify_coordinator::ExtractionPolicy {
     }
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
+pub struct ContentMetadataId {
```
diptanu, line 489: Create a new() here, version will always be 1, so accept only id. If we ever have to override this for tests, we can create the struct directly.

crates/indexify_internal_api/src/lib.rs
```diff
@@ -465,10 +485,81 @@ impl From<ExtractionPolicy> for indexify_coordinator::ExtractionPolicy {
...
+    pub version: u64,
+}
+
+impl Display for ContentMetadataId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}::v{}", self.id, self.version)
+    }
+}
+
+impl TryFrom<&String> for ContentMetadataId {
```
diptanu, line 500: Repeats what you have below, can't you call the one below by cloning the value? Sell::try_from(value.to_string)

crates/indexify_internal_api/src/lib.rs
```diff
@@ -495,16 +587,48 @@ impl From<ContentMetadata> for indexify_coordinator::ContentMetadata {
...
 }
 
+impl TryFrom<indexify_coordinator::ContentMetadata> for ContentMetadata {
+    type Error = anyhow::Error;
+
+    fn try_from(value: indexify_coordinator::ContentMetadata) -> Result<Self, Self::Error> {
+        Ok(Self {
+            id: ContentMetadataId {
+                id: value.id,
+                version: 1,
```
diptanu, line 603: The id to ContentMetadata doesn't have the version? So we always assume that the API server is working with the latest version? Feels like version: 1 is incorrect, as it needs to point to the latest version or a specific version.

diptanu (changes requested):

crates/indexify_internal_api/src/lib.rs
redixhumayun: But we don't want any component outside of the coordinator to be aware of the version right? So, yes, any request that comes in from the ingestion server will automatically operate on the latest version available. 

All coordinator internal requests will operate on specific versions.

src/state/mod.rs
```diff
@@ -969,71 +900,188 @@ impl App {
...
+        &self,
+        namespace: &str,
+        content_ids: &[String],
+    ) -> Result<(), anyhow::Error> {
         let mut state_changes = vec![];
-        for content in &content_metadata {
+        let mut updated_content_ids = Vec::new();
+
+        for content_id in content_ids.iter() {
+            let latest_version = self
```
diptanu, line 1010: Getting the latest version is the main logic here, everything else is repeated for this function and the one below which does it for a specific version. Can you just call the one below after getting the latest version of every content?

src/state/mod.rs
```diff
@@ -969,71 +900,188 @@ impl App {
...
+                    identical_content.push(new_content);
+                }
+            } else {
+                //  this is fresh content
+                new_incoming_content.push(new_content);
+            }
+        }
+
+        //  write identical content and don't create state changes for it
+        if !identical_content.is_empty() {
```
diptanu, line 934: Why write identical content?

src/state/mod.rs
```diff
@@ -969,71 +900,188 @@ impl App {
...
+            self.forwardable_raft.client_write(req).await.map_err(|e| {
+                anyhow!(
+                    "unable to create identical content metadata: {}",
+                    e.to_string()
+                )
+            })?;
+        }
+
+        //  write the updated content
+        if !content_to_be_updated.is_empty() {
```
diptanu, line 951: All the code here and below is same, why not append the two lists and write in one shot.

diptanu (changes requested):

src/coordinator.rs
```diff
@@ -370,13 +380,56 @@ impl Coordinator {
...
         }
 
         //  this coordinator node is the leader
-        self.create_gc_tasks(&change.object_id).await
+        self.create_gc_tasks(&change).await
+    }
+
+    async fn handle_task_completion_state_change(&self, change: StateChange) -> Result<()> {
+        let content_id: ContentMetadataId = change.object_id.clone().try_into()?;
+        let is_content_processed = self.shared_state.is_content_processed(&content_id).await;
```
diptanu, line 394: Couple of questions - 1. Ehat is the statechage? Task Completion with the object id being the task id or the content metadata id? If it's the later why do you need to know if it's processed? And processed by which extraction policy?

src/coordinator.rs
```diff
@@ -370,13 +380,56 @@ impl Coordinator {
...
+
+        //  previous version needs to be tombstoned
+        if content_id.version <= 1 {
+            //  this content was not updated
+            self.shared_state
+                .mark_change_events_as_processed(vec![change])
+                .await?;
+            return Ok(());
+        }
+        let previous_version = ContentMetadataId {
```
diptanu, line 408: Make a new function for this.

src/coordinator.rs
```diff
@@ -370,13 +380,56 @@ impl Coordinator {
...
+            id: content_id.id,
+            version: content_id.version - 1,
+        };
+        let content_metadata = self
+            .shared_state
+            .get_content_metadata_with_version(&previous_version)
+            .await?;
+        let content_metadata = content_metadata.first().ok_or_else(|| {
```
diptanu, line 416: Just add this to the previous statement.

crates/indexify_internal_api/src/lib.rs
redixhumayun: @diptanu I did this to avoid having to clone values for `ContentMetadataId`. Or rather, I think it would be better to get rid of the one below and just use a reference everywhere. What do you think?

protos/coordinator_service.proto
```diff
@@ -356,7 +356,8 @@ message ContentMetadata {
     string namespace = 8;
     string source = 9;
     uint64 size_bytes = 10;
-    map<string, uint64> extraction_policy_ids = 11;
+    string hash = 11;
```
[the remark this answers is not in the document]
redixhumayun: The content hash that's constructed on the ingestion server from the content frames. Should I call it `content_hash` or something?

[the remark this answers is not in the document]
redixhumayun: This is resolved

crates/indexify_internal_api/src/lib.rs
redixhumayun: `TryFrom<String>` now calls `TryFrom<&String>` internally

src/coordinator.rs
redixhumayun: `ChangeType::TaskCompleted` now has a field called `content_id` and uses `task.id` as `object_id`

diptanu (changes requested):

src/state/store/state_machine_objects.rs
```diff
@@ -951,12 +1148,12 @@ impl IndexifyState {
...
                 StateMachineError::DatabaseError(format!(
-                    "No content policies applied on content found for id {}",
+                    "Content not found while updating applied extraction policies {}",
                     content_id
                 ))
             })?;
-        let content_meta = JsonEncoder::decode::<internal_api::ContentMetadata>(&value)?;
-        let _epoch_time = policy_completion_time
+        let mut content_meta = JsonEncoder::decode::<internal_api::ContentMetadata>(&value)?;
+        let epoch_time = policy_completion_time
```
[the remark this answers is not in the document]
redixhumayun: @diptanu it looks like you've already taken care of this. The timestamp is being set in `update_task` in `state/mod.rs`

src/coordinator.rs
redixhumayun: I can't do that since a Vec returned from a function is a temp value

src/state/mod.rs
redixhumayun: Yeah, `tombstone_content_batch` now calls `tombstone_content_batch_with_version`

PR #477, opened 2024-04-03
implement storing metadata alongside vector

maxkozlovsky:
Add metadata to add_embedding() arguments, store it together with the vector.

Add get_points() and update_metadata() functions to VectorDb to retrieve metadata and update it.

When creating content for the first time, send metadata together with the embedding.

If updating existing content, retrieve existing metadata and combine it with the new one. If embedding is also updated, send updated metadata with add_embedding(). If only metadata is updated, use update_metadata() to change existing vector.


diptanu (changes requested):

src/data_manager.rs
```diff
@@ -500,15 +515,90 @@ impl DataManager {
         Ok(())
     }
 
+    fn combine_extracted_metadata(metadata: &[ExtractedMetadata]) -> serde_json::Value {
```
diptanu, line 518: Looks like it's transforming serve_json::Value to serde_json::Value. Why?

src/data_manager.rs
```diff
@@ -500,15 +515,90 @@ impl DataManager {
...
+        let mut combined_metadata = serde_json::Map::new();
+        for m in metadata {
+            for (k, v) in m.metadata.as_object().unwrap() {
+                combined_metadata.insert(k.clone(), v.clone());
+            }
+        }
+        serde_json::Value::Object(combined_metadata)
+    }
+
+    fn combine_metadata(
```
diptanu, line 528: Is new metadata for content combined to make it easy to update the metadata of the vectors in vector store?

src/data_manager.rs
```diff
@@ -500,15 +515,90 @@ impl DataManager {
...
+        content_meta: &indexify_coordinator::ContentMetadata,
+        features: Vec<api::Feature>,
+        output_index_map: &HashMap<String, String>,
+        index_tables: &[String],
+    ) -> Result<()> {
+        let existing_metadata = self
+            .metadata_index_manager
+            .get_metadata_for_content(&content_meta.namespace, &content_meta.id)
+            .await?;
+        let existing_metadata = Self::combine_extracted_metadata(&existing_metadata);
```
diptanu, line 562: What is this for? Looks like we are combining things below again.

diptanu (changes requested):

protos/coordinator_service.proto
```diff
@@ -218,6 +218,7 @@ message Task {
     string extraction_policy_id = 6;
     map<string, string> output_index_mapping = 7;
     TaskOutcome outcome = 8;
+    repeated string index_tables = 9;
```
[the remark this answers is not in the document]
maxkozlovsky: index mapping includes only the mapping for the names used by this extractor, with the index being the map value. index_tables is all indices that may include the content. We could reuse output_index_mapping, but we'll have to come up with fake unique names for keys, which doesn't feel clean.

src/data_manager.rs
maxkozlovsky: It combines a list of jsons  into a single one. Single json is the lowest common denominator since qdrant only supports one json field per embedding.

PR #479, opened 2024-04-04
start server after coordinator server is started in dev mode

maxkozlovsky:
Since server is trying to connect to coordinator server during initalization, start coordinator server first.


diptanu:
@maxkozlovsky instead of changing the order let’s make the server keep retrying to reach the coordinator

PR #481, opened 2024-04-04
Delete content changes

redixhumayun:
## This PR

* moves broadcast sender for task deletion events to state machine
* refactors garbage collector to be simpler
* adds documentation for rag with deletion
* stores node and coordinator address in RocksDB
* uses forwardable coordinator


src/state/store/state_machine_objects.rs
```diff
@@ -470,7 +463,49 @@ impl fmt::Display for IndexifyState {
     }
 }
 
+#[derive(Clone)]
+pub struct IndexifyState {
```
diptanu, line 467: Why wrap the data in state? Initialize the IndexifyState with rx, tx, and pass the tx to the state machine. And whenever the coordinator requests for the rx, clone and hand it out. This abstraction doesn't feel necessary just to wrap the internal state machine store and the sender.

protos/raft.proto
```diff
@@ -19,5 +19,8 @@ service RaftApi {
     rpc InstallSnapshot(RaftRequest) returns (RaftReply);
     rpc Vote(RaftRequest) returns (RaftReply);
     rpc JoinCluster(RaftRequest) returns (RaftReply);
+    rpc RegisterIngestionServer(RaftRequest) returns (RaftReply);
```
diptanu, line 22: Can we move them to the Coordinator Service GRPC API?

diptanu (changes requested):

protos/raft.proto
redixhumayun: But that's where I had it initially. Why move it back again?

src/state/store/state_machine_objects.rs
redixhumayun: That's what I tried initially. Ran into issues with serialisation and de-serialisation. Serde skip directive didn't work. 

It's easy to change this in the future cause this file is self-contained

PR #495, opened 2024-04-09
use buffered writes

maxkozlovsky:


diptanu (changes requested):
Can you please fix the test? Looks fine to me. Did you benchmark the before   and after write throughput of this change?

maxkozlovsky:
I'll merge the test separately it has dependency on the other pending changes.

PR #496, opened 2024-04-10
Extraction graph

diptanu:
## This PR

* introduces extraction graph modeling
* now only content that is added along with an extraction graph will run through extractors


protos/coordinator_service.proto
```diff
@@ -325,6 +334,7 @@ message ExtractionPolicy {
     string input_params = 4;
     map<string, string> filters = 5;
     string content_source = 6;
+    string graph_name = 7;
```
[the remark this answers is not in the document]
redixhumayun: We decided not to expose id's outside the coordinator. We should maintain that because it makes for a simpler mental model vs exposing some thing as id's and some things as names. 

Also, we really need to separate the interface between the coordinator <-> executor and coordinator <-> ingestion. Right now we have a distributed monolith not a distributed system because we reuse types all over the place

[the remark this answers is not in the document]
diptanu: I see, I guess this is translated to the user facing ExtractorPolicy. Lets add the namespace in that case to this struct.

src/server.rs
```diff
@@ -872,7 +914,17 @@ async fn upload_file(
...
+                .upload_file(
+                    &namespace,
+                    stream,
+                    &name,
+                    content_mime,
+                    labels,
+                    Some(&id),
+                    params
+                        .extraction_graph_names
+                        .unwrap_or_else(std::vec::Vec::new),
```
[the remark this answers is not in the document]
redixhumayun: No, I'm not going to allow that in the new model. They can't upload content without creating and specifying extraction graphs. I'm going to force the client to error out in that case. I'll add server side checks as well.

crates/indexify_internal_api/src/lib.rs
```diff
@@ -607,50 +787,60 @@ impl Default for ContentMetadataId {
...
-    // Namespace name == Namespace ID
-    pub namespace: String,
+    pub namespace: NamespaceName,
     pub name: String,
     pub content_type: String,
     pub labels: HashMap<String, String>,
     pub storage_url: String,
     pub created_at: i64,
-    pub source: String,
+    pub source: Vec<ContentMetadataSource>,
```
diptanu, line 828: Why was this replaced by a Vec<T> and not ContentMetadataSource ? A content can is produced by a single extraction policy.

src/coordinator.rs
```diff
@@ -53,17 +58,200 @@ impl Coordinator {
...
+        let mut content_meta_list = Vec::new();
+        for content in content_list {
+            let extraction_graphs = self
+                .shared_state
+                .get_extraction_graphs(&content.extraction_graph_ids)?
+                .ok_or_else(|| {
+                    anyhow!("could not find extraction graph for content {}", content.id)
+                })?;
+            let extraction_graph_names = extraction_graphs.into_iter().map(|eg| eg.name).collect();
+            let sources: Result<Vec<_>> = content
```
diptanu, line 75: Not necessary if we assume a content is produced by a single source.

src/coordinator.rs
```diff
@@ -319,6 +535,52 @@ impl Coordinator {
...
+    ) -> Result<Vec<internal_api::Index>> {
+        let structured_data_schema = StructuredDataSchema::default();
+        let mut indexes_to_create = Vec::new();
+        for extraction_policy in &extraction_policies {
+            let extractor = self.get_extractor(&extraction_policy.extractor)?;
+            extractor.validate_input_params(&extraction_policy.input_params)?;
+            for (output_name, output_schema) in extractor.outputs {
+                match output_schema {
+                    OutputSchema::Embedding(embeddings) => {
+                        let mut index_to_create = internal_api::Index {
```
diptanu, line 551: Move this to a builder method inside the Index struct

src/coordinator.rs
```diff
@@ -319,6 +535,52 @@ impl Coordinator {
...
+            }
+        }
+        self.shared_state
+            .create_extraction_graph(
+                extraction_graph,
+                extraction_policies,
+                structured_data_schema,
+                indexes_to_create.clone(),
+            )
+            .await?;
```
diptanu, line 580: What happens if the graph exists but ingestion server couldn't create the indexes? The client will retry, and this will throw an error, the ingestion server will not know about the indexes it was supposed to create. The system would be in a state where the graph would exist but indexes will never be created. 

Return the indexes here if the graph exists.

src/server.rs
```diff
@@ -566,6 +570,34 @@ async fn create_extraction_policy(
     Ok(Json(ExtractionPolicyResponse { index_names }))
 }
 
+#[utoipa::path(
+    post,
+    path = "/namespace/{namespace}/extraction_policies",
+    request_body = ExtractionPolicyRequest,
+    tag = "indexify",
```
maxkozlovsky, line 577: this refers to policies instead of graph

src/state/mod.rs
```diff
@@ -371,43 +376,45 @@ impl App {
...
-                .cloned()
-                .unwrap_or_default()
-        };
-        let extraction_policies = self
-            .state_machine
-            .get_extraction_policies_from_ids(extraction_policy_ids)
-            .await?
-            .unwrap_or_else(Vec::new);
-
+        println!("content_metadata: {:?}", content_metadata);
```
maxkozlovsky, line 387: letfover println! s

crates/indexify_internal_api/src/lib.rs
```diff
@@ -608,23 +800,79 @@ impl Default for ContentMetadataId {
...
     pub created_at: i64,
-    pub source: String,
+    pub source: ContentSource,
     pub size_bytes: u64,
     pub tombstoned: bool,
     pub hash: String,
-    pub extraction_policy_ids: HashMap<String, u64>,
+    pub extraction_policy_ids: HashMap<ExtractionPolicyId, u64>, /*  map of completion time for
+                                                                  * each extraction policy id */
+    pub extraction_graph_names: Vec<ExtractionGraphName>,
```
maxkozlovsky, line 875: Does this need to be stored in database? It only is used when new content is created to find policies, right?

src/api.rs
```diff
@@ -425,13 +443,44 @@ impl From<indexify_internal_api::ContentMetadata> for ContentMetadata {
...
+    pub extraction_policy_id: String,
+    pub output_index_table_mapping: HashMap<String, String>,
+    pub namespace: String,
+    pub content_metadata: ContentMetadata,
+    pub input_params: serde_json::Value,
+    pub outcome: i32,
+    pub index_tables: Vec<String>,
+}
+
+impl From<indexify_coordinator::Task> for Task {
```
maxkozlovsky, line 466: when is this necessary? We shouldn't receive tasks from rpcs.

crates/indexify_internal_api/src/lib.rs
```diff
@@ -6,23 +6,75 @@ use std::{
...
+    pub fn create_id(name: &str, namespace: &str) -> String {
+        let mut s = DefaultHasher::new();
+        name.hash(&mut s);
+        namespace.hash(&mut s);
+        format!("{:x}", s.finish())
+    }
+}
+
+impl ExtractionGraphBuilder {
+    pub fn build(&self) -> Result<ExtractionGraph> {
```
maxkozlovsky, line 41: self can be consumed instead of passing reference this may avoid extra clones

src/coordinator.rs
```diff
@@ -55,6 +58,59 @@ impl Coordinator {
...
+    }
+
+    pub fn internal_namespace_to_external(
+        &self,
+        namespaces: Vec<internal_api::Namespace>,
+    ) -> Result<Vec<indexify_coordinator::Namespace>> {
+        let mut namespaces_to_return: Vec<indexify_coordinator::Namespace> = Vec::new();
+        for namespace in namespaces {
+            let extraction_policies =
+                self.internal_extraction_policy_to_external(namespace.extraction_policies.clone())?;
```
maxkozlovsky, line 103: how does this fail, if we have valid internal structure it should always be possible to produce external structure.

crates/indexify_internal_api/src/lib.rs
diptanu: Good point.

src/coordinator.rs
diptanu: Good point

PR #507, opened 2024-04-14
implement server and coordianator metrics

maxkozlovsky:
Add metrics for number and bytes of uploaded and extracted contents.

Add metrics for tasks completed/failed, current number of tasks and executors in progress.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,7 @@ openraft = { git = "https://github.com/datafuselabs/openraft.git", rev = "bde63c
 opensearch = { version = "2", default_features = false, features = [
     "rustls-tls",
 ] }
-opentelemetry = { version = "0.20", features = ["rt-async-std"] }
+opentelemetry = { version = "0.22", features = ["metrics"] }
 pgvector = { version = "0.3", features = ["sqlx"] }
 prost = { version = "0.12" }
 prost-types = { version = "0.12" }
@@ -200,6 +200,9 @@ pin-project-lite = "0.2.13"
 gluesql = {workspace=true, default-features=false}
 uuid = "1.8.0"
 lancedb = {version = "0.4.15", default_features = false}
+opentelemetry-prometheus = "0.15"
+prometheus = "0.13"
+opentelemetry_sdk = { version = "0.22", features = ["metrics"] }
 
 [dev-dependencies]
 tracing-test = { version = "0.2", features = ["no-env-filter"] }
--- a/src/coordinator_service.rs
+++ b/src/coordinator_service.rs
@@ -816,7 +816,7 @@ async fn metrics_handler(
     )))
 }
 
-fn start_server(app: &CoordinatorServer) -> JoinHandle<Result<()>> {
+fn start_server(app: &CoordinatorServer) -> Result<JoinHandle<Result<()>>> {
     let server = axum::Router::new()
         .route("/metrics", get(metrics_handler))
         .with_state(app.shared_state.clone());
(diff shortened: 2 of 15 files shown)


Cargo.toml
```diff
@@ -52,6 +52,7 @@ opensearch = { version = "2", default_features = false, features = [
 ] }
 opentelemetry = { version = "0.20", features = ["rt-async-std"] }
 pgvector = { version = "0.3", features = ["sqlx"] }
+prometheus = "0.12.0"
```
diptanu, line 55: We use opentelemetry, isn’t it possible to use that?

diptanu (changes requested):

src/state/mod.rs
```diff
@@ -124,12 +131,47 @@ pub struct App {
...
+    let mut buffer = vec![];
+    let encoder = prometheus::TextEncoder::new();
+    encoder.encode(&metric_families, &mut buffer).unwrap();
+
+    Ok(axum::response::Response::new(axum::body::Body::from(
+        buffer,
+    )))
+}
+
+fn start_server(app: Arc<App>, server_config: &ServerConfig) -> JoinHandle<Result<()>> {
```
diptanu, line 155: Please move this to the coordinator_server.rs so that the same shutdown handler shuts this server, along with the grpc server.

Or move this into a separate module, but start and stop it in coordinator_service.rs

src/state/mod.rs
```diff
@@ -210,6 +252,8 @@ impl App {
         ));
         let (leader_change_tx, leader_change_rx) = tokio::sync::watch::channel::<bool>(false);
 
+        let metrics = Metrics::new(state_machine.clone());
```
diptanu, line 255: Isn't there a global metrics holder/registry that you can pass around to different modules in the service?

src/state/store/state_machine_objects.rs
```diff
@@ -1080,6 +1119,15 @@ impl IndexifyState {
                 if task.terminal_state() {
                     //  If the task is meant to be marked finished and has an executor id, remove it
                     // from the list of tasks assigned to an executor
+                    match task.outcome {
```
diptanu, line 1122: Move this code as a function inside the Metrics module and pass the TaskOutcome as a reference to mutate the counters .

diptanu (changes requested):

src/state/mod.rs
maxkozlovsky: This is the same file where the grpc server is started. The shutdown is handled below, the app.server_handle is used to shutdown.

maxkozlovsky: ok moved the functions to coordinator_service.rs

src/state/store/state_machine_objects.rs
maxkozlovsky: moved into a function

src/state/mod.rs
maxkozlovsky: there is a global one, I don't know if it better or necessary to use it.

src/coordinator_service.rs
```diff
@@ -795,6 +799,43 @@ pub struct CoordinatorServer {
...
+fn start_server(app: &CoordinatorServer) -> JoinHandle<Result<()>> {
+    let server = axum::Router::new()
+        .route("/metrics", get(metrics_handler))
+        .with_state(app.shared_state.clone());
+    let addr: SocketAddr = format!(
+        "{}:{}",
+        app.config.listen_if, app.config.coordinator_http_port
+    )
+    .parse()
+    .unwrap();
```
diptanu, line 828: Unwrap is probably not going to work here. Can we return an error and make the startup fail completely. Please test by specifying something wrong.

diptanu (changes requested):

src/coordinator_service.rs
maxkozlovsky: added check

PR #514, opened 2024-04-18
add test for policy filtering

maxkozlovsky:


diptanu:
Looks like the code for filtering is in state/mod.rs - can you move the test there so that we test those methods instead of adding a test here?

maxkozlovsky:
> Looks like the code for filtering is in state/mod.rs - can you move the test there so that we test those methods instead of adding a test here?

That's how I started, but creating tasks requires running scheduler which is connected through coordinator, so the test has to be there.

PR #521, opened 2024-04-22
allow specifying labels in http data

maxkozlovsky:
Retrieve labels from http metadata and add them to content metadata.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -783,56 +783,74 @@ async fn upload_file(
     State(state): State<NamespaceEndpointState>,
     mut files: Multipart,
 ) -> Result<(), IndexifyAPIError> {
-    while let Some(file) = files.next_field().await.unwrap() {
-        let name = file
-            .file_name()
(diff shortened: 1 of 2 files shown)


src/server.rs
```diff
@@ -783,58 +783,68 @@ async fn upload_file(
...
-                    StatusCode::BAD_REQUEST,
-                    &format!("failed to create content for file: {}", e),
-                )
-            })?;
-        state.metrics.node_content_uploads.add(1, &[]);
-        state
-            .metrics
-            .node_content_bytes_uploaded
-            .add(size_bytes, &[]);
+            labels.insert(name, value);
```
diptanu, line 841: This will work if the labels are added first to the map and at the end the file is added. You could have uploaded the file in the loop, and accumulated the labels. The content metadata could be written outside the loop. It would require creating an optional ContentMetadata at the top. I will merge this but let's come back and take care of this at some point.

PR #524, opened 2024-04-23
Tls config

redixhumayun:


src/server.rs
```diff
@@ -321,10 +322,25 @@ impl Server {
...
-            .serve(app.into_make_service())
-            .await?;
+        if use_tls {
+            if let Some(tls_config) = self.config.tls.clone() {
+                let config = build_mtls_config(&tls_config)?;
+                let rustls_config = RustlsConfig::from_config(config);
+                axum_server::tls_rustls::bind_rustls(self.addr, rustls_config)
+                    .handle(handle)
+                    .serve(app.into_make_service())
+                    .await
```
maxkozlovsky, line 332: should handle error instead of unwrap.

diptanu:
Update the docs also in this PR. I want to see the user experience of setting this up.

PR #533, opened 2024-04-24
add root id to content metadata

maxkozlovsky:
Add root id and propagate it to children on content extraction.


crates/indexify_internal_api/src/lib.rs
```diff
@@ -582,6 +582,7 @@ impl Default for ContentMetadataId {
 pub struct ContentMetadata {
     pub id: ContentMetadataId,
     pub parent_id: ContentMetadataId,
+    pub root_id: ContentMetadataId,
```
diptanu, line 585: Can we please rename this to root_content_id?

diptanu (changes requested):

crates/indexify_internal_api/src/lib.rs
maxkozlovsky: ok renamed

crates/indexify_internal_api/src/lib.rs
```diff
@@ -628,6 +630,10 @@ impl TryFrom<indexify_coordinator::ContentMetadata> for ContentMetadata {
                 id: value.parent_id,
                 version: 1,
             },
+            root_content_id: ContentMetadataId {
```
redixhumayun, line 633: @maxkozlovsky also need to add root_content_id when we create or update content in `create_content_batch`. If content is updated and the parent pointer needs to flip, even the `root_content_id` will have to change.

maxkozlovsky: root can't really change because the parent id value remains the same even if data associated with content id changes

PR #544, opened 2024-04-25
handle errors during ingest extracted content

maxkozlovsky:
If any error occurred during extracted content ingest, propagate it back to the client.

### DIFF
--- a/src/ingest_extracted_content.rs
+++ b/src/ingest_extracted_content.rs
@@ -1,6 +1,7 @@
 use std::collections::HashMap;
 
 use anyhow::{anyhow, Result};
+use axum::extract::ws;
 use axum_typed_websockets::{Message, WebSocket};
 use indexify_proto::indexify_coordinator;
 use sha2::{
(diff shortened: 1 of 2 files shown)


src/ingest_extracted_content.rs
```diff
@@ -277,52 +278,66 @@ impl IngestExtractedContentState {
...
+                            let msg = match res {
+                                Ok(_) => IngestExtractedContentResponse::Success,
+                                Err(e) => IngestExtractedContentResponse::Error(e.to_string()),
+                            };
+                            let _ = socket.send(Message::Item(msg)).await;
+                            break;
                         }
-                        return;
+                    };
+                    if let Err(e) = res {
```
diptanu, line 313: We need information on which message we were not able to handle

src/ingest_extracted_content.rs
```diff
@@ -277,52 +278,66 @@ impl IngestExtractedContentState {
...
+                        tracing::error!("Error handling message {:?}", e);
+                        let _ = socket
+                            .send(Message::Close(Some(ws::CloseFrame {
+                                code: 1002,
```
diptanu, line 317: What is this number? Pls comment and extract into a constant.

diptanu (changes requested):

src/ingest_extracted_content.rs
maxkozlovsky: added type to trace

src/ingest_extracted_content.rs
maxkozlovsky: added constant

PR #548, opened 2024-04-26
register executor with list of extractors

edwinkys:
This PR changes the infrastructure from registering Executor with an Extractor to accepting Executor with multiple extractors.

Along with: https://github.com/tensorlakeai/indexify-extractors/pull/149

To test:

1. Run the server.
2. Register executor with multiple extractors from the SDK PR above.
3. Add content or launch a task somehow.


src/task_allocator/planner/load_aware_distributor.rs
```diff
@@ -167,22 +167,24 @@ impl LoadAwareDistributor {
...
-                        .or_default()
-                        // use `Reverse` here to make it a min-heap
-                        .push(Reverse(ExecutorLoad {
-                            executor_id: executor_id.clone(),
-                            running_task_count,
-                        }));
+                    for extractor_name in extractor_names {
+                        executors_load_min_heap
+                            .entry(extractor_name)
+                            .or_default()
```
maxkozlovsky, line 185: Don't think this will work, the calling function assumes that single executor can be present only in one heap. If this is the case, assigning load to an executor may cause multiple heaps to be rearranged which the caller does not do. Maybe better to have a single heap and search for the next available executor matching extractor type. T

edwinkys: Thanks for feedback. I will check this out.
I'm currently testing this from the client and the extractor side. I'll get back to this ASAP.

edwinkys: As discussed with Diptanu, this functionalities will be optimize in the near future.

maxkozlovsky (approved):
looks good.

PR #555, opened 2024-04-30
ignore executor does not exist when removing

maxkozlovsky:
This can happen if the extractor retries unregister request, ignore does not exist error.

### DIFF
--- a/src/state/store/state_machine_objects.rs
+++ b/src/state/store/state_machine_objects.rs
@@ -961,23 +961,22 @@ impl IndexifyState {
         db: &Arc<OptimisticTransactionDB>,
         txn: &rocksdb::Transaction<OptimisticTransactionDB>,
         executor_id: &str,
-    ) -> Result<internal_api::ExecutorMetadata, StateMachineError> {
+    ) -> Result<Option<internal_api::ExecutorMetadata>, StateMachineError> {
         //  Get a handle on the executor before deleting it from the DB
         let executors_cf = StateMachineColumns::Executors.cf(db);
-        let serialized_executor = txn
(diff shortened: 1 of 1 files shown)


src/state/store/state_machine_objects.rs
```diff
@@ -200,6 +200,13 @@ impl ExtractorExecutorsTable {
             .remove(executor_id);
     }
 
+    pub fn remove_by_executor(&self, executor_id: &ExecutorId) {
```
diptanu, line 203: Add some comments here?

src/state/store/state_machine_objects.rs
```diff
@@ -1252,8 +1260,12 @@ impl IndexifyState {
                     .map_err(|e| StateMachineError::TransactionError(e.to_string()))?;
 
                 //  Remove the the extractor from the executor -> extractor mapping table
-                self.extractor_executors_table
-                    .remove(&executor_meta.extractor.name, &executor_meta.id);
+                if let Some(executor_meta) = executor_meta {
+                    self.extractor_executors_table
+                        .remove(&executor_meta.extractor.name, &executor_meta.id);
+                } else {
+                    self.extractor_executors_table.remove_by_executor(executor_id);
```
diptanu, line 1267: why is this necessary? A single executor belongs to only one extractor right now. When do we need to remove the executor entry from all the extractor -> executor table?

diptanu (changes requested):
Fix the tests and please add comments.

src/state/store/state_machine_objects.rs
maxkozlovsky: Ok removed trying remove by id

PR #556, opened 2024-05-01
added metrics for tasks per executor

redixhumayun:
Note: additional files have changed as a result of running `make fmt && cargo clippy --fix --allow-dirty`

### DIFF
--- a/src/metrics/mod.rs
+++ b/src/metrics/mod.rs
@@ -605,7 +605,7 @@ pub mod coordinator {
                         let counts = app.data.indexify_state.executor_running_task_count.inner();
                         for (executor_id, count) in counts.iter() {
                             observer.observe(
-                                (*count).try_into().unwrap(),
+                                *count,
                                 &[KeyValue::new("executor_id", executor_id.to_string())],
                             );
                         }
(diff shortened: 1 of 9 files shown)


maxkozlovsky:
There is  pub executor_running_task_count: ExecutorRunningTaskCount, field already, I think it is equivalent to the new one.

redixhumayun:
> There is pub executor_running_task_count: ExecutorRunningTaskCount, field already, I think it is equivalent to the new one.

@maxkozlovsky yeah, you're right. I've changed the code. Thanks for pointing it out

src/metrics/mod.rs
```diff
@@ -593,6 +597,23 @@ pub mod coordinator {
...
 
+            let tasks_per_executor = meter
+                .u64_observable_gauge("indexify.coordinator.tasks_per_executor")
+                .with_callback({
+                    let app = app.clone();
+                    move |observer| {
+                        let counts = app.data.indexify_state.executor_running_task_count.inner();
+                        for (executor_id, count) in counts.iter() {
+                            observer.observe(
+                                (*count).try_into().unwrap(),
```
diptanu, line 608: Can we change the data type for executor running task count and not do unwrap here please? Everything else looks fine.

diptanu (changes requested):

src/metrics/mod.rs
redixhumayun: @diptanu this is done

PR #562, opened 2024-05-03
feat: add option to include/exclude download blob

edwinkys:
This PR adds an optional parameter when performing semantic search on the vector database to exclude or include the content stored in the blob storage.


diptanu:
This is perfect!

PR #575, opened 2024-05-08
Raft snapshot rework

redixhumayun:
## This PR:

* correctly builds the snapshot from the RocksDB forward indexes
* installs the snapshot into the RocksDB forward indexes and builds the in-memory reverse indexes
* adds a test

Tested with 50k small document uploads. No errors popped up and snapshot regularly installs on both leader & follower. 

However, indexify takes up a lot of memory by the end, roughly around 1GB for the leader node and ~750MB for the follower

For a single node system and the same workload, the memory usage is ~750MB

Screenshot 2024-05-15 at 6 58 21 PM

indexify-mem-usage-single-node

## UPDATE

Tested again with 30k documents after removing the `rocksdb` transaction from the write path of the snapshot. It's helped significantly. Memory difference between the two branches is now down to 120MB

First image is memory usage for the snapshot branch with snapshots occurring for every 1000 logs

snapshot_branch_mem

Below is memory usage for the main branch with same snapshot policy

main_branch_mem

## AFTER OPENRAFT UPDATE

Content upload on main gives below memory usage

after_merge_main_branch_mem

Content upload on snapshot branch gives below memory usage

after_merge_snapshot_branch_mem


redixhumayun:
@diptanu had to make changes to configs on the following

* set size limit on `gRPC` server and client to 10MB (up from default of 4MB), otherwise we run into size limits with snapshots.
* set `install_snapshot_timeout` of 2 seconds on `OpenRaft` config so that larger snapshots can be sent. 

I think both of these should be configurable via the `local_server_config.yml` file and I'd like to add them there. Let me know your thoughts.

src/main.rs
```diff
@@ -153,14 +152,16 @@ fn setup_tracing(trace_type: &str) -> Result<()> {
...
 fn setup_fmt_tracing() {
-    let subscriber = tracing_subscriber::Registry::default().with(
-        tracing_subscriber::fmt::layer()
-            .with_writer(std::io::stderr)
-            .with_filter(LevelFilter::from_level(Level::INFO)),
-    );
-    if let Err(e) = tracing::subscriber::set_global_default(subscriber) {
-        eprintln!("failed to set global default subscriber: {}", e);
-    }
+    let env_filter = tracing_subscriber::EnvFilter::try_from_default_env()
```
diptanu, line 155: Do we need all this? Guessing something was not working with the existing tracing code?

src/state/grpc_config.rs
```diff
@@ -0,0 +1,10 @@
+pub struct GrpcConfig {}
+
+impl GrpcConfig {
```
diptanu, line 3: What is this for? We want to send up to 16 MB?

src/state/mod.rs
```diff
@@ -155,6 +157,8 @@ impl App {
             election_timeout_min: 1500,
             election_timeout_max: 3000,
             enable_heartbeat: true,
+            install_snapshot_timeout: 2000,
+            snapshot_max_chunk_size: 4194304, //  4MB
```
diptanu, line 161: This is 4 MB but the grpc config is 16.

src/state/network.rs
```diff
@@ -1,6 +1,7 @@
 use std::{error::Error, fmt::Display, sync::Arc};
 
 use anyerror::AnyError;
+use anyhow::Ok;
```
diptanu, line 4: Not necessary

src/state/store/state_machine_objects.rs
```diff
@@ -2188,24 +2191,169 @@ impl IndexifyState {
...
-            executor_running_task_count: self.get_executor_running_task_count(),
-            schemas_by_namespace: self.get_schemas_by_namespace(),
-            content_children_table: self.get_content_children_table(),
-            pending_tasks_for_content: self.get_pending_tasks_for_content(),
-        }
-    }
-
-    pub fn install_snapshot(&self, snapshot: IndexifyStateSnapshot) {
-        let mut unassigned_tasks_guard = self.unassigned_tasks.unassigned_tasks.write().unwrap();
+    pub fn build_snapshot(
```
diptanu, line 2194: We need ExtractionGraphs to be added to the snapshot.

Remove ExtractionPoliciesAppliedOnContent since we don't seem to use it.

diptanu (changes requested):

src/state/store/mod.rs
```diff
@@ -478,9 +478,11 @@ impl StateMachineStore {
     where
         V: DeserializeOwned,
     {
+        let txn = self.db.transaction();
```
diptanu, line 481: Where is the txn closed ?

src/state/store/state_machine_objects.rs
```diff
@@ -2188,24 +2191,169 @@ impl IndexifyState {
...
+        &self,
+        db: &Arc<OptimisticTransactionDB>,
+    ) -> Result<IndexifyStateSnapshot, StateMachineError> {
+        let txn = db.transaction();
```
diptanu, line 2198: What happens to this transaction and others? They need to be committed or closed or something no?

src/main.rs
```diff
@@ -153,14 +152,16 @@ fn setup_tracing(trace_type: &str) -> Result<()> {
 }
 
 fn setup_fmt_tracing() {
-    let subscriber = tracing_subscriber::Registry::default().with(
-        tracing_subscriber::fmt::layer()
-            .with_writer(std::io::stderr)
-            .with_filter(LevelFilter::from_level(Level::INFO)),
```
diptanu, line 159: .with_filter(env_filter)

// Initialize the env_filter before this.

src/state/store/state_machine_objects.rs
redixhumayun: Done

src/state/mod.rs
redixhumayun: Databend configs

src/main.rs
redixhumayun: Done

src/state/grpc_config.rs
redixhumayun: Databend configs

src/main.rs
redixhumayun: Done

src/state/store/state_machine_objects.rs
redixhumayun: Not relevant. No txn in read path anymore

src/state/network.rs
redixhumayun: done

maxkozlovsky (commented):
There is a problem with the implementation that it creates an in-memory copy of the whole database and serializes it into on-disk file. This does not scale with the database size, and eventually will fail the snapshot when database does not fit in memory. Snapshot  should be using rocksdb checkpoint, which is much cheaper to create and does not require storing all db in memory (rocksdb creates hard links to all the level files in checkpoint directory, next time a level file is overwritten with write merge checkpoint retains the state) . To transfer snapshot transfer all files in the rocksdb checkpoint directory, and replace rocksdb dir with that.

PR #589, opened 2024-05-14
modify key construction scheme to allow direct lookup

maxkozlovsky:
Store the latest version of content with key made from just id. This allows direct lookup of content without prefix search. This applies to root content only, as children content are never overwritten, they are either identical and don't change or a new id is allocated.


crates/indexify_internal_api/src/lib.rs
```diff
@@ -900,6 +881,31 @@ pub struct ContentMetadata {
...
+impl ContentMetadata {
+    pub fn get_root_id(&self) -> &str {
+        self.root_content_id.as_ref().unwrap_or(&self.id.id)
+    }
+
+    // Return key to store structure in k/v store. The latest version of root and
+    // children are stored with id as key (children always have version 1 and
+    // are never overwritten). Overwritten or deleted roots keys are
+    // formed from id and version.
+    pub fn id_key(&self) -> String {
```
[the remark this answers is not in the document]
maxkozlovsky: This not a general to_string(), this is specifically for making a key.

src/coordinator.rs
```diff
@@ -84,15 +83,11 @@ impl Coordinator {
...
         labels_eq: &HashMap<String, String>,
     ) -> Result<Vec<internal_api::ContentMetadata>> {
-        let content = self
-            .shared_state
-            .list_content(namespace)
-            .await?
-            .into_iter()
-            .flatten()
-            .collect_vec();
-        let filtered_list = list_content_filter(content, source, parent_id, labels_eq);
```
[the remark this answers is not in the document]
maxkozlovsky: It is called from list_content http api

src/state/store/mod.rs
```diff
@@ -395,14 +403,34 @@ impl StateMachineStore {
...
+    pub fn list_content(
         &self,
-        content_ids: Vec<ContentMetadataId>,
-    ) -> Result<Vec<Option<ContentMetadata>>> {
-        self.data
-            .indexify_state
-            .get_content_from_ids_with_version(content_ids, &self.db)
-            .map_err(|e| anyhow::anyhow!(e))
+        namespace: &str,
+        parent_id: &str,
```
diptanu, line 409: Should it not be just id ? Parent ID implies that we are looking for list of content with parent id

src/state/store/state_machine_objects.rs
```diff
@@ -705,13 +711,19 @@ impl IndexifyState {
         &self,
         db: &Arc<OptimisticTransactionDB>,
         txn: &rocksdb::Transaction<OptimisticTransactionDB>,
-        state_changes: &Vec<StateChange>,
+        state_changes: &mut Vec<StateChange>,
     ) -> Result<(), StateMachineError> {
+        let mut change_id = self.get_next_change_ids(state_changes.len());
```
diptanu, line 716: This needs to be done outside of the state machine, since you don't usually want side effects being made at the time of application of logs on raft nodes. The proposer should allocate the ids and such and apply the logs.

diptanu (changes requested):

src/state/store/mod.rs
maxkozlovsky: that's what it does looks for content with parent_id, I did not change the api. Does not look very useful to me either.

src/state/store/state_machine_objects.rs
maxkozlovsky: Does not look possible to allocate ids outside in case of parallel updates since the order is established only when the changes are written to the log. The proposer wouldn't know the correct order until it happens.

diptanu: @maxkozlovsky Can we use CAS like semantics to do this? Make the proposer write to the log with the index it read and if it doesn't go through, it can retry N times.

maxkozlovsky: Does not look like there is anything gained by it, it is the same as raft log updates, the proposer writes the log but it does not know the exact log id it will be written at, nor should it care. Allocating ids inside once the order is established seems the most efficient.

PR #598, opened 2024-05-16
handle upload and ingest errors

maxkozlovsky:
Create upload files in temporary directory and move into real location only when upload completed successfully, delete file on failure.


diptanu:
Do we not want this for S3 as well or do they do it automatically?

PR #600, opened 2024-05-16
return status from create_content()

maxkozlovsky:
Return status from create_content() so the caller can determine if content was added or a duplicate was detected. Delete the uploaded content in case of duplicate and don't add new embeddings/metadata.

### DIFF
--- a/protos/coordinator_service.proto
+++ b/protos/coordinator_service.proto
@@ -395,12 +395,17 @@ message ContentMetadata {
     repeated string extraction_graph_names = 14;
 }
 
+enum CreateContentStatus {
+    CREATED = 0;
+    DUPLICATE = 1;
+}
+
 message CreateContentRequest {
     ContentMetadata content = 2;
 }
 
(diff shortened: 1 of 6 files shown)


protos/coordinator_service.proto
```diff
@@ -395,12 +395,17 @@ message ContentMetadata {
...
+    CREATED = 0;
+    DUPLICATE = 1;
+}
+
 message CreateContentRequest {
     ContentMetadata content = 2;
 }
 
 message CreateContentResponse {
-    string id = 1;
```
diptanu, line 403: Why are we removing the ID? If the user doesn't pass an ID, we generate it.

diptanu (changes requested):

protos/coordinator_service.proto
maxkozlovsky: Id is generated on the server. coordinator create_content() does not change id. The callers never used the returned value.

PR #608, opened 2024-05-19
Upgrade openraft

diptanu:
Upgrade openraft


drmingdrmer:
It looks good to me:)

PR #609, opened 2024-05-20
move creation of state changes outside of sm code

maxkozlovsky:


src/scheduler/mod.rs
```diff
@@ -85,36 +45,69 @@ impl Scheduler {
...
+                    Vec::new()
+                };
+                self.shared_state
+                    .mark_change_events_as_processed(vec![state_change], new_state_changes)
+                    .await
             }
-            _ => Vec::new(),
-        };
-        Ok(tasks)
+        } else {
```
diptanu, line 106: Too many branches. Can we just use returns wherever possible and re-write the if-else's ?

diptanu (changes requested):

src/scheduler/mod.rs
maxkozlovsky: updated

PR #640, opened 2024-06-04
operations: add k8s setup examples that work

grampelberg:
This adds an initial example of running everything locally and ephemerally so that you can complete the getting started tutorial on your own k8s environment.


diptanu:
@grampelberg This is great! Thanks for the PR.

PR #647, opened 2024-06-07
remove inactive executors

maxkozlovsky:
Keep track of last heartbeat time for all executors and remove the ones that failed to communicate for a period of time.

This handles the case when executor dies right after registering itself or when the server it is registered with is down.

### DIFF
--- a/src/state/store/state_machine_objects.rs
+++ b/src/state/store/state_machine_objects.rs
@@ -673,12 +673,6 @@ pub struct IndexifyState {
 
     /// Next change id
     pub change_id: std::sync::Mutex<u64>,
-
-    /// Executors registered on this node.
-    pub my_executors: std::sync::Mutex<HashMap<ExecutorId, SystemTime>>,
-
-    /// All executors registered on the cluster.
-    pub all_executors: std::sync::Mutex<HashMap<ExecutorId, SystemTime>>,
 }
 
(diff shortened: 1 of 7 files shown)


src/state/store/state_machine_objects.rs
```diff
@@ -1454,6 +1463,12 @@ impl IndexifyState {
                 };
                 // initialize executor load at 0
                 self.executor_running_task_count.insert(&executor_id, 0);
+
+                self.all_executors
+                    .lock()
+                    .unwrap()
+                    .insert(executor_id.clone(), SystemTime::now());
```
diptanu, line 1470: This is not going to work, pass it from the outside.

src/state/store/state_machine_objects.rs
```diff
@@ -673,6 +673,12 @@ pub struct IndexifyState {
 
     /// Next change id
     pub change_id: std::sync::Mutex<u64>,
+
+    /// Executors registered on this node.
+    pub my_executors: std::sync::Mutex<HashMap<ExecutorId, SystemTime>>,
```
diptanu, line 678: Why not record the node id where the executor is currently heartbeating, and then this becomes a reverse index.

diptanu (changes requested):

PR #678, opened 2024-06-23
operations: adding a helm chart

grampelberg:


diptanu:
@grampelberg Thanks for the PR! There are some merge conflicts on this PR, would you be able to rebase against main and update this please?

grampelberg:
That's what I get for modifying the same file in two branches. @diptanu added in some instructions for using helm. This should be ready to merge now.

I'm wondering if README.md is the right place for this to live, I ended up looking in the wrong place earlier today. Maybe it should be part of the website docs instead.

diptanu:
@grampelberg I agree, it would be good to do this on the docs website.

PR #681, opened 2024-06-24
allow to filter content by graph name and top level content

maxkozlovsky:
Allow filtering content by graph name.

Specifying 'ingestion' for source filter will select top level content.

### DIFF
--- a/protos/coordinator_service.proto
+++ b/protos/coordinator_service.proto
@@ -1,5 +1,4 @@
 syntax = "proto3";
-
 import "google/protobuf/struct.proto";
 
 package indexify_coordinator;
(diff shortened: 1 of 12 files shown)


protos/coordinator_service.proto
```diff
@@ -313,14 +312,26 @@ message GetNamespaceResponse {
     Namespace namespace = 1;
 }
 
+message Empty {
+}
+
+message ContentSource {
+    oneof value {
```
diptanu, line 319: Can't this be enum?

diptanu (changes requested):

protos/coordinator_service.proto
maxkozlovsky: this is translated into raft enum

PR #698, opened 2024-06-26
feature to allow to link two extraction graphs

maxkozlovsky:
Feature to support making an output of an extraction graph an input to another extraction graph. All top level policies (with content source ingestion) of the linked graph are applied to the specified graph/content_source combination.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -16,6 +16,13 @@ use utoipa::{IntoParams, ToSchema};
 use crate::{api_utils, metadata_storage, vectordbs};
 
 #[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+pub struct ExtractionGraphLink {
+    pub source_graph_name: String,
+    pub content_source: String,
+    pub linked_graph_name: String,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
 pub struct ExtractionGraph {
     pub id: String,
     pub name: String,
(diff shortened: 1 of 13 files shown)


src/api.rs
```diff
@@ -15,6 +15,13 @@ use utoipa::{IntoParams, ToSchema};
 
 use crate::{api_utils, metadata_storage, vectordbs};
 
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+pub struct ExtractionGraphLink {
+    pub source_graph_name: String,
+    pub content_source: String,
+    pub linked_graph_name: String,
```
diptanu, line 22: Are there any checks to prevent two graphs from pointing to each other?

diptanu (changes requested):
Left some comments, looking good otherwise.

src/state/store/state_machine_objects.rs
```diff
@@ -698,6 +700,9 @@ pub struct IndexifyState {
 
     /// Next change id
     pub change_id: std::sync::Mutex<u64>,
+
+    /// Graph+ContentSource->Policy Id
+    pub graph_links: Arc<RwLock<HashMap<ExtractionGraphNode, HashSet<String>>>>,
```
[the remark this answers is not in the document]
maxkozlovsky: ok

src/api.rs
maxkozlovsky: ok

PR #738, opened 2024-07-15
fix: header image of getting_started.md

ritwikraha:


ritwikraha:
cc: @diptanu This is just the same thing without the gradient in the header image, it can be merged (IMO)

PR #744, opened 2024-07-15
chore: update filtering of content data

adithyaakrishna:


adithyaakrishna:
WIP: Will add the fix for the tasks one once #743 is merged and then merge this

PR #750, opened 2024-07-16
update: refactor index.md

ritwikraha:
# Tasks

- [x] Refactor home page
- [x] Add Key Concepts
- [x] Add a minimal example inside
- [x] Add backlinks to other docs
- [x] Add diagrams/images
- [ ] What SEO are we trying to rank for?


ritwikraha:
cc: @diptanu what do you think?

also what SEO keywords are we trying to rank for with this page?

PR #752, opened 2024-07-16
chore: fix data display

adithyaakrishna:


adithyaakrishna:
WIP: There is infinite API calls happening due to the page being rendered for every content entry, this PR is a fix for that.

PR #787, opened 2024-07-25
Update getting-started-basic.mdx to use llama_cpp instead of openai

rishiraj:


docs/docs/getting-started-basic.mdx
```diff
@@ -161,16 +171,49 @@ indexify-extractor join-server
...
+      overlap: 100
+  - extractor: "tensorlake/minilm-l6"
+    name: "wikiembedding"
+    content_source: "chunker"
+```
+```yaml graph.yaml for OpenAI
 name: "wiki_extraction_pipeline"
 extraction_policies:
   - extractor: "tensorlake/openai"
+    name: "entity-extractor"
```
diptanu, line 200: You could do Content Tabs on Mintlify and use code blocks inside them. So you can have three blocks - Local Model, OpenAI and Mistral.

docs/docs/getting-started-basic.mdx
```diff
@@ -267,18 +310,72 @@ You can query Indexify to -
...
 
 For now let us create a file named `query.py` with the following content:
 
-```python query.py
+```python query.py for Llama.CPP
+from indexify import IndexifyClient
+from llama_cpp import Llama
+
+client = IndexifyClient()
+llm = Llama.from_pretrained(repo_id='microsoft/Phi-3-mini-4k-instruct-gguf', filename='*q4.gguf', verbose=False)
```
diptanu, line 332: Why not Q8? Wouldn't that work better?

diptanu (changes requested):

docs/docs/getting-started-basic.mdx
rishiraj: Q8 would become 4.4 GB file download (double of current 2.2 GB) for a very minute difference in results. Also Microsoft provides just q4 and fp16 (around 8 GB) gguf files.

docs/docs/getting-started-basic.mdx
rishiraj: done

PR #792, opened 2024-07-25
implement receive a stream of new content updates

maxkozlovsky:
Add api to receive stream of new content updates from the server.

Sample request:

    curl http://localhost:8900/namespaces/default/extraction_graphs/gif/extraction_policies/moondream-description/new_content

    curl http://localhost:8900/namespaces/default/extraction_graphs/gif/extraction_policies/moondream-description/new_content?offset=10


diptanu:
Lets change from web sockets to SSEs.

PR #793, opened 2024-07-26
Update pdf guide

sadath-12:
Correcting few guide mistakes in readme I came across


diptanu:
@sadath-12 Thanks for this!

PR #796, opened 2024-07-28
use prefix for iteration

diptanu:
* Use pre-fetch while iterating (Before - 3.25 seconds while working with a database of 100k rows, after: 18 milliseconds).
* Remove scanning the whole database for getting total count
* Create a reverse index for total tasks, content, etc.

### DIFF
--- a/src/state/store/mod.rs
+++ b/src/state/store/mod.rs
@@ -50,15 +50,7 @@ use requests::{
     V1StateMachineUpdateRequest,
 };
 use rocksdb::{
-    checkpoint::Checkpoint,
-    ColumnFamily,
-    ColumnFamilyDescriptor,
-    DBCommon,
-    Direction,
-    IteratorMode,
-    OptimisticTransactionDB,
-    Options,
-    SingleThreaded,
+    checkpoint::Checkpoint, ColumnFamily, ColumnFamilyDescriptor, DBCommon, Direction, IteratorMode, OptimisticTransactionDB, Options, ReadOptions, SingleThreaded
 };
 use serde::{de::DeserializeOwned, Deserialize};
 use strum::{AsRefStr, IntoEnumIterator};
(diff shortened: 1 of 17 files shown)


src/state/store/state_machine_objects.rs
```diff
@@ -865,7 +846,63 @@ impl IndexifyState {
...
+            .get_cf(StateMachineColumns::TaskAnalytics.cf(db), key.clone())
+            .map_err(|e| StateMachineError::DatabaseError(e.to_string()))?;
+        let mut task_analytics: TaskAnalytics = task_analytics
+            .map(|db_vec| {
+                JsonEncoder::decode(&db_vec)
+                    .map_err(|e| StateMachineError::DatabaseError(e.to_string()))
+            })
+            .unwrap_or_else(|| Ok(TaskAnalytics::default()))?;
+        if prev_task.is_none() {
+            task_analytics.pending();
```
maxkozlovsky, line 882: increments pending twice, None is handled in match as well

src/state/store/mod.rs
```diff
@@ -755,10 +760,44 @@ impl StateMachineStore {
...
+            read_options,
+            mode,
+        );
+
+        let mut extraction_graph_analytics = ExtractionGraphAnalytics::default();
+
+        for kv in itr {
+            let (key, value) = kv.map_err(|e| anyhow::anyhow!("Failed to read db: {}", e))?;
+            let key = String::from_utf8(key.to_vec())?;
+            let extraction_policy_name = key.strip_prefix(&prefix);
```
maxkozlovsky, line 789: the key is extraction_policy_id, however it is used as name here

src/state/store/state_machine_objects.rs
diptanu: Oh that explains why pending count wasn't going down lol. Thanks :)

src/state/store/mod.rs
diptanu: I was looking into it. I don't think we are setting extraction_policy_id anywhere really.

PR #808, opened 2024-08-01
add: video of pipeline

ritwikraha:
Adding the video of the End-to-End pipeline

GIF exports from Arcade are of lower-quality

Any way to add videos directly?


diptanu:
@ritwikraha We can serve the video from clouflare on the docs. It would be faster. Slack me the link to the video

PR #813, opened 2024-08-02
backward compatibility for Task change

maxkozlovsky:
Fix backward compatibility for the change in Task structure. Convert policy id to policy name on startup. Initialize task counters to match existing tasks.


diptanu:
Merging this since I fixed the test in main

PR #819, opened 2024-08-05
allow to specify parameters to upload file in any order

maxkozlovsky:


src/server.rs
```diff
@@ -1184,10 +1158,92 @@ async fn upload_file(
             })?;
         }
     }
-    Err(IndexifyAPIError::new(
-        StatusCode::BAD_REQUEST,
-        "no file provided",
-    ))
+    if let Some(write_result) = write_result {
+        let content_mime = labels.get("mime_type").and_then(|v| v.as_str());
+        let content_mime = content_mime.map(|v| Mime::from_str(v).unwrap());
```
diptanu, line 1163: This could crash the server. We should return an error here to the user.  Unknown mime

diptanu (changes requested):

src/server.rs
maxkozlovsky: ok

PR #830, opened 2024-08-10
Documentation: Minor tweaks and fixes image retrieval example

stevekinney:
Adds syntax highlighting and some fixes to the image retrieval example.

There are probably a few more issues that need to be worked out:

- `client.ingest_remote_file` resulted in a 404. I got around this by using local images as I worked through the example.
- I got back an empty list when running `client.sql_query`. This was true even when running `select * from imageknowledgebase;`.


diptanu:
@stevekinney thanks for the fixes to the docs. I will take a look at the SQL and remote file bugs you reported. Which docs page/example were you following?

PR #835, opened 2024-08-14
add nodeselectors to helm

sadath-12:


jvidalg:
> Are you sure you want the node selectors to be the same for everything? It feels like the extractors should be distinct so that they can land on gpu nodes for example.

Hi @grampelberg @sadath-12  based on the values structure, it seems that each component is already independently customizable. Adding `nodeSelector: {}` to the values file could make this clearer, for example:
```
api:
  enabled: true
  nodeSelector: {}

coordinator:
  enabled: true
  nodeSelector: {}

extractors:
  - name: extractor1
    nodeSelector: {}

  - name: extractor2
    nodeSelector: {}
  ```
  But we would still have an issue if we wanted to deploy multiple extractors that need to be in different node type.

sadath-12:
thanks done

sadath-12:
> > Are you sure you want the node selectors to be the same for everything? It feels like the extractors should be distinct so that they can land on gpu nodes for example.
> 
> Hi @grampelberg @sadath-12 based on the values structure, it seems that each component is already independently customizable. Adding `nodeSelector: {}` to the values file could make this clearer, for example:
> 
> ```
> api:
>   enabled: true
>   nodeSelector: {}
> 
> coordinator:
>   enabled: true
>   nodeSelector: {}
> 
> extractors:
>   - name: extractor1
>     nodeSelector: {}
> 
>   - name: extractor2
>     nodeSelector: {}
> ```

ya but I wanted in values.yaml itself

jvidalg (approved):

grampelberg (approved):

PR #846, opened 2024-08-21
feat: added version display component

adithyaakrishna:


adithyaakrishna:
Currently gets it from the GH Releases, will make the changes to the code and add the banner once I wake up in the morning

adithyaakrishna:
The version mismatch error looks like this

Screenshot 2024-08-23 at 11 04 27 PM

and by default its shown on the footer section of the menu/drawer

Screenshot 2024-08-23 at 11 10 11 PM

diptanu:
@adithyaakrishna "A new version of Indexify is available.... "

adithyaakrishna:
@diptanu Added the banner in such a way that 

- When the user dismisses the announcement, it will disappear and remain hidden for the rest of the day
- If the user opens the app on a new day, the dismissed state will be reset, and the announcement bar will be shown again

PR #852, opened 2024-08-28
add hw/sw information to executor metadata

maxkozlovsky:


crates/indexify_internal_api/src/lib.rs
```diff
@@ -1146,12 +1146,58 @@ impl Default for ContentMetadata {
...
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
 pub struct ExecutorMetadata {
     pub id: String,
-    pub last_seen: u64,
     pub addr: String,
     pub extractors: Vec<ExtractorDescription>,
+    #[serde(default)]
+    pub os_type: String,
+    #[serde(default)]
+    pub os_version: VersionInfo,
```
diptanu, line 1163: This probably needs to be a string or some kind of type which denotes a range.

crates/indexify_internal_api/src/lib.rs
```diff
@@ -1146,12 +1146,58 @@ impl Default for ContentMetadata {
...
+    #[serde(default)]
+    pub python_version: VersionInfo,
+    #[serde(default)]
+    pub memory: u64,
+    #[serde(default)]
+    pub num_cpus: u32,
+    #[serde(default)]
+    pub gpu_memory: u64,
```
diptanu, line 1171: Lets rename this to gpu_memory_mb

crates/indexify_internal_api/src/lib.rs
```diff
@@ -1146,12 +1146,58 @@ impl Default for ContentMetadata {
...
     pub addr: String,
     pub extractors: Vec<ExtractorDescription>,
+    #[serde(default)]
+    pub os_type: String,
+    #[serde(default)]
+    pub os_version: VersionInfo,
+    #[serde(default)]
+    pub python_version: VersionInfo,
+    #[serde(default)]
+    pub memory: u64,
```
diptanu, line 1167: Lets rename this to memory_mb

protos/coordinator_service.proto
```diff
@@ -229,10 +229,21 @@ message Content {
...
 message RegisterExecutorRequest {
     string executor_id = 1;
     string addr = 2;
     repeated Extractor extractors = 3;
+    string os_type = 4;
+    VersionInfo os_version = 5;
+    VersionInfo python_version = 6;
+    uint32 num_cpus = 7;
+    uint64 memory = 8;
+    uint64 gpu_memory = 9;
```
diptanu, line 246: Rename the memory vars to have _mb

src/state/store/compat/mod.rs
```diff
@@ -10,12 +10,15 @@ use crate::state::{NodeId, StateMachineColumns};
 mod convert_v1;
 mod convert_v2;
 mod convert_v3;
+mod convert_v4;
```
diptanu, line 13: Note for later : Lets remove 1, 2 and 3. We need to maintain compatibility with last released version.

diptanu (changes requested):
Left some comments.

crates/indexify_internal_api/src/lib.rs
maxkozlovsky: the version is concrete for executor instance, why would it be a range?

crates/indexify_internal_api/src/lib.rs
maxkozlovsky: the values are in bytes it is simpler to not use specific units

maxkozlovsky: added type

PR #859, opened 2024-09-06
task stream

maxkozlovsky:


server-next/state_store/src/state_machine.rs
```diff
@@ -212,11 +218,16 @@ pub(crate) fn create_tasks(
             "{}_{}_{}",
             task.namespace, task.compute_graph_name, task.ingested_data_id
         );
-        let graph_ctx = txn.get_cf(&IndexifyObjectsColumns::GraphInvocationCtx.cf_db(&db), &key)?;
-        if graph_ctx.is_none() {
-            error!("Graph context not found for task: {}", task.key());
-        }
-        let mut graph_ctx: GraphInvocationCtx = JsonEncoder::decode(&graph_ctx.unwrap())?;
+        let mut graph_ctx =
```
diptanu, line 221: This shouldn't be here. We should create a Graph Context when we invoke compute graph is called. For every invocation we should create the graph invocation context, since that will hold all the accounting data for a specific invocation.

server-next/state_store/src/state_machine.rs
```diff
@@ -235,33 +246,36 @@ pub(crate) fn create_tasks(
...
-        .get_cf(
-            &IndexifyObjectsColumns::ExecutorTaskAssignments.cf(db),
-            &executor_id,
-        )?
-        .unwrap_or_default();
-    let mut task_assignments: HashSet<TaskId> = task_assignments
-        .iter()
-        .map(|task_id| task_id.to_string())
-        .collect();
+    let key = Task::make_executor_key(
```
diptanu, line 259: The name `make_executor_key` is misleading, I am guessing you want to create an assignment key? Let's use the word "allocation" instead of assignment.

server-next/state_store/src/scanner.rs
```diff
@@ -351,6 +351,20 @@ impl StateReader {
         )?;
         Ok(data_objects)
     }
+
+    pub fn get_tasks_by_executor(&self, executor: &ExecutorId, limit: usize) -> Result<Vec<Task>> {
+        let prefix = format!("{}_", executor);
+        let res = self.filter_join_cf(
+            IndexifyObjectsColumns::ExecutorTaskAssignments,
```
diptanu, line 358: lets rename this column to TaskAllocations

server-next/data_model/src/lib.rs
```diff
@@ -244,6 +265,36 @@ impl Task {
             self.namespace, self.compute_graph_name, self.compute_fn_name, self.id
         )
     }
+
+    pub fn make_executor_key(
+        executor_id: &ExecutorId,
```
diptanu, line 270: Pass &self here, and read a bunch of these function from the task struct, and also pass an executor object( guessing we will have the executor object when we are going to create an task allocation? ). This will reduce the number of arguments to this function. Rename the method to allocation_key

server-next/data_model/src/lib.rs
```diff
@@ -244,6 +265,36 @@ impl Task {
...
+        let secs = duration.as_secs() as u128;
+        let nsecs = duration.subsec_nanos() as u128;
+        let nsecs = secs * 1_000_000_000 + nsecs;
+        format!(
+            "{}_{}_{}_{}_{}_{}",
+            executor_id, nsecs, namespace, compute_graph_name, compute_fn_name, task_id,
+        )
+    }
+
+    pub fn key_from_executor_key(executor_key: &[u8]) -> Result<Vec<u8>> {
```
diptanu, line 287: Write a test for this pair of methods

server-next/src/http_objects.rs
```diff
@@ -307,12 +307,39 @@ pub enum TaskOutcome {
...
 pub struct Task {
-    pub id: String,
-    pub status: TaskOutcome,
-    pub created_at: u64,
-    pub updated_at: u64,
+    pub id: TaskId,
+    pub namespace: String,
+    pub compute_fn_name: String,
+    pub compute_graph_name: String,
+    pub ingested_data_id: String,
```
diptanu, line 326: invocation_id

server-next/state_store/src/lib.rs
```diff
@@ -1,21 +1,96 @@
...
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.creation_time
+            .cmp(&other.creation_time)
+            .then_with(|| self.id.cmp(&other.id))
+    }
+}
+
+#[derive(Debug)]
+pub struct ExecutorUnfinishedTasks {
+    pub new_task_channel: broadcast::Sender<()>,
```
diptanu, line 59: how does this fire for the first time when an executor connects to a server?

diptanu (changes requested):

server-next/state_store/src/lib.rs
maxkozlovsky: it reads the list first then waits

server-next/src/http_objects.rs
maxkozlovsky: ok

server-next/data_model/src/lib.rs
maxkozlovsky: it is tested through the task stream test

server-next/data_model/src/lib.rs
maxkozlovsky: there is no task available when this function is called

server-next/state_store/src/scanner.rs
maxkozlovsky: ok

server-next/state_store/src/state_machine.rs
maxkozlovsky: ok

server-next/state_store/src/state_machine.rs
maxkozlovsky: doesn't look like it matters much, this is just some statistics

server-next/data_model/src/lib.rs
diptanu: You are passing a task_id into the function, how come there is no task available?

diptanu (changes requested):

diptanu (changes requested):
Not all comments are addressed

server-next/state_store/src/state_machine.rs
maxkozlovsky: ok

PR #862, opened 2024-09-09
timeout executors

maxkozlovsky:


server-next/data_model/src/lib.rs
```diff
@@ -27,6 +27,10 @@ impl ExecutorId {
     pub fn new(id: String) -> Self {
         Self(id)
     }
+
+    pub fn get(&self) -> &str {
```
diptanu, line 31: Implement the Deref trait *value would return a reference

server-next/src/main.rs
```diff
@@ -22,8 +22,10 @@ struct Cli {
 
 #[tokio::main]
 async fn main() {
+    let env_filter = tracing_subscriber::EnvFilter::try_from_default_env()
+        .unwrap_or_else(|_| tracing_subscriber::EnvFilter::new("info"));
     tracing_subscriber::registry()
-        .with(tracing_subscriber::fmt::layer())
+        .with(tracing_subscriber::fmt::layer().with_filter(env_filter))
```
diptanu, line 28: What does this do?

server-next/src/routes.rs
```diff
@@ -505,27 +506,22 @@ async fn notify_on_change(
 }
 
 async fn executor_tasks(
-    Path(executor_id): Path<String>,
+    Path(_executor_id): Path<String>,
```
diptanu, line 509: Remove this from the path if this is not being used.

server-next/src/routes.rs
```diff
@@ -505,27 +506,22 @@ async fn notify_on_change(
...
         .await
         .map_err(|e| IndexifyAPIError::internal_error(e))?;
-    let stream = state_store::task_stream(
-        state.indexify_state,
-        data_model::ExecutorId::new(executor_id.clone()),
-        10,
-    );
-    let executor_id_clone = executor_id.clone();
-    let executor_manager_clone = state.executor_manager.clone();
+    let stream = state_store::task_stream(state.indexify_state, payload.id.clone(), 10);
```
diptanu, line 523: What does 10 do here? Limit upto sending 10 tasks? Extract this into a constant .

server-next/state_store/src/lib.rs
```diff
@@ -129,13 +135,40 @@ impl IndexifyState {
...
             db: Arc::new(db),
             state_change_tx: tx,
             state_change_rx: rx,
             last_state_change_id: Arc::new(AtomicU64::new(0)),
-            unfinished_tasks_by_executor: RwLock::new(HashMap::new()),
-        })
+            executor_states: RwLock::new(HashMap::new()),
+        });
+
+        let executors = s.reader().get_all_executors()?;
```
diptanu, line 146: Move all this into the executor manager and leave the state store to manage only the state, and not have logic related to removing and adding executors.

server-next/state_store/src/lib.rs
```diff
@@ -204,15 +237,33 @@ impl IndexifyState {
                 vec![]
             }
             requests::RequestPayload::RegisterExecutor(request) => {
+                {
+                    let mut states = self.executor_states.write().unwrap();
+                    let entry = states.entry(request.executor.id.clone()).or_default();
+                    entry.num_registered += 1;
```
diptanu, line 243: The reference count based approach is pretty convoluted for no additional benefits. We are incrementing the value when we get a register request and when we deregister, we decrement the ref count by 1 and remove it only when ref count becomes zero. 
Simpler logic would be to start timers to remove all the executors at server startup in the ExecutorManager, and when we get registrations just cancel the timers. This wouldn't involve any additional logic spread across many different state machine transitions.

diptanu (changes requested):

server-next/src/main.rs
maxkozlovsky: allows to modify logging level from environment variable

server-next/state_store/src/lib.rs
maxkozlovsky: Reference count is necessary since deregister is not guaranteed to be called before new connection is opened

server-next/state_store/src/lib.rs
maxkozlovsky: I am not sure how to accomplish that. There is no pointer to executor manager here.

server-next/src/routes.rs
maxkozlovsky: ok

server-next/src/routes.rs
maxkozlovsky: ok

server-next/state_store/src/lib.rs
diptanu: @maxkozlovsky There is a de-register method in the ExecutorManager object. So when a new connection is opened, you would just cancel the timer for the deregister call to state store there.

PR #863, opened 2024-09-10
add test for schedule executor delete

maxkozlovsky:


server-next/src/executors.rs
```diff
@@ -43,3 +41,89 @@ impl ExecutorManager {
...
+        let executors = indexify_state.reader().get_all_executors()?;
+
+        assert_eq!(executors.len(), 1);
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn test_deregister_executor() -> Result<()> {
+        let temp_dir = tempfile::tempdir()?;
```
diptanu, line 87: Create a new method in TestStateStore for these setups

diptanu (changes requested):

server-next/src/executors.rs
maxkozlovsky: does not work since ExecutorManager is not in scope

PR #864, opened 2024-09-11
add tests for executor delete

maxkozlovsky:
change separator value
remove unnecessary value setting in task allocations

### DIFF
--- a/server-next/src/scheduler.rs
+++ b/server-next/src/scheduler.rs
@@ -256,10 +256,16 @@ impl Scheduler {
 
 #[cfg(test)]
 mod tests {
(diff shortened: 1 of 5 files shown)


server-next/src/scheduler.rs
```diff
@@ -336,4 +342,109 @@ mod tests {
         assert_eq!(unprocessed_state_changes.len(), 1);
         Ok(())
     }
+
+    pub async fn schedule_all(indexify_state: &IndexifyState, scheduler: &Scheduler) -> Result<()> {
```
diptanu, line 346: @maxkozlovsky Where is this being called from besides tests?

diptanu (changes requested):

PR #877, opened 2024-09-25
Add `stdout` and `stderr` to server ingest

stangirala:
Capture stdout and stderr from the executor and save in the blob store,
- Add `LoggingProcessPoolExecutor` to redirect streams.
- Add `FunctionWorkerOutput` to capture stream data.
- Add multi-part upload for these streams in `task_reporter.py`
- Add support for `ingest.rs` for new data and write to blob store.
- Add test.


server-next/src/routes/internal_ingest.rs
```diff
@@ -127,10 +130,15 @@ pub async fn ingest_files_from_executor(
                     task_result.compute_graph,
                     task_result.compute_fn,
                     task_result.invocation_id,
-                    "exception_msg"
+                    name_ref,
```
diptanu, line 133: &name

server-next/src/routes/internal_ingest.rs
```diff
@@ -117,7 +120,7 @@ pub async fn ingest_files_from_executor(
                 let res = write_to_disk(state.clone().blob_storage, &mut field, &file_name).await?;
                 node_output_sequence += 1;
                 output_objects.push(res.clone());
-            } else if name == "exception_msg" {
+            } else if name == "exception_msg" || name == "stdout" || name == "stderr" {
```
diptanu, line 123: Is there a in syntax ?

server-next/src/routes/internal_ingest.rs
```diff
@@ -127,10 +130,15 @@ pub async fn ingest_files_from_executor(
...
-                    "exception_msg"
+                    name_ref,
                 );
                 let res = write_to_disk(state.clone().blob_storage, &mut field, &file_name).await?;
-                exception_msg = Some(res);
+                match name_ref.as_str() {
+                    "exception_msg" => { exception_msg = Some(res) },
+                    "stdout" => { stdout_msg = Some(res) },
+                    "stderr" => { stderr_msg = Some(res) },
+                    _ => { panic! ("should not match here") }
```
diptanu, line 140: panic, just do error!

server-next/src/routes/internal_ingest.rs
```diff
@@ -165,20 +173,35 @@ pub async fn ingest_files_from_executor(
         node_outputs.push(node_output);
     }
 
-    let task_diagnostic = exception_msg.map(|exception_msg| {
-        let exception_payload = DataPayload {
+    let exception_payload = exception_msg.map(|exception_msg| {
```
diptanu, line 176: Extract a function for these, and then inline where you are using `exception_payload`, etc.

python-sdk/indexify/executor/function_worker.py
```diff
@@ -56,10 +89,13 @@ async def async_submit(
...
             self._executor.shutdown(wait=True, cancel_futures=True)
             raise mp
-        except Exception as e:
-            raise Exception(traceback.format_exc())
 
-        return resp
+        return FunctionWorkerOutput(
+            indexify_data=result,
+            exception=exception,
+            stdout=stdout,
```
diptanu, line 96: Should we upload these only when get an exception? Otherwise you probably don't need them. In the future we could pipe in a `force_log` through the invoke method to allow users selectively log even if there are no exceptions.

python-sdk/indexify/executor/agent.py
```diff
@@ -117,15 +118,18 @@ async def task_completion_reporter(self):
                         border_style="info",
                     )
                 )
-                task: Task = self._task_store.get_task(task_outcome.task.id)
+
                 try:
+                    print(f'task_outcome: {task_outcome}')
```
diptanu, line 123: Remove

python-sdk/indexify/executor/agent.py
```diff
@@ -279,6 +305,7 @@ async def task_launcher(self):
                             )
                             + Text(f"Exception: {e}", style="red")
                         )
+                        print('----here {}', type(async_task.task))
```
diptanu, line 308: Remove

diptanu (changes requested):

python-sdk/indexify/executor/function_worker.py
stangirala: I'll just add this as a cli flag next. Would prefer to let the client/executor control which logs are being written.

PR #878, opened 2024-09-25
Fixed the reduction logic and a bunch of other bugs

diptanu:
Fixed some logic related to accounting of outstanding tasks and sending tasks to executors which are not previously sent


server-next/state_store/src/lib.rs
```diff
@@ -53,17 +50,23 @@ impl ExecutorState {
...
+    pub fn added(&mut self, task_ids: &Vec<TaskId>) {
+        self.task_ids_sent.extend(task_ids.clone());
+    }
+
     // Send notification for remove because new task can be available if
     // were at maximum number of tasks before task completion.
-    pub fn removed(&mut self) {
+    pub fn removed(&mut self, task_id: TaskId) {
         let _ = self.new_task_channel.send(());
+        self.task_ids_sent.remove(&task_id);
```
maxkozlovsky, line 69: should update sent ids first then notify

server-next/task_scheduler/src/task_creator.rs
```diff
@@ -159,29 +159,44 @@ pub async fn handle_task_finished(
...
-                compute_graph: task.compute_graph_name.clone(),
-                invocation_id: task.invocation_id.clone(),
-                tasks: vec![],
-                new_reduction_tasks: vec![],
-                processed_reduction_tasks: vec![],
-                invocation_finished: true,
-            });
-        }
+    if edges.is_none() {
+        let invocation_finished = if invocation_ctx.outstanding_tasks == 0 {
```
maxkozlovsky, line 163: this may not work as intended, if multiple task finished are processed when outstanding_tasks == 0, multiple notifications can be generated. Does not account for any pending events that may result in new tasks being generated. My pr has correct implementation.

PR #884, opened 2024-09-29
feat: trigger website build on release

adithyaakrishna:


adithyaakrishna:
Instead of using cURL for creating, I'll update it to use vercel cli

.github/workflows/trigger-vercel-build.yml
```diff
@@ -0,0 +1,32 @@
...
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  trigger-vercel-build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout indexify-website
```
diptanu, line 11: This is not needed since you are just using curl

diptanu (changes requested):

PR #885, opened 2024-09-30
Fix router validations

diptanu:
Handles the following router signatures 

```
List[Union[name1, name2]]

Union[name1, name2]

Optional[name1]

Optional[Union[name1, name2]
```


stangirala:
+1

PR #902, opened 2024-10-02
feat: ability to set namespace

sadath-12:


diptanu:
Can’t you create a new client if you want to use a different namespace?

sadath-12:
but you need a client to create namespace so procedure would be 
- create client 
- create namespace
- create new client 

which doesnt seem to look gud

diptanu:
@sadath-12 Good point, I would suggest to change this slightly. Lets do something like this - 

```python
client = Client.new_namespace("foo", server_addr="....")
```

That creates a new namespace and returns the client. So you are able to create a ns and get the client at the same time.
```
@classmethod
def new_namespace(cls, namespace: str, server_addr: Optional[str] = "http://localhost:8900"):
   pass
 ```

sadath-12:
thanks @diptanu , please do check it

PR #907, opened 2024-10-02
feat: add docker for pdf

sadath-12:


diptanu:
Not needed, we have created images to deploy the example in containers. We need an image which has only py-inkwell and dependencies in the inkwell repo. May be you can move this Dockerfile there?

PR #908, opened 2024-10-02
Setup new interface without having to create a Client

stangirala:
Setup the graph without needing a client. Client is not aggregated into the Graph object with client and graphds (the old graph). Here the user can create a Graph as usual (see readme example) but all they have to do is run(). This change remove the few extra steps towards creating a client. Also makes it easy to run inprocess (ie local) or from a remote reference already deployed on the server.

See readme example that has commented out code for 3 scenarios - in process, local server, local server from Graph reference.


diptanu:
Closing this since this is done in main

PR #922, opened 2024-10-06
Add Contextual RAG Example

stangirala:
Add locally running example.


examples/contextual_rag/workflow.py
```diff
@@ -0,0 +1,288 @@
...
+            {"role": "system", "content": "You are an AI assistant tasked with analyzing documents and answering questions about them."},
+            {"role": "user", "content": payload}
+        ]
+    )
+
+    return response.choices[0].message.content, response.usage
+
+
+if __name__ == '__main__':
+    # g: Graph = Graph("test", start_node=generate_chunk_contexts)
```
diptanu, line 188: Why are these commented? How does someone run the example?

examples/contextual_rag/workflow.py
```diff
@@ -0,0 +1,288 @@
...
+    console.print("Asking Question -- " + f"`[bold red]{question}[/]`\n")
+    question_embd = model.encode(question)
+
+    console.print('----')
+
+    console.print("\nRAG Output", style="bold red")
+
+    chunks = l_client.open_table("chunk-embeddings").search(question_embd).limit(5).to_list()
+    d = []
+    for ii, i in enumerate(chunks):
```
diptanu, line 220: for chunk_id, i

examples/contextual_rag/workflow.py
```diff
@@ -0,0 +1,288 @@
...
+        d.append('chunk_id: ' + str(ii))
+        d.append('chunk: ' + i['chunk'])
+    p = '\n'.join(d)
+
+    regular_prompt = f"""
+    You are presented with information from a recent SEC10K document. The document has been chunked, and we are providing the chunks and the chunk id. The information is presented in the form:
```
diptanu, line 226: Since we are showing a specific example here, might be worth setting up the example with the right document. You could dump the document in a text file here

stangirala: updated the example to GET a paste file that has the 10k document.

Issue #910, opened 2024-10-03
Decouple the TS Client from UI Completely 


adithyaakrishna:
**Description:**

The current UI uses TS Client for a few APIs and for the types. This issue is to make the UI completely independent of the TS Client and rely upon axios calls directly from the UI

PR #923, opened 2024-10-07
feat: decouple tsclient from the ui

adithyaakrishna:
**Description:**

- This PR aims to decouple the TS Client completely and make calls to the indexify server directly from the UI
- Moved types to UI


adithyaakrishna:
This Partially resolves #910

PR #939, opened 2024-10-10
fix: s3 and helm dep

sadath-12:
The pr enables users to set their own configurations via k8s . 

I have focused on enabling to set s3 bucket and region via this and I was able to see the outputs store in s3 which was earlier only storing in disk even tho user sets the s3 config

For testing: I used ```local.yaml``` for installing k8s except the change that I kept 

```yaml 
minio:
    enabled: false
```

and blobstore

```yaml
blobStore:
  allowHTTP: true
  credentialSecret: blob-creds
  config:
    backend: s3
    s3:
      bucket: indexify
      region: us-east-1
      accessKey: 
      secretKey: 
```

### DIFF
--- a/operations/k8s/helm/templates/config.yaml
+++ b/operations/k8s/helm/templates/config.yaml
@@ -1,0 +1,13 @@
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: indexify
+  labels:
+    {{- include "labels" (dict "name" "indexify" "component" "config" "global" $) | nindent 4 }}
+data:
+  config.yaml: |-
+    listen_addr: 0.0.0.0:8900
+    state_store_path: /tmp/indexify/state
+    blob_storage:
+      {{- .Values.blobStore.config | toYaml | nindent 6 }}
--- a/operations/k8s/helm/templates/config.yaml
+++ b/operations/k8s/helm/templates/config.yaml
@@ -2,7 +2,7 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: indexify
+  name: indexify-server
   labels:
     {{- include "labels" (dict "name" "indexify" "component" "config" "global" $) | nindent 4 }}
 data:
--- a/operations/k8s/helm/templates/server.yaml
+++ b/operations/k8s/helm/templates/server.yaml
@@ -23,7 +23,6 @@ spec:
     metadata:
       labels:
         {{- include "labels" (dict "name" "server" "component" "server" "global" $) | nindent 8 }}
-
     spec:
       {{- if .nodeSelector }}
       nodeSelector:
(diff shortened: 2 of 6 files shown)


operations/k8s/helm/templates/config.yaml
```diff
@@ -0,0 +1,13 @@
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: indexify
```
miguelhrocha, line 5: would you mind renaming this to `indexify-server`? We want to keep the door open to config-maps for executors, which would be more explicit.

operations/k8s/helm/templates/server.yaml
```diff
@@ -32,15 +31,23 @@ spec:
...
-            - mountPath: /tmp/indexify-blob-storage
-              name: data
-          {{- end}}
+            - name: config
+              mountPath: /indexify/config
+              readOnly: true
+            {{- if not $.Values.blobStore.config.s3 }}
+            - name: data
+              mountPath: /tmp/indexify-blob-storage
+            {{- end }}
```
miguelhrocha, line 45: I believe these volumes are not interchangeable. The server needs a writable path where to write the state, which is not the same as the blobs persisted by minio, or the like.

I would request to mount this unconditionally and set the path to match the one in the config maps state_store_path

miguelhrocha (changes requested):
Thank you for the PR @sadath-12! Would you mind adding a short description on the context of this change, and the steps to reproduce the expected deployment?

sadath-12:
thanks @miguelhrocha . Please take a look on changes

operations/k8s/helm/templates/server.yaml
```diff
@@ -51,7 +54,12 @@ spec:
               path: /
               port: 8900
 
-  {{- with .persistance }}
+      volumes:
+        - name: config
+          configMap:
+            name: indexify-server
+
+  {{- with .persistence }}
```
miguelhrocha, line 62: This line is fixing a type 😄, would you mind renaming the attribute in `local.yaml` and `values.yaml` for `persistence` too? Otherwise this deployment doesn't work

miguelhrocha (changes requested):

operations/k8s/helm/templates/server.yaml
sadath-12: wow ..

miguelhrocha:
@sadath-12 would you mind changing the `persistance` for `persistence` in the `values.yaml` too?

diptanu:
@sadath-12 Looks good to me, can you please add some steps in the description which you used to test the changes.

miguelhrocha (approved):
Thank you for this!

Issue #935, opened 2024-10-09
Make catching optional


sadath-12:
Initially just to be safe some users who run indexify centrally for all users might prefer to disable catching .
so there is no  chance for data leak. 

Also in lot of cases the cache actually is misleading where I used to get errors for the new code , I would have to delete the cache and retry to get changes into effect and let them work

diptanu:
There is no caching happening at the moment when graphs are deployed in Indexify server. We cache results only when Graphs are run in-process on laptops by users to speed up testing.

PR #942, opened 2024-10-10
feat: make cache optional

sadath-12:
addressing #935

### DIFF
--- a/python-sdk/indexify/functions_sdk/graph.py
+++ b/python-sdk/indexify/functions_sdk/graph.py
@@ -61,7 +61,7 @@ def is_pydantic_model_from_annotation(type_annotation):
 
 class Graph:
     def __init__(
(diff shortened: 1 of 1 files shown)


python-sdk/indexify/functions_sdk/graph.py
```diff
@@ -61,14 +61,15 @@ def is_pydantic_model_from_annotation(type_annotation):
 
 class Graph:
     def __init__(
-        self, name: str, start_node: IndexifyFunction, description: Optional[str] = None
+        self, name: str, start_node: IndexifyFunction, description: Optional[str] = None,enable_cache: Optional[bool] = True
```
diptanu, line 64: This shouldn't be here. Add this in graph.run() api. So you can toggle caching selectively for local runs

diptanu (changes requested):

PR #944, opened 2024-10-11
fix: creating namespaces with graph

### LINKED ISSUE #928: Consider not to function without namespaces
Currently I see we could actually create compute graphs .. etc while we don't have any namespaces .

Consider an hard limit on namespaces because some of us enable access controls via namespaces

sadath-12:
Users can tell indexify to create namespace when they deploy graph , if its already available it will use that namespace .

Also this fixes indexify not having default namespace when its already using it

closes #928

### DIFF
--- a/python-sdk/indexify/http_client.py
+++ b/python-sdk/indexify/http_client.py
@@ -205,6 +205,10 @@ class IndexifyClient:
         # Create a new client instance with the specified server address
         client = cls(service_url=server_addr)
 
+        if namespace in cls.namespaces:
+            client.namespace = namespace
+            return client
+
         try:
             # Create the new namespace using the client
             client.create_namespace(namespace)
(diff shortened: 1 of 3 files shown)


diptanu:
@sadath-12 We should create the default namespace when the server starts up. Secondly, if a namespace doesn't exist I believe we shouldn't create it at the time of creating a new graph. We will have security controls around who can create a namespace in the server.

sadath-12:
@diptanu this might be important as we can't pass namespace while creating a graph

vcknorket:
Is this something you can review @diptanu ? In our case, we are allowing flexibility with namespace and schema and believe might be something many of folks would need instead of putting all under "default' schema

seriousben:
Just for future reference, we introduced a way to use namespaces easily in https://github.com/tensorlakeai/indexify/pull/978 which should be available on the latest indexify sdk version.

It enables something like this:
```
client = IndexifyClient(namespace="object_detection")
client.create_namespace("object_detection")
g = RemoteGraph.deploy(g, client=client)
```

PR #974, opened 2024-10-23
Add image builder for executor

stangirala:
[x] Run `make fmt`.


python-sdk/indexify/executor/agent.py
```diff
@@ -369,3 +449,36 @@ async def _shutdown(self, loop):
     def shutdown(self, loop):
         self._function_worker.shutdown()
         loop.create_task(self._shutdown(loop))
+
+
+
+async def get_compute_graph_from_server(url):
```
diptanu, line 455: Use http_client.py it has the methods to get compute graph

python-sdk/indexify/executor/agent.py
```diff
@@ -369,3 +449,36 @@ async def _shutdown(self, loop):
...
+        async with httpx.AsyncClient() as client:
+            resp = await client.get(url=url, headers={"Content-Type": "application/json"})
+            return resp.json()
+    except Exception as e:
+        console.print(
+            Text("Unable to get graph information from Server", style="red bold")
+        )
+
+
+async def get_run_strs_from_image(task: Task, protocol, server_addr):
```
diptanu, line 466:
```
async def get_image_info_for_compute_graph(task: Task) -> ImageInfo:
  pass
```

python-sdk/indexify/executor/agent.py
```diff
@@ -144,15 +149,90 @@ async def task_completion_reporter(self):
...
         async_tasks: List[asyncio.Task] = []
         fn_queue: List[FunctionInput] = []
+
         async_tasks.append(
             asyncio.create_task(
                 self._task_store.get_runnable_tasks(), name="get_runnable_tasks"
             )
         )
+
+        run_first = True
```
diptanu, line 159: We don't need this if we figure out if the image needs bootstrapping.

python-sdk/indexify/executor/agent.py
```diff
@@ -144,15 +149,90 @@ async def task_completion_reporter(self):
...
+
         while True:
             fn: FunctionInput
             for fn in fn_queue:
                 task: Task = self._task_store.get_task(fn.task_id)
+
+                # Bootstrap this executor.
+                if run_first:
```
diptanu, line 167:
```
if self._require_image_bootstrap:
    image_info = get_image_info(task)
    executor_image_builder(image_info)
    self._require_image_bootstrap = False
```

python-sdk/indexify/executor/agent.py
```diff
@@ -144,15 +149,90 @@ async def task_completion_reporter(self):
...
+                                    f"Exception bootstrapping the Executor",
+                                    style="red bold",
+                                )
+                                + Text(f"Exception: {e}", style="red")
+                                + Text(
+                                    f"Failing the executor launch",
+                                    style="red bold",
+                                )
+                            )
+                            exit(-1)
```
stangirala, line 232: Don't append a runnable task if this happens, just write a completed task. Add the exception.

python-sdk/indexify/executor/agent.py
```diff
@@ -308,16 +354,10 @@ def to_sentence_case(snake_str):
...
-                # install dependencies
-                # rewrite the image name
-                pass
-
             data = ExecutorMetadata(
                 id=self._executor_id,
                 addr="",
-                image_name=runtime_probe.image_name,
+                image_name=self.name_alias,
                 labels=runtime_probe.labels,
```
stangirala, line 361: should we add the alias here?

diptanu: Labels doesn't have the image name so its fine.

python-sdk/indexify/executor/image_dependency_installer.py
```diff
@@ -0,0 +1,51 @@
...
+    }
+)
+
+console = Console(theme=custom_theme)
+
+
+def _record_image_name(name: str):
+    file_path = os.path.expanduser("~/.indexify/image_name")
+
+    if os.path.exists(file_path):
```
diptanu, line 26: Remove this check since it will always be False when it's a base image.

python-sdk/indexify/executor/image_dependency_installer.py
```diff
@@ -0,0 +1,51 @@
...
+        "error": "red",
+        "success": "green",
+    }
+)
+
+console = Console(theme=custom_theme)
+
+
+def _record_image_name(name: str):
+    file_path = os.path.expanduser("~/.indexify/image_name")
```
diptanu, line 24: makedirs

diptanu (changes requested):
Looks great.

python-sdk/indexify/executor/agent.py
```diff
@@ -57,7 +62,24 @@ def __init__(
...
+
+        self._probe = RuntimeProbes()
+
+        runtime_probe: ProbeInfo = self._probe.probe()
+        self._require_image_bootstrap = (
+            True
+            if (runtime_probe.is_default_executor and self.name_alias is not None)
+            else False
+        )
+        self._executor_failed = False
```
diptanu, line 77: Rename this to - self._executor_bootstrap_failed

python-sdk/indexify/executor/agent.py
```diff
@@ -144,15 +165,55 @@ async def task_completion_reporter(self):
...
+                # Bootstrap this executor. Fail the task if we can't.
+                if self._require_image_bootstrap:
+                    try:
+                        image_info = await _get_image_info_for_compute_graph(
+                            task, self._protocol, self._server_addr
+                        )
+                        image_dependency_installer.executor_image_builder(
+                            image_info, self.name_alias
+                        )
+                        self._require_image_bootstrap = False
```
diptanu, line 199: This should be set to true even if this fails no? so something like Finally: ....

python-sdk/tests/image_builder_docker_e2e/test_graph_image_builder_broken.py
```diff
@@ -0,0 +1,66 @@
+import unittest
+
+from indexify.functions_sdk.image import Image
+from pydantic import BaseModel
+from indexify import indexify_function, Graph, RemoteGraph
+from typing import List
+
+
+"""
```
diptanu, line 9: These tests are great. May be rename the folder and the test files to say - image_bootstraper ... We will likely have to build something which actually builds the container images later :)

diptanu (approved):
Left some small comments, looks good otherwise!

python-sdk/indexify/executor/agent.py
stangirala: If it fails we just use the `self._executor_failed` flag to fail all the subsequent tasks. So we don't have to set it if the bootstrap fails.

PR #976, opened 2024-10-24
feature(api): enforcing namespace existence on all namespaced routes

seriousben:
Fixes https://github.com/tensorlakeai/indexify/issues/928

This PR adds middleware to validate the existence of namespaces for all external API calls under `/namespaces/:namespace/...`. It also automatically creates the default namespace required by the Python SDK on server initialization.

### Initialization
image

### API Errors

Note: Server initialization to cause this.

image

### SDK Error

image

## Implementation

Adding a middleware to all namespaces routes required nesting them in order to ensure the middleware was only evaluated when needed.

The default namespace creation checks for the existing of the namespace in order to not overwrite an existing default namespace. Overwriting would overwrite the creation time of the namespace.

## PR Checks

Note: I also deleted all storage and caches.

[x] Run `make fmt`.
```
Ran 7 tests in 1.667s

OK
```

### DIFF
--- a/server/src/routes.rs
+++ b/server/src/routes.rs
@@ -3,8 +3,18 @@ use std::{sync::Arc, time::Duration};
 use anyhow::Result;
 use axum::{
     body::Body,
-    extract::{DefaultBodyLimit, MatchedPath, Multipart, Path, Query, Request, State},
+    extract::{
+        DefaultBodyLimit,
+        MatchedPath,
+        Multipart,
+        Path,
+        Query,
+        RawPathParams,
+        Request,
+        State,
+    },
     http::{Method, Response},
+    middleware::{self, Next},
     response::{sse::Event, IntoResponse},
     routing::{delete, get, post},
     Json,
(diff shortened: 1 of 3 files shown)


server/src/routes.rs
```diff
@@ -276,6 +232,98 @@ async fn ui_handler(Path(url): Path<String>) -> impl IntoResponse {
...
+
+    // if the namespace path variable is found, check if the namespace exists
+    if let Some((_, namespace)) = namespace_param {
+        let reader = route_state.indexify_state.reader();
+        let ns = reader
+            .get_namespace(namespace)
+            .map_err(IndexifyAPIError::internal_error)?;
+
+        if ns.is_none() {
+            return Err(IndexifyAPIError::not_found("Namespace not found"));
```
seriousben, line 320: Looking at the screenshots, I think we could improve the DX a lot by mentioning the namespace name.

I'll will change this.

diptanu (approved):
This looks great, let's merge this after you make the change to include the namespace name in the response.

seriousben:
Testing with new error message:

image

PR #985, opened 2024-10-29
Add Image Version to SDK and Server

stangirala:
Add support for Image Version. Change SDK API to accept the new param on the executor. Change default value for probe. Setup ExecutorMetadata with the new param. Change Server API and Data Model for the new param. Add image hash and version bump logic to state store. Add unit test.

[x] Run `make fmt`.

Tested with then new test file using 3 aliased executors.

### DIFF
--- a/python-sdk/indexify/cli.py
+++ b/python-sdk/indexify/cli.py
@@ -170,6 +170,9 @@ def executor(
     name_alias: Optional[str] = typer.Option(
         None, help="Name alias for the executor if it's spun up with the base image"
     ),
+    image_version: Optional[int] = typer.Option(
+        "1", help="Requested Image Version for this executor"
+    ),
 ):
     id = nanoid.generate()
     console.print(
@@ -179,7 +182,8 @@ def executor(
             f"Server address: {server_addr}\n"
             f"Executor ID: {id}\n"
             f"Executor cache: {executor_cache}\n"
-            f"Name Alias: {name_alias}",
+            f"Name Alias: {name_alias}"
+            f"Image Version: {image_version}\n",
             title="Agent Configuration",
             border_style="info",
         )
@@ -201,6 +205,7 @@ def executor(
         config_path=config_path,
         code_path=executor_cache,
         name_alias=name_alias,
+        image_version=image_version,
     )
 
     try:
--- a/python-sdk/indexify/executor/image_dependency_installer.py
+++ b/python-sdk/indexify/executor/image_dependency_installer.py
@@ -28,6 +28,7 @@ def _record_image_name(name: str):
     with open(file_path, "w") as file:
         file.write(name)
 
+
 def _record_image_version(version: int):
     dir_path = os.path.expanduser("~/.indexify/")
     file_path = os.path.expanduser("~/.indexify/image_version")
@@ -44,7 +45,9 @@ def _install_dependencies(run_str: str):
         raise Exception(f"Unable to install dep `{run_str}`")
 
 
-def executor_image_builder(image_info: ImageInformation, name_alias: str, image_version: int):
+def executor_image_builder(
+    image_info: ImageInformation, name_alias: str, image_version: int
+):
     console.print(Text("Attempting Executor Bootstrap.", style="red bold"))
 
     run_strs = image_info.run_strs
@@ -56,6 +59,11 @@ def executor_image_builder(image_info: ImageInformation, name_alias: str, image_
 
     console.print(Text("Install dependencies done.", style="red bold"))
 
-    console.print(Text(f"Recording image name {name_alias} and version {image_version}", style="red bold"))
+    console.print(
+        Text(
+            f"Recording image name {name_alias} and version {image_version}",
+            style="red bold",
+        )
+    )
     _record_image_name(name_alias)
     _record_image_version(image_version)
--- a/python-sdk/indexify/executor/image_dependency_installer.py
+++ b/python-sdk/indexify/executor/image_dependency_installer.py
@@ -20,20 +20,16 @@ custom_theme = Theme(
 console = Console(theme=custom_theme)
 
 
-def _record_image_name(name: str):
+def _record_image_name(name: str, version: int):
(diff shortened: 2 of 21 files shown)


python-sdk/indexify/cli.py
```diff
@@ -170,6 +170,9 @@ def executor(
     name_alias: Optional[str] = typer.Option(
         None, help="Name alias for the executor if it's spun up with the base image"
     ),
+    image_version: Optional[int] = typer.Option(
```
diptanu, line 173: Is this so that we can alias the version as well? Can we make the name_alias have the version as name:version.

python-sdk/indexify/executor/image_dependency_installer.py
```diff
@@ -48,5 +59,11 @@ def executor_image_builder(image_info: ImageInformation, name_alias: str):
...
 
-    console.print(Text(f"Recording image name {name_alias}", style="red bold"))
+    console.print(
+        Text(
+            f"Recording image name {name_alias} and version {image_version}",
+            style="red bold",
+        )
+    )
     _record_image_name(name_alias)
+    _record_image_version(image_version)
```
diptanu, line 69: Merge this and the image name together

server/data_model/src/lib.rs
```diff
@@ -84,6 +85,27 @@ pub struct ImageInformation {
     pub tag: String,
     pub base_image: String,
     pub run_strs: Vec<String>,
+    pub image_hash: String,
+    pub version: ImageVersion, // this maps with image_hash
```
diptanu, line 89: // This gets updated when hash changes

server/data_model/src/lib.rs
```diff
@@ -84,6 +85,27 @@ pub struct ImageInformation {
...
+}
+
+impl ImageInformation {
+    pub fn new(image_name: String, tag: String, base_image: String, run_strs: Vec<String>) -> Self {
+        let mut image_hasher = Sha256::new();
+        image_hasher.update(image_name.clone());
+        image_hasher.update(tag.clone());
+        image_hasher.update(base_image.clone());
+        image_hasher.update(run_strs.clone().join("-----"));
```
diptanu, line 98: Why —— and not just add the strings?

server/data_model/src/lib.rs
```diff
@@ -111,7 +133,8 @@ pub struct ComputeFn {
 
 impl ComputeFn {
     pub fn matches_executor(&self, executor: &ExecutorMetadata) -> bool {
-        self.placement_constraints.matches(&executor.labels)
+        (self.image_information.version.0 == executor.image_version) &&
```
diptanu, line 136: Can we merge the image name and version in the executor labels? That we we don’t have to have more predicates besides matching the labels

server/data_model/src/lib.rs
```diff
@@ -222,6 +270,21 @@ impl Default for GraphVersion {
...
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, PartialOrd, Ord, Eq, Copy)]
+pub struct ImageVersion(pub u32);
+
+impl ImageVersion {
+    pub fn next(&self) -> Self {
+        Self(self.0 + 1)
+    }
+}
+
+impl Default for ImageVersion {
```
diptanu, line 282: You have a constant defined at the top, reuse that here

diptanu (changes requested):
Looks good overall, did a partial review and left some minor comments.

python-sdk/indexify/cli.py
stangirala: yea that's right, I figured it might be simpler to have them as two separate strings that the executor can use to as needed. Having `<name:version>` also means we need to fill in `latest` or `1` as a zero if the user doesn't provide it. 

I just prefer keeping it explicit.

server/data_model/src/lib.rs
stangirala: either works, I just had a preference to include a non-empty separator.

server/data_model/src/lib.rs
stangirala: Would prefer to not add too many things to label since it would be untyped.

server/data_model/src/lib.rs
diptanu: In general, it doesn't make sense to hash anything non unique across all objects.

server/data_model/src/lib.rs
stangirala: which constant?

server/state_store/src/state_machine.rs
```diff
@@ -390,6 +390,19 @@ pub(crate) fn create_compute_graph(
...
             compute_graph.version = existing_compute_graph.version.next();
         }
+
+        for (node_name, node) in compute_graph.nodes.iter_mut() {
+            let existing_node = existing_compute_graph.nodes.get(node_name).unwrap();
+
+            let existing_hash = existing_compute_graph
+                .nodes
+                .get(node_name)
+                .unwrap()
```
diptanu, line 400: Don't use unwrap - .map_err(|| anyhow!{"unable to find function {}, node_name})?

server/data_model/src/lib.rs
```diff
@@ -136,6 +159,42 @@ impl Node {
...
+            Node::Router(ref mut router) => {
+                router.image_information.version = ImageVersion(image_version);
+            }
+            Node::Compute(compute) => {
+                compute.image_information.version = ImageVersion(image_version);
+            }
+        }
+    }
+
+    pub fn image_version_assign_next(&mut self) {
```
diptanu, line 187: You are not using this in the creat_or_update_compute_graph. You could rename this to image_version_next() and return the next image and continue using image_version_assign as you have done already.

diptanu (changes requested):

server/data_model/src/lib.rs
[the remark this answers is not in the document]
stangirala: makes sense, moving.

PR #990, opened 2024-10-30
Added an API to create context for graphs

### LINKED ISSUE #983: Graph Context for sharing metadata across functions
## Problem 
Some use-cases requiring sharing information between functions in a graph. Ex, in a PDF processing workflow - 

```python
def parse_docment(file: File) -> Document:
   # file contains metadata such as file name, url ,owner, etc, which needs to be 
  # written into databases but the output of this function only contains Pages 
  # and the file metadata is not shared
   ...
   
def chunk(doc: Document) -> List[Chunk]:
  ...

def embed_and_Write(chunk: Chunk) -> None:
   # Need the file name here so that it can be written into the vector db along side the chunk and embeddings
   # for citations
 ```
 
In this example we want some information from `File` to be available in downstream functions. It's possible to add these in the output of the first function but that becomes cumbersome and often repetitive if the information has to be shared across all functions. 

## Solution -

Provide a shared Dictionary to be mutated and read from every function. A new dictionary is initialized for every Graph invocation and is persisted by Indexify

```python
from indexify import get_graph_ctx


@indexify_function()
def parse_document(file: File) -> Document:
   ctx = get_graph_ctx()
   ctx.state["file_name"] = file.metadata["file_name"]
   ctx.state["some_other_metadata"] = file.some_other_metadata
   .....
   
   
@indexify_function()
def chunk_and_embed(chunk: Chunk) -> None:
   ctx = get_graph_ctx()
   file_name = ctx.state[“file_name”]
   vector_db.write(embeddings, metadata={"file_name": file_name}
```

Any function can get a reference to the Dictionary, and read/write to it to share information across them. In the backend we will implement persistence storage for them transparently.

diptanu:
Added Server API to add some state for a Graph Invocation that can be shared across functions. 

### Set Ctx State Key
```bash
curl -v -X POST -H"Content-Type: application/json" http://localhost:8900/internal/namespace/default/compute_graphs/simple_pipeline/invocations/3f9594e13e133bd9/ctx -d '{"key": "file_name", "value": "foo"}'
Note: Unnecessary use of -X or --request, POST is already inferred.
*   Trying 127.0.0.1:8900...
* Connected to localhost (127.0.0.1) port 8900 (#0)
> POST /internal/namespace/default/compute_graphs/simple_pipeline/invocations/3f9594e13e133bd9/ctx HTTP/1.1
> Host: localhost:8900
> User-Agent: curl/7.81.0
> Accept: */*
> Content-Type: application/json
> Content-Length: 132
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< vary: origin, access-control-request-method, access-control-request-headers
< access-control-allow-origin: *
< content-length: 0
< date: Wed, 30 Oct 2024 07:04:27 GMT
< 
* Connection #0 to host localhost left intact
* 
```

### Get Context Key 

```bash
curl -X GET -v -H"Content-Type: application/json"  http://localhost:8900/internal/namespace/default/compute_graphs/simple_pipeline/invocations/3f9594e13e133bd9/ctx -d '{"key": "file_name"}'
*   Trying 127.0.0.1:8900...
* Connected to localhost (127.0.0.1) port 8900 (#0)
> GET /internal/namespace/default/compute_graphs/simple_pipeline/invocations/3f9594e13e133bd9/ctx HTTP/1.1
> Host: localhost:8900
> User-Agent: curl/7.81.0
> Accept: */*
> Content-Type: application/json
> Content-Length: 20
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< content-type: application/json
< content-length: 15
< vary: origin, access-control-request-method, access-control-request-headers
< access-control-allow-origin: *
< date: Wed, 30 Oct 2024 07:07:50 GMT
< 
* Connection #0 to host localhost left intact
{"value":"foo"}
```

Fixes #983 

[x] Run `make fmt`.


server/state_store/src/kv.rs
```diff
@@ -0,0 +1,48 @@
...
+        let options = DbOptions::default();
+        let kv_store = Db::open_with_opts(Path::from(path), options, object_store).await?;
+        Ok(KVS {
+            kv_store: Arc::new(kv_store),
+        })
+    }
+
+    pub async fn put_ctx_state(&self, req: WriteContextData) -> Result<()> {
+        let key = format!(
+            "{}.{}.{}.{}",
```
seriousben, line 30: Could we reuse the same key format used in rocksdb?

server/src/http_objects.rs
```diff
@@ -517,6 +517,28 @@ pub struct InvocationQueryParams {
...
 }
 
+#[derive(Debug, Serialize, Deserialize, ToSchema)]
+pub struct CtxStatePutRequest {
+    pub key: String,
+
+    // Could have encoded this as string but
+    // making sure we get valid json from user
+    // code
+    pub value: serde_json::Value,
```
seriousben, line 527: If we enforce that values are json, I wonder if we should not also enforce the max size of values.

I could see a future server configuration where per-namespace/per-organization soft and hard-limits (for this features but also for many others) are defined to both protect the server and users but still allow for more specific usage. On the cloud, there could possibly be a billing implication to increasing some of these limits.

server/src/service.rs
```diff
@@ -35,8 +36,15 @@ impl Service {
         let indexify_state = IndexifyState::new(self.config.state_store_path.parse()?).await?;
         let blob_storage = Arc::new(BlobStorage::new(self.config.blob_storage.clone())?);
         let executor_manager = Arc::new(ExecutorManager::new(indexify_state.clone()).await);
+        let object_store = self.config.blob_storage.object_store()?;
+        let kvs_manifest_path = Path::new(&self.config.state_store_path).join("graph_ctx_state");
+        let kvs_manifest_path = kvs_manifest_path
+            .to_str()
+            .ok_or(anyhow!(format!("unable to create kv store pathl")))?;
+        let kvs = KVS::new(object_store.clone(), kvs_manifest_path).await?;
```
seriousben, line 44: Instead of getting the object_store from blob_storage config, can we make KVS support getting a BlobStorage instance and from it, get the object_store?

```suggestion
        let kvs_manifest_path = kvs_manifest_path
            .to_str()
            .ok_or(anyhow!(format!("unable to create kv store pathl")))?;
        let kvs = KVS::new(blob_storage.clone(), kvs_manifest_path).await?;
```

seriousben (approved):
LGTM, please consider my comments. But this is good to merge if you don't think they are actionable.

server/src/http_objects.rs
diptanu: @seriousben We are not enforcing max size for values. Do you see that anywhere? If we are, I probably made a mistake. Raw bytes is probably better than JSON yeah.

server/src/service.rs
diptanu: @seriousben Updated the logic here, lmk what you think. Simplified the config to have only the path, and from there create an object store object.

Issue #984, opened 2024-10-28
Add support for Json Serialization


stangirala:
## Problem
Right now indexify supports Cloud Pickle and Message Pack for Serialization. This means that the data written to the blob store will saved as binaries and can only be retrieved as usable/human readable data using the Python SDK (Indexify Client). We already have the code and abstraction setup to swap the serializer. This code is only in the Python SDK.

We would need to modify `get_serializer` in object_serializer.py to make this happen.

## User DX

The DX after making this change would be,

```
@indexify_function(payload_encoder="json")
def simple_function(x: MyObject) -> MyObject:
    return MyObject(x=x.x + "b")
```

## Testing
1. Test graph where payload encoder is not specified anywhere (make sure it uses default). Test the outputs (ie. use an exiting graph and check that the tests passes).
2. Test graph where one function has the json encoder. Verify that functions outputs are retrievable using the curl endpoint and are in json.
3. All functions in the graph are specified to use the json encoder. Verify that functions outputs are retrievable using the curl endpoint and are in json.

Default2882:
Hi, can you please assign this to me? I'd like to work on it, thanks!

diptanu:
This was completed

PR #991, opened 2024-10-30
(feat): Add support for Json Serialization #984

Default2882:
Adds support for Json Serialization #984
Adds `output` method in `Pipeline` class
Made changes in unittests to run locally.


python-sdk/indexify/functions_sdk/data_objects.py
```diff
@@ -16,7 +16,7 @@ class RouterOutput(BaseModel):
 
 class IndexifyData(BaseModel):
     id: Optional[str] = None
-    payload: bytes
+    payload: Union[bytes | str]
```
diptanu, line 19: you could leave this as bytes. As the payloads can be decoded from bytes to string in the JSONEncoder - `payload.encode("utf-8")`

python-sdk/indexify/functions_sdk/data_objects.py
```diff
@@ -16,7 +16,7 @@ class RouterOutput(BaseModel):
...
     payload_encoding: str = "cloudpickle"
```
diptanu, line 20: Literal["cloudpickle", "json"]

python-sdk/indexify/functions_sdk/indexify_functions.py
```diff
@@ -297,28 +297,36 @@ def invoke_router(self, name: str, input: IndexifyData) -> RouterCallResult:
...
-        if is_pydantic_model_from_annotation(arg_type):
-            if len(payload.keys()) == 1 and isinstance(list(payload.values())[0], dict):
-                payload = list(payload.values())[0]
-            return arg_type.model_validate(payload)
-        return payload
+        payload_encoder = self.indexify_function.payload_encoder
+        payload = indexify_data.payload
+
+        if payload_encoder == "cloudpickle":
+            return CloudPickleSerializer.deserialize(payload)
```
diptanu, line 304: You can get the serializer from the get_serializer method in object_serializer

python-sdk/indexify/functions_sdk/indexify_functions.py
```diff
@@ -297,28 +297,36 @@ def invoke_router(self, name: str, input: IndexifyData) -> RouterCallResult:
...
+        if payload_encoder == "json":
+            return JsonSerializer.deserialize(payload)
+        elif payload_encoder == "msgpack":
```
diptanu, line 307: Payload encoder can't be msgpack, at the moment it's cloud pickle or JSON. So what happens is - 

1. We convert the payload to bytes by either encoding it into cloud pickle or json (and then convert that to bytes)
2. Convert the entire IndexifyData object to bytes using MSGPack. There is really no need to do this if we upload the payload first into indexify server, get some URL back and then encode the IndexifyData as JSON (since it won't have the raw bytes anymore).  But this could be done in another PR.

python-sdk/indexify/functions_sdk/object_serializer.py
```diff
@@ -12,8 +13,26 @@ def get_serializer(serializer_type: str) -> Any:
...
-    else:
-        raise ValueError(f"Unknown serializer type: {serializer_type}")
+    elif serializer_type == "json":
+        return JsonSerializer()
+    raise ValueError(f"Unknown serializer type: {serializer_type}")
+
+class JsonSerializer:
+    @staticmethod
+    def serialize(data: Any) -> str:
+        return jsonpickle.encode(data)
```
diptanu, line 23: what is jsonpickle for? json.dumps should be just fine no?

diptanu (changes requested):

vcknorket:
@diptanu @Default2882 Is this expected to be merged soon?

python-sdk/indexify/functions_sdk/object_serializer.py
Default2882: the standard json lib does not support conversion of complex objects. Example - 
```
>>> from dataclasses import dataclass
>>> @dataclass
... class Example:
...     data: str
...
>>> ex = Example("test")
>>> import json
>>> json.dumps(ex)
  File "<stdin>", line 1, in <module>
  File "/opt/homebrew/Cellar/python@3.10/3.10.15/Frameworks/Python.framework/Versions/3.10/lib/python3.10/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/opt/homebrew/Cellar/python@3.10/3.10.15/Frameworks/Python.framework/Versions/3.10/lib/python3.10/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/opt/homebrew/Cellar/python@3.10/3.10.15/Frameworks/Python.framework/Versions/3.10/lib/python3.10/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/opt/homebrew/Cellar/python@3.10/3.10.15/Frameworks/Python.framework/Versions/3.10/lib/python3.10/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type Example is not JSON serializable
>>> import jsonpickle
>>> jsonpickle.encode(ex)
'{"py/object": "__main__.Example", "data": "test"}'
>>>
```

If we want to use the standard json lib, we'd have to write custom json encoding and decoding as shown [here](https://docs.python.org/3/library/json.html) -

#### Specializing JSON object encoding:
```
import json
def custom_json(obj):
    if isinstance(obj, complex):
        return {'__complex__': True, 'real': obj.real, 'imag': obj.imag}
    raise TypeError(f'Cannot serialize object of {type(obj)}')

json.dumps(1 + 2j, default=custom_json)
```

python-sdk/indexify/functions_sdk/data_objects.py
Default2882: Yes, but this is needed since we are using `jsonpickle` as pointed out by you below.

python-sdk/indexify/functions_sdk/data_objects.py
Default2882: This can be - `msgpack` too right? As seen [here](https://github.com/tensorlakeai/indexify/blob/main/python-sdk/indexify/functions_sdk/object_serializer.py#L13)

python-sdk/indexify/functions_sdk/indexify_functions.py
Default2882: Yes that is what I was going for, but as pointed out in this comment [here](https://github.com/tensorlakeai/indexify/issues/984#issuecomment-2447577537) there's some custom logic related to msgpack parsing. 

Would you suggest moving all the custom logic into `MsgPackSerializer` so that I can directly use `get_serializer `?

python-sdk/indexify/functions_sdk/indexify_functions.py
Default2882: > Payload encoder can't be msgpack, at the moment it's cloud pickle or JSON. 
 But we already have an implementation of supporting msgpack as seen [here](https://github.com/tensorlakeai/indexify/blob/main/python-sdk/indexify/functions_sdk/object_serializer.py#L13)

I don't think I understand comment 1 and 2. Can you provide some code pointers?

python-sdk/indexify/functions_sdk/data_objects.py
diptanu: @Default2882 No. We are not going to support msgpack .

diptanu:
@Default2882 Did you run the graph behavior tests?

python-sdk/tests/test_graph_behaviours.py
```diff
@@ -157,14 +176,27 @@ def test_simple_function(self):
         graph = Graph(
             name="test_simple_function", description="test", start_node=simple_function
         )
-        graph = RemoteGraph.deploy(graph)
```
diptanu, line 160: Why are these deleted? They test the graph with the server.

diptanu (commented):
@Default2882 Please add back the code to deploy the graphs to the server, since these tests are "integration" tests.

python-sdk/tests/test_graph_behaviours.py
Default2882: Added them back

Default2882:
> @Default2882 Did you run the graph behavior tests?

I did, we are supposed to start the server using `cargo build && cargo run` and then start an executor from python-sdk using `poetry run indexify-cli executor` and then submit the tests using `poetry run python tests/test_graph_behaviours.py` right?

Default2882:
> @Default2882 Thanks for your contribution!! 🚀

Most Welcome!

PR #992, opened 2024-10-30
Add metrics endpoint and Scheduler State Read

stangirala:
- [x] Run `make fmt`.


server/src/routes.rs
```diff
@@ -186,6 +187,12 @@ pub fn create_routes(route_state: RouteState) -> Router {
         )
         .route("/ui", get(ui_index_handler))
         .route("/ui/*rest", get(ui_handler))
+
+        // TODO do you need RW lock?
```
diptanu, line 191: Put the lock on the data structures where the counters and such are added

server/src/routes.rs
```diff
@@ -796,6 +803,43 @@ async fn delete_invocation(
...
 
+async fn pull_metrics(
+    State(state): State<RouteState>
+) -> Result<Response<Body>, IndexifyAPIError> {
+
+    let unallocated_tasks = state.indexify_state.reader().unallocated_tasks();
+
+    match unallocated_tasks {
+        Ok(tasks) => {
+            for task in tasks {
```
diptanu, line 814: Move these things out of here into a separate struct. See how we did this in the v1 branch.

server/src/routes.rs
```diff
@@ -796,6 +803,43 @@ async fn delete_invocation(
...
+                let compute_graph = state.indexify_state.reader().get_compute_graph(&task.namespace, &task.compute_graph_name).unwrap().unwrap();
```
diptanu, line 815: Return an error to the caller here. Unwraps are going to panic.

stangirala: Returning internal error.

server/src/routes.rs
stangirala: Moved to observer.

server/src/routes.rs
stangirala: Added a lock on the state. I'm using a counter where I read from the state store and update the counter each time. There is no need to keep the previous value around (what v1 does).

server/src/routes.rs
```diff
@@ -141,6 +143,8 @@ pub struct RouteState {
     pub indexify_state: Arc<IndexifyState>,
     pub blob_storage: Arc<blob_store::BlobStorage>,
     pub executor_manager: Arc<ExecutorManager>,
+    pub metrics_registry: Arc<prometheus::Registry>,
+    pub metrics: Arc<MetricsData>,
```
diptanu, line 147: Where is this being used? Looks like it's not being referenced anywhere

server/src/main.rs
```diff
@@ -22,6 +28,85 @@ struct Cli {
     config: Option<PathBuf>,
 }
 
+#[derive(Clone)]
+struct MetricsData {
```
diptanu, line 32: Move this to a module like metrics.rs. This needs to be sharable from other crates too, since we are doing a lot of business logic in other crates, like the scheduler, state_machine, etc.

stangirala: Done.

server/src/routes.rs
stangirala: This is being setup using the global meter provider. I've changed the name to make this clear. The object itself is setup here `metrics: Arc::new(MetricsData::new(indexify_state.clone())),`

diptanu:
Closing this in favor of a new PR

PR #993, opened 2024-10-31
bug(graph): Fix a panic happening when changing nodes of a graph

seriousben:
1. Adding a PR github action acceptance test to catch regressions.
2. Fix a regression recently introduced preventing updating nodes of graphs.

- [x] Run `make fmt`.

### DIFF
--- a/server/state_store/src/state_machine.rs
+++ b/server/state_store/src/state_machine.rs
@@ -392,22 +392,23 @@ pub(crate) fn create_or_update_compute_graph(
         }
 
         for (node_name, node) in compute_graph.nodes.iter_mut() {
-            let existing_node = existing_compute_graph.nodes.get(node_name).unwrap();
+            // If the node already exists in the graph, check if the image hash has changed
+            if let Some(existing_node) = existing_compute_graph.nodes.get(node_name) {
+                let existing_hash = existing_compute_graph
+                    .nodes
+                    .get(node_name)
+                    .ok_or_else(|| {
+                        anyhow!(
+                            "unable to find function {} in graph {}",
+                            node_name,
+                            compute_graph.name
+                        )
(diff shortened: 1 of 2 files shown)


server/state_store/src/state_machine.rs
```diff
@@ -392,22 +392,23 @@ pub(crate) fn create_or_update_compute_graph(
         }
 
         for (node_name, node) in compute_graph.nodes.iter_mut() {
-            let existing_node = existing_compute_graph.nodes.get(node_name).unwrap();
```
seriousben, line 395: If a new node did not exist in the old graph, this would panic.

diptanu: 100% - good catch

server/state_store/src/state_machine.rs
```diff
@@ -392,22 +392,23 @@ pub(crate) fn create_or_update_compute_graph(
...
-                .image_hash();
-
-            if node.image_hash() != existing_hash {
-                node.set_image_version(existing_node.clone().image_version_next());
+            // If the node already exists in the graph, check if the image hash has changed
+            if let Some(existing_node) = existing_compute_graph.nodes.get(node_name) {
+                let existing_hash = existing_compute_graph
+                    .nodes
+                    .get(node_name)
+                    .ok_or_else(|| {
```
diptanu, line 400: If the new node doesn't exist, and the node is new we don't have to do anything related to updating the image version.

diptanu (changes requested):

server/state_store/src/state_machine.rs
seriousben: Thanks, makes a lot of sense. I think I addressed this.

PR #994, opened 2024-10-31
update agent and http client to show errors in response

diptanu:
We should now see the actual errors from the http responses when SSE based calls fails and not see the cryptic Content Type should be event stream. 

Tested this by forcing the server to return an exception - 

```
File "/Users/diptanuc/Projects/indexify/python-sdk/ve/lib/python3.12/site-packages/indexify/http_client.py", line 273, in invoke_graph_with_object
    raise Exception(f"failed to invoke graph: {resp}")
Exception: failed to invoke graph: not implemented 111
```

[x] Run `make fmt`.

### DIFF
--- a/python-sdk/indexify/executor/agent.py
+++ b/python-sdk/indexify/executor/agent.py
@@ -419,13 +419,14 @@ class ExtractorAgent:
                         json=data,
                         headers={"Content-Type": "application/json"},
                     ) as event_source:
+                        if not event_source.response.is_success:
+                            resp = await event_source.response.aread().decode('utf-8')
+                            console.print(f"failed to register: {str(resp)}")
+                            await asyncio.sleep(5)
+                            continue
                         console.print(
                             Text("executor registered successfully", style="bold green")
                         )
-                        if not event_source.response.is_success:
-                            console.print(f"failed to register: {event_source.response.text}")
-                            await asyncio.sleep(5)
-                            continue
                         async for sse in event_source.aiter_sse():
                             data = json.loads(sse.data)
                             tasks = []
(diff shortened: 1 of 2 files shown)


seriousben (approved):

python-sdk/indexify/executor/agent.py
```diff
@@ -418,6 +418,10 @@ def to_sentence_case(snake_str):
                         console.print(
                             Text("executor registered successfully", style="bold green")
                         )
+                        if not event_source.response.is_success:
```
seriousben, line 421: can you move this before the `executor registered successfully` console.print?

python-sdk/indexify/executor/agent.py
```diff
@@ -418,6 +418,10 @@ def to_sentence_case(snake_str):
...
+                            console.print(f"failed to register: {event_source.response.text}")
```
seriousben, line 422: I am pretty sure `await event_source.response.aread()` is needed for the `text` property to not throw.

python-sdk/indexify/executor/agent.py
```diff
@@ -418,6 +418,10 @@ def to_sentence_case(snake_str):
...
+                            await asyncio.sleep(5)
+                            continue
```
seriousben, line 424: can we move this before the executor registered successfully console.print?

python-sdk/indexify/executor/agent.py
diptanu: You are probably right. This is why I haven't merged, need to test it by returning an error from the server

diptanu: Fixed

python-sdk/indexify/executor/agent.py
diptanu: Done

PR #1002, opened 2024-11-04
chore: append task's exception to stderr

miguelhrocha:
The following changes have been made:

1. **Function Worker Refactor**: The handling and storage of exception messages have been removed from the function worker. Errors are now directed straight to the stderr output, enhancing the clarity of error reporting and streamlining the function execution flow.

2. **Task Diagnostics Improvement**: In the task diagnostics process, references to the `exception_msg` have been eliminated. The focus is now solely on `stdout` and `stderr`, leading to a more concise data payload. This adjustment simplifies the code structure and improves the clarity of diagnostics.

These modifications enhance the overall error handling process, making it more straightforward while reducing unnecessary complexity in the implementation. 

## Checklist

- [x] Run `make fmt`.

### Results from test_graph_behaviours

Ran 8 tests in 1.601s

OK

### Running a failing example

image

image

### DIFF
--- a/python-sdk/indexify/executor/agent.py
+++ b/python-sdk/indexify/executor/agent.py
@@ -18,7 +18,6 @@ from rich.theme import Theme
 from indexify.functions_sdk.data_objects import (
     FunctionWorkerOutput,
     IndexifyData,
-    RouterOutput,
 )
 from indexify.functions_sdk.graph_definition import ComputeGraphMetadata
 from indexify.http_client import IndexifyClient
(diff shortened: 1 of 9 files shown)


python-sdk/indexify/executor/agent.py
```diff
@@ -315,7 +314,7 @@ async def task_launcher(self):
                             task=async_task.task,
                             task_outcome="failure",
                             outputs=[],
-                            errors=str(async_task.exception()),
+                            stderr=str(async_task.exception()),
```
diptanu, line 317: @miguelhrocha If the task fails, we should add the stdout and stderr logs regardless - 
```
stdout=outputs.stdout,
stderr=outputs.stderr,
```
This should be enough

diptanu (changes requested):
Small comment, but looks good overall

python-sdk/indexify/executor/agent.py
diptanu: Oh realizing this would happen if the user code didn't execute and we only had the exception from our code. May be we can capture the traceback here?

PR #1003, opened 2024-11-04
bug(blob): fix support for AWS environment variables

seriousben:
## Context

https://github.com/tensorlakeai/indexify/pull/990 changed how blob_storage was configured. Doing so it made the S3 support default to instance metadata authentication.

## What

- Injecting AWS_ environment variables in config to support credential auth.
- Using the same object_store for blob and kvs storage
- Using a child key to the specify the kv storage path.

## Testing

TODO

## Contribution Checklist

### DIFF
--- a/server/blob_store/src/lib.rs
+++ b/server/blob_store/src/lib.rs
@@ -4,7 +4,7 @@ use anyhow::{anyhow, Result};
 use bytes::{Bytes, BytesMut};
 use futures::{stream::BoxStream, StreamExt};
 use object_store::{
-    aws::AmazonS3ConfigKey,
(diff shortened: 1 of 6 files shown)


server/blob_store/src/lib.rs
```diff
@@ -56,13 +64,48 @@ pub struct BlobStorage {
 
 impl BlobStorage {
     pub fn new(config: BlobStorageConfig) -> Result<Self> {
-        let (object_store, path) = parse_url(&config.path.clone().unwrap().parse::<Url>()?)?;
+        let url = &config.path.clone().unwrap();
```
diptanu, line 67: Remove the unwrap and return the error.

diptanu (commented):
Looks good, left a small comment about unwrap.

PR #1008, opened 2024-11-06
feat: cleanup and remove other namespaces support except default

adithyaakrishna:
Description:

- Removed namespaces selector
- Removed namespaces page
- Updated to only allow default namespace
- Removed compute function column from outputs
- Added compute function to the header


adithyaakrishna:
Updated Header to show Compute Fn 

image

Removed Namespaces page and Now only one namespace can be used i.e. `default`

image

No more compute function column

image

server/ui/src/components/InfoBox.tsx
miguelhrocha, line 1: 💭  I appreciate the refactoring, but for future PRs would you mind creating a PR just for a refactoring? This would help with reviewing speeds and context loading.

miguelhrocha (approved):
This PR looks good, but I do have a small comment to make collaboration easier in the future 👍🏼

server/ui/src/components/InfoBox.tsx
adithyaakrishna: Yes chieff

PR #1013, opened 2024-11-09
Add unittest toggle for remote and local tests

stangirala:
## Context

`test_graph_behaviours.py` has several tests for the core indexify functions. We should be able to run these tests either locally or as a Remote deployment (ie. starting a server and executor). Right now it's hard-coded for Remote testing.

## What
Add an env toggle before running the unit test and use it to either setup a remote or local graph.

## Testing

Verified toggle works.

## Contribution Checklist

### DIFF
--- a/.github/workflows/acceptance_tests.yaml
+++ b/.github/workflows/acceptance_tests.yaml
@@ -111,10 +111,15 @@ jobs:
                 sleep 5
             fi
           done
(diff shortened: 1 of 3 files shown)


diptanu (commented):
This is great! Please merge after the tests passes. You need to update the Github Actions too.

PR #1018, opened 2024-11-11
Add tests for simple graph, with multiple inputs

stangirala:
## Context

Add a couple of tests for a simple graph function where the function takes in multiple inputs. This should work given how we setup the args but we didn't have a test for it. Also add test, for sanity, with json encoding.

## What

Add tests for graph behaviours. This will run as a workflow action.

## Testing
N/A

## Contribution Checklist

### DIFF
--- a/python-sdk/tests/test_graph_behaviours.py
+++ b/python-sdk/tests/test_graph_behaviours.py
@@ -1,3 +1,4 @@
+import sys
 import unittest
 from pathlib import Path
 from typing import List, Union
@@ -26,6 +27,12 @@ class MyObject(BaseModel):
 @indexify_function()
 def simple_function(x: MyObject) -> MyObject:
     return MyObject(x=x.x + "b")
+
+
+@indexify_function()
+def simple_function_multiple_inputs(x: MyObject, y: int) -> MyObject:
+    suf = ''.join(["b" for _ in range(y)])
+    return MyObject(x=x.x + suf)
 
 
 @indexify_function(encoder="json")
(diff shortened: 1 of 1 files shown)


python-sdk/tests/test_graph_behaviours.py
```diff
@@ -208,6 +215,18 @@ def test_simple_function(self, is_remote):
         output = graph.output(invocation_id, "simple_function")
         self.assertEqual(output, [MyObject(x="ab")])
 
+    # @parameterized.expand([(False), (True)])
```
diptanu, line 218: This probably shouldn't be commented.

python-sdk/tests/test_graph_behaviours.py
```diff
@@ -208,6 +215,18 @@ def test_simple_function(self, is_remote):
...
+    def test_simple_function_multiple_inputs(self, is_remote=True):
+        print(sys.version_info.major, sys.version_info.minor)
```
diptanu, line 220: Unintented?

python-sdk/tests/test_graph_behaviours.py
```diff
@@ -28,6 +29,12 @@ def simple_function(x: MyObject) -> MyObject:
     return MyObject(x=x.x + "b")
 
 
+@indexify_function()
```
diptanu, line 32: Probably worth testing if this works with json encoding as well. May be it works.

diptanu (commented):
Left some comments, merge it once the tests are green!

python-sdk/tests/test_graph_behaviours.py
stangirala: Ran into this on my laptop testing,`ModuleNotFoundError: No module named 'test_graph_behaviours'`. I just pushed the PR to run the test and see if it worked. Will fix.

PR #1021, opened 2024-11-12
feature(sdk): support graph deletion

seriousben:
## Context

There is currently no way to delete a compute graph using the SDK.

## What

A new `delete_compute_graph` function is added to the `IndexifyClient`.

## Contribution Checklist

### DIFF
--- a/python-sdk/indexify/http_client.py
+++ b/python-sdk/indexify/http_client.py
@@ -175,6 +175,20 @@ class IndexifyClient:
         for fn_name, fn in graph.nodes.items():
             self._fns[f"{graph.name}/{fn_name}"] = fn
 
+    def delete_compute_graph(
+        self,
+        graph_name: str,
+    ) -> None:
+        """
+        Deletes a graph and all of its invocations from the namespace.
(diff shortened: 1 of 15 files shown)


python-sdk/indexify/http_client.py
```diff
@@ -126,9 +126,7 @@ def with_mtls(
             raise ValueError("Both cert and key must be provided for mTLS")
 
         client = get_sync_or_async_client(
-            cert_path=cert_path,
-            key_path=key_path,
-            ca_bundle_path=ca_bundle_path
+            cert_path=cert_path, key_path=key_path, ca_bundle_path=ca_bundle_path
```
diptanu, line 129: Did you run `make fmt` Feels like this would be re-formatted back to what we had before.

diptanu (approved):

seriousben:
@diptanu - I made it so that linting will be enforced by CI starting. I also made the last release acceptance gated behind a specific label since it is expected to fail a lot until we get to v1.0 - I do think that last release test is useful to be intentional about breaking changes.

Will be merging this asap.

PR #1045, opened 2024-11-20
Setup `document_ai_api_version_workflow.py` example which uses the DocumentAI endpoint for PDF Parsing

stangirala:
## Context

Add DocumentAI example for customer.

## What

Add example where the parse step is an API call.

This is a complete example similar to `workflow.py`. We call OAI to do the structured extraction, and this step should be replaced as needed.

## Testing

Local workflow run and as a Remote Graph.

Screenshot 2024-11-20 at 1 00 01 PM

## Contribution Checklist

### DIFF
--- a/examples/pdf_structured_extraction/document_ai_api_version_workflow.py
+++ b/examples/pdf_structured_extraction/document_ai_api_version_workflow.py
@@ -62,7 +62,7 @@ def tensorlake_document_ai_parse(file: File) -> str:
     resp_json = response.json()
     if resp_json["status"] == "successful":
         # Only return data if the job is successful.
-        return resp_json["result"][0]
+        return resp_json["result"]
 
     return None
 
--- a/examples/pdf_structured_extraction/document_ai_api_version_workflow.py
+++ b/examples/pdf_structured_extraction/document_ai_api_version_workflow.py
@@ -93,7 +93,7 @@ class BillSchema(BaseModel):
     account_no: str = Field(..., description="Account number of the customer")
     address: str = Field(..., description="Address of the customer")
     statement_date: Optional[date] = Field(..., description="Date when the statement was issued")
-    due_date: date = Field(..., description="Date by which the payment is due")
+    due_date: str = Field(..., description="Date by which the payment is due")
     total_amount_due: float = Field(..., description="Total amount due on the bill", ge=0)
     payment_received_since_last_statement: float = Field(..., description="Payment received since the last statement", ge=0)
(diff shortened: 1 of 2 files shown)


examples/pdf_structured_extraction/document_ai_api_version_workflow.py
```diff
@@ -0,0 +1,104 @@
...
+    import requests as rs
+
+    url = "https://api.tensorlake.ai/documents/v1/upload"
+    headers = {
+        "Authorization": "Bearer <EDIT THIS>",
+        "accept": "application/json"
+    }
+
+    with tempfile.NamedTemporaryFile(mode="wb+", suffix=".pdf") as f:
+        f.write(file.data)
```
diptanu, line 32: You can use io.Bytes or something to upload a file already on memory to a server. No need to write into temporary files.

examples/pdf_structured_extraction/document_ai_api_version_workflow.py
```diff
@@ -0,0 +1,104 @@
...
+def tensorlake_document_ai_parse(file: File) -> str:
+    """
+    Call the inkwell parse API. TODO replace the token below for `Bearer`.
+    :param file: The file data we want to parse.
+    :return: The parsed data as a String.
+    """
+    import requests as rs
+
+    url = "https://api.tensorlake.ai/documents/v1/upload"
+    headers = {
```
diptanu, line 26: Can we add a comment here to get API keys from the request access link?

examples/pdf_structured_extraction/document_ai_api_version_workflow.py
```diff
@@ -0,0 +1,104 @@
...
+        f.write(file.data)
+        f.seek(0)
+
+        files = {
+            "file": (f.name, f, 'application/pdf')
+        }
+
+        try:
+            resp = rs.post(url, headers=headers, files=files).json()
+            print(resp)
```
diptanu, line 41: delete

examples/pdf_structured_extraction/document_ai_api_version_workflow.py
```diff
@@ -0,0 +1,104 @@
...
+
+    class Config:
+        populate_by_name = True
+        json_encoders = {
+            date: lambda v: v.strftime("%m-%d-%Y"),
+            Decimal: lambda v: str(v)
+        }
+
+def create_graph() -> Graph:
+    g = Graph(name="bill_workflow_document_ai_version", start_node=tensorlake_document_ai_parse)
```
diptanu, line 87: Where is the structured extraction happening after the parse step?

Issue #1017, opened 2024-11-11
Support Async Functions 


diptanu:
Currently functions in graph has to be synchronous -

```python
def foo(url: str) -> str:
 pass
```

This is a problem because people might want async functions for functions that does IO.

```python
async def foo(url: str) -> str:
  return await httpx.get(url).content
```

The function worker needs to detect if the user function is multiprocess and create an event loop, and call the function with `asyncio.run` or something of that nature. This might require our sdk to have async interfaces as well. That needs to be explored.

diptanu:
cc @Default2882

Default2882:
Will take a look at this! Feel free to assign it to me!

PR #1046, opened 2024-11-20
python-sdk(feat): Support async function #1017

Default2882:
## Context
Currently indexify does not support async functions, this is important because users would like to run functions which are I/O bound. Addresses #1017.

## What
This change refactors the executor agent so that the downloader, function_worker and the task_reporter run in a [ThreadPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#threadpoolexecutor).

## Testing
Added a basic test, will be adding more tests.

## Contribution Checklist


python-sdk/indexify/functions_sdk/indexify_functions.py
```diff
@@ -272,6 +281,30 @@ def run_fn(
         )
         return output, None
 
+    async def run_fn_async(
```
Default2882, line 284: This and other functions can use some refactoring to avoid code duplication.

python-sdk/indexify/executor/agent.py
```diff
@@ -231,132 +157,149 @@ async def task_launcher(self):
...
             for async_task in done:
-                if async_task.get_name() == "get_runnable_tasks":
-                    if async_task.exception():
-                        console.print(
-                            Text("Task Launcher Error: ", style="red bold")
-                            + Text(
-                                f"Failed to get runnable tasks: {async_task.exception()}",
-                                style="red",
+                task_name = TaskEnum.from_value(async_task.get_name())
+                match task_name:
```
Default2882, line 161: This switch case can possibly use a strategy pattern

python-sdk/indexify/executor/function_worker/function_worker.py
```diff
@@ -46,92 +43,80 @@ class FunctionOutput(BaseModel):
...
-        graph_name=graph_name,
-        graph_version=str(version),
-        indexify_client=indexify_client,
-    )
-    function_wrapper = IndexifyFunctionWrapper(
-        cloudpickle.loads(pickled_functions[fn_name]),
-        context,
-    )
-    function_wrapper_map[key] = function_wrapper
+class Job(BaseModel):
```
Default2882, line 46: TODO :- Delete this as it is unused.

python-sdk/indexify/executor/task_reporter/task_reporter.py
```diff
@@ -34,14 +40,38 @@ class ReportingData(BaseModel):
...
-        self, base_url: str, executor_id: str, config_path: Optional[str] = None
+        self,
+        base_url: str,
+        executor_id: str,
+        task_store: TaskStore,
+        config_path: Optional[str] = None,
     ):
         self._base_url = base_url
         self._executor_id = executor_id
         self._client = get_httpx_client(config_path)
```
Default2882, line 51: Make task reported use `IndexifyClient` because using the basic http client does not set the API key.

python-sdk/tests/test_graph_behaviours.py
```diff
@@ -225,7 +225,23 @@ def remote_or_local_pipeline(pipeline, remote=True):
     return pipeline
 
 
+@indexify_function()
```
Default2882, line 228: Add more tests.

python-sdk/indexify/executor/downloader.py
```diff
@@ -69,10 +69,17 @@ async def download_graph(self, namespace: str, name: str, version: int) -> str:
             )
         )
 
-        response = self._client.download_graph(namespace, name)
+        response = self._indexify_client.download_graph(namespace, name)
         os.makedirs(os.path.dirname(path), exist_ok=True)
         with open(path, "wb") as f:
             f.write(response.content)
+        console.print(
```
eabatalov, line 76: This is an informational message without any details. Looks redundant. Is it?

python-sdk/indexify/executor/agent.py
```diff
@@ -395,15 +381,15 @@ def to_sentence_case(snake_str):
     async def _shutdown(self, loop):
         console.print(Text("shutting down agent...", style="bold yellow"))
         self._should_run = False
+        self._thread_pool.shutdown(cancel_futures=True)
         for task in asyncio.all_tasks(loop):
             task.cancel()
 
     def shutdown(self, loop):
-        self._function_worker.shutdown()
```
eabatalov, line 444: Why this line is removed?

python-sdk/indexify/executor/agent.py
```diff
@@ -189,121 +157,139 @@ async def task_launcher(self):
...
+    def _console_log_exception(self, async_task: asyncio.Task):
+        console.print(
+            Text("Task Launcher Error: ", style="red bold")
+            + Text(
+                f"Failed to get runnable tasks: {async_task.exception()}",
+                style="red",
+            )
+        )
+
+    def mark_task_as_failed(self, task: Task, stderr: str = None):
```
eabatalov, line 265: Why this is a public method?
This is a helper function for TaskStore. Should it be defined in the same file as a non-member function?

python-sdk/indexify/executor/agent.py
```diff
@@ -189,121 +157,139 @@ async def task_launcher(self):
...
-                        continue
-                    except Exception as e:
-                        console.print(
-                            Text(
-                                f"Failed to execute task {async_task.task.id}\n",
-                                style="red bold",
+                    case TaskEnum.DOWNLOAD_INPUT_TASK:
+                        async_task: DownloadTask
+                        if async_task.exception():
+                            self._console_log_exception(async_task)
```
eabatalov, line 194: This will log "Failed to get runnable tasks:" which is a misleading message in this case DOWNLOAD_INPUT_TASK. Same for other cases afaik.

python-sdk/indexify/executor/agent.py
eabatalov: Yep, this is a large switch it's be better to just a call a method per each case here.

eabatalov (commented):
There are a few comments to address but overall looks good. Thank you.

python-sdk/indexify/executor/function_worker/function_worker.py
```diff
@@ -75,60 +81,109 @@ def _load_function(
...
+    code_path: str
+    version: int
+    init_value: Optional[IndexifyData] = None
+    invocation_id: Optional[str] = None
 
 class FunctionWorker:
     def __init__(
-        self, workers: int = 1, indexify_client: IndexifyClient = None
+        self,
+        workers: int = get_optimal_process_count(),
```
eabatalov: Here we add extra responsibility on deciding how many worker processes to run. Do we have to add this responsibility here given that it's already configurable by the caller?

python-sdk/indexify/executor/function_worker/function_worker.py
```diff
@@ -75,60 +81,109 @@ def _load_function(
...
-        )
-        self._workers = workers
-        self._indexify_client = indexify_client
+        self._workers: int = workers
+        self._indexify_client: IndexifyClient = indexify_client
+        self.job_queue: mp.Queue[(Future, Job)] = mp.Queue(maxsize=pool_size)
+        self.shutdown_event = mp.Event()
+        self.running_processes: list[mp.Process] = []
+        self.finished_jobs = mp.Queue()
+        self._run()
```
eabatalov: This way we hijack the thread which called the __init__ function. It doesn't look like the caller currently expects this as it's not the previous behavior. We also don't document this important thing in the docstring. Should we create a new thread in this constructor and call self._run() from the new thread instead?

python-sdk/indexify/executor/function_worker/function_worker.py
```diff
@@ -75,60 +81,109 @@ def _load_function(
...
-        input: IndexifyData,
-        code_path: str,
-        version: int,
-        init_value: Optional[IndexifyData] = None,
-        invocation_id: Optional[str] = None,
-    ) -> FunctionWorkerOutput:
+    def _run(self):
+        while not self.shutdown_event.is_set():
+            if not self.job_queue.empty():
+                if len(self.running_processes) < self._workers:
```
eabatalov: Why do we manage our own pool of processes and track how many we run instead of ProcessPoolExecutor letting us do this?

eabatalov (commented):
Thank you for looking into this.

I started to review per commit but I see that there's no clear separation of commits in the PR. I'll then review full changes soon and let you know the feedback.

I also think that all the commits in this PR need to be squashed into a single commit because the commit titles are currently misleading and they'll just make the commit log and history confusing.

python-sdk/indexify/executor/function_worker/function_worker.py
eabatalov: Let's do this :)

python-sdk/indexify/executor/agent.py
```diff
@@ -63,6 +64,9 @@ def __init__(
         name_alias: Optional[str] = None,
         image_version: Optional[int] = None,
     ):
+        event_loop = asyncio.get_event_loop()
+        self._thread_pool = ThreadPoolExecutor(max_workers=num_workers)
```
eabatalov, line 68: Currently ProcessPoolExecutor is used. Why do we switch to ThreadPoolExecutor here? What are the implications of this switch? E.g. how do we make sure that there are no deadlocks between the threads now and in the future?

python-sdk/indexify/executor/agent.py
```diff
@@ -75,100 +79,53 @@ def __init__(
...
         if config_path:
             console.print("Running the extractor with TLS enabled", style="cyan bold")
             self._protocol = "https"
         else:
             self._protocol = "http"
 
+        self._base_url = f"{self._protocol}://{self._server_addr}"
         self._task_store: TaskStore = TaskStore()
         self._executor_id = executor_id
+        console.print("Starting FunctionWorker", style="cyan bold")
```
eabatalov, line 97: Is this a redundant message?

eabatalov (commented):
Overall question about this PR.

Do we need to refactor whole Executor code base to be async to support async functions?

As it's stated in the issue https://github.com/tensorlakeai/indexify/issues/1017:

> The function worker needs to detect if the user function is multiprocess and create an event loop, and call the function with asyncio.run or something of that nature. This might require our sdk to have async interfaces as well. That needs to be explored.

How this PR helps us to get there?

fyi we plan to implement running function code in separate processes soon. In this case it's even less important if Executor runs fully in async event loop or not.

python-sdk/indexify/executor/agent.py
Default2882: Even though we were importing and initialising `ProcessPoolExecutor`, it wasn't being used anywhere. As per the existing implementation of agent, downloader, worker_function, and task_reported, all of them were executing on a single event loop. 

As per - https://docs.python.org/3/library/concurrent.futures.html
ProcessPoolExecutor and ThreadPoolExecutor, behave the same way except the fact that one uses sub processes and one uses threads. 

> E.g. how do we make sure that there are no deadlocks between the threads now and in the future?
Each thread will be assigned a single unit of work, I don't see how we can deadlock here.

python-sdk/indexify/executor/agent.py
Default2882: Yea, we can make this better.

python-sdk/indexify/executor/agent.py
Default2882: Yea you are right, `_console_log_exception` should also have a `message argument`.

python-sdk/indexify/executor/agent.py
Default2882: Yep, this should be a private method and start with an `_`

python-sdk/indexify/executor/task_reporter/task_reporter_utils.py
```diff
@@ -0,0 +1,36 @@
...
+            f"Exception: {type(e).__name__}({e})\n"
+            f"Retries: {task_outcome.reporting_retries}\n"
+            "Retrying...",
+            title="Reporting Error",
+            border_style="error",
+        )
+    )
+
+
+def _log(task_outcome):
```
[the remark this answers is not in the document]
Default2882: > These functions are supposed to be visible outside of task_reporter_utils.py file

No, they shouldn't be. The idea is to use a custom Console theme for different parts of the executor, although I can see that there are some misses related to this in other areas of the code.

Default2882:
> Overall question about this PR.
> 
> Do we need to refactor whole Executor code base to be async to support async functions?
> 
> As it's stated in the issue #1017:
> 
> > The function worker needs to detect if the user function is multiprocess and create an event loop, and call the function with asyncio.run or something of that nature. This might require our sdk to have async interfaces as well. That needs to be explored.
> 
> How this PR helps us to get there?
> 
> fyi we plan to implement running function code in separate processes soon. In this case it's even less important if Executor runs fully in async event loop or not.

I am aware of this and was trying to re-use the event loop to make the executor completely async (also took the liberty of re-factoring the code a bit), and give the consumers of python-sdk the option to run async indexify functions by using ThreadPoolExecutor with an event loop. However if we want to completely move away from event loop and use processes/threads, then the executor needs to be completely re-written due to lack of proper abstractions.

eabatalov:
> > Overall question about this PR.
> > Do we need to refactor whole Executor code base to be async to support async functions?
> > As it's stated in the issue #1017:
> > > The function worker needs to detect if the user function is multiprocess and create an event loop, and call the function with asyncio.run or something of that nature. This might require our sdk to have async interfaces as well. That needs to be explored.
> > 
> > 
> > How this PR helps us to get there?
> > fyi we plan to implement running function code in separate processes soon. In this case it's even less important if Executor runs fully in async event loop or not.
> 
> I am aware of this and was trying to re-use the event loop to make the executor completely async (also took the liberty of re-factoring the code a bit), and give the consumers of python-sdk the option to run async indexify functions by using ThreadPoolExecutor with an event loop. However if we want to completely move away from event loop and use processes/threads, then the executor needs to be completely re-written due to lack of proper abstractions.

Yep, we'll discuss this internally and I'll come back to you in a few days with an answer about the plan for the code base.

python-sdk/indexify/executor/agent.py
eabatalov: Yep, this is right. Indeed ProcessPoolExecutor is not used. It seems unnecessary to explicitly create a ThreadPoolExecutor. See https://stackoverflow.com/questions/60204054/default-executor-asyncio. A ThreadPool is required for Python event loop anyway cause some operations like file reads/writes are not supported in asyncio yet.

eabatalov:
> > I am aware of this and was trying to re-use the event loop to make the executor completely async (also took the liberty of re-factoring the code a bit), and give the consumers of python-sdk the option to run async indexify functions by using ThreadPoolExecutor with an event loop. However if we want to completely move away from event loop and use processes/threads, then the executor needs to be completely re-written due to lack of proper abstractions.
> 
> Yep, we'll discuss this internally and I'll come back to you in a few days with an answer about the plan for the code base.

We're going to implement function execution in separate processes in OSS version next week. It's going to be a sizable patch. This allows functions to free all GPU resources they consume once finished and provides some other benefits.

Default2882:
>We're going to implement function execution in separate processes in OSS version next week. It's going to be a sizable patch. This allows functions to free all GPU resources they consume once finished and provides some other benefits.

Got it, should I cancel this PR? Because it won't be relevant anymore.

diptanu:
@Default2882 Lets close this since Eugene is re-writing the executor, we can revisit this feature after the new executor has landed.

Issue #1037, opened 2024-11-18
error in response_body for s3 uploads using json encoder 


sadath-12:
I'm following https://github.com/tensorlakeai/indexify/blob/main/examples/readme/readme_example.py excpet that I'm I set ```encoder="json"``` in ever function . 

When I run it I get 
```
{"timestamp":"2024-11-18T07:57:14.465179Z","level":"INFO","message":"Checking for error in response_body: <?xml version=\"1.0\" encoding=\"UTF-8\"?>\n\n<CompleteMultipartUploadResult xmlns=\"http://s3.amazonaws.com/doc/2006-03-01/\"><Location>https://s3.us-east-1.amazonaws.com/indexify_bucket/afa2c059-7e42-4f3a-8e62-</Location><Bucket>indexify_bucket</Bucket><Key>afa2c059-7e42-4f3a</Key><ETag>\"a11d6ef96cf2fa93a2d521b31d1ce2c8-1\"</ETag></CompleteMultipartUploadResult>","target":"object_store::client::retry","span":{"matched_path":"/namespaces/:namespace/compute_graphs/:compute_graph/invoke_object","method":"POST","uri":"/namespaces/default/compute_graphs/sequence_summer/invoke_object?block_until_finish=true","name":"request"}}
```
And the task is hanged up there .

My environment 

I used k8s to deploy the indexify-server and executor with values.yaml

```yaml
blobStore:
  allowHTTP: true
    #endpoint: http://blob-store:9000
  credentialSecret: blob-creds
  config:
    backend: s3
    path: s3://indexify_bucket
    dynamodb_table: graph_ctx
    s3:
      bucket: indexify_bucket
      region: us-east-1
      accessKey: 123
      secretKey: 123
```

And used indexify version 
```Version: 0.2.29```

**Comment by sadath-12** [comment]
The uploads to s3 increases latency and I was running into timeouts . Increasing timeouts was the workaround

**Comment by seriousben** [comment]
Thanks for reporting this.

More info:

When uploading large fields or lots of different function outcomes, the executor's Task Report can be stuck in a retry loop that will never succeed.

Error seen:
```
task-reporter failed to report task outcome timed out.

╭──────────────────────────────────── Reporting Error ─────────────────────────────────────╮
│ Failed to report task 902492df-006e-4c63-8c2d-660101f7f12e                               │
│ Exception:timed out   Retrying...                                                        │
╰──────────────────────────────────────────────────────────────────────────────────────────╯
```

[...]

PR #1048, opened 2024-11-20
bug(sdk): increase task reporting timeout from 5s to 5m

seriousben:
## Context

Fixes https://github.com/tensorlakeai/indexify/issues/1037

The httpx client library sets various timeouts to 5s. Unfortunately, when uploading large files or a lot of files, this is not sufficient.

## What

This PR does 2 things:
- increase visibility into task reporting retries by modifying logs and adding a retry counter.
- increase the default 5s Read Timeout to 5 minutes.

Note that while this should increase the success rate of most graphs outcome reporting, there are still some graphs with a very big number of files or very large files that would still have issues being uploaded. An alternative solution will be needed for example writing directly to blob storage using a pre-signed URL, splitting uploads in multiple requests to the server, doing chunk uploads, ...

## Testing

Using S3 run this graph:

<details>
<summary>Code</summary>

```python
import subprocess, os, platform, sys
from typing import List, Tuple
from indexify import RemoteGraph, Graph, Image, indexify_function
import time

# @indexify_function(encoder="json")
@indexify_function()
def start_node(text: str) -> List[Tuple[int, str]]:
    return [(index, char) for index, char in enumerate(text)]


@indexify_function()
def concat_char_idx(char_data: Tuple[int, str]) -> str:
    return f"{char_data[1]}={char_data[0]}"


class Acc(BaseModel):
    values: List[str] = list()


@indexify_function(accumulate=Acc)
def append_to_list(acc: Acc, char_data: str) -> List[str]:
    char, index = char_data.split("=")
    if len(acc.values) <= int(index):
        acc.values.extend([""] * (int(index) - len(acc.values) + 1))
    acc.values[int(index)] = char
    return acc


@indexify_function()
def end_node(acc: Acc) -> str:
    return "".join(acc.values)


if __name__ == "__main__":
    graph = Graph(name="stress-test_graph", start_node=start_node)
    graph.add_edge(start_node, concat_char_idx)
    graph.add_edge(concat_char_idx, append_to_list)
    graph.add_edge(append_to_list, end_node)
    graph = RemoteGraph.deploy(graph)

    text = "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Morbi metus ipsum, tincidunt a ligula sed, feugiat luctus dui."
    # text = "hello world"
    invocation_id = graph.run(block_until_done=True, text=text)
    output = graph.output(invocation_id, "end_node")

    print(output)

```

</details>

## Contribution Checklist

### DIFF
--- a/python-sdk/indexify/executor/agent.py
+++ b/python-sdk/indexify/executor/agent.py
@@ -112,9 +112,15 @@ class ExtractorAgent:
         console.print(Text("Starting task completion reporter", style="bold cyan"))
         # We should copy only the keys and not the values
         url = f"{self._protocol}://{self._server_addr}/write_content"
+
         while True:
             outcomes = await self._task_store.task_outcomes()
             for task_outcome in outcomes:
+                retryStr = (
+                    f"\nRetries: {task_outcome.reporting_retries}"
+                    if task_outcome.reporting_retries > 0
+                    else ""
+                )
                 outcome = task_outcome.task_outcome
                 style_outcome = (
                     f"[bold red] {outcome} [/]"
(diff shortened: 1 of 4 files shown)


diptanu (approved):

seriousben:
Tracking the more involved improvement to this in https://github.com/tensorlakeai/indexify/issues/1049

PR #1050, opened 2024-11-20
bug(sdk): fix using json encoding when using an api key

seriousben:
## Context

Content-Type would be removed from headers when an API key was added.

## What

When adding an API key merge it with the existing headers.

## Contribution Checklist


diptanu:
Amazing :D

PR #1053, opened 2024-11-21
Remove pytest from python-sdk and fix commands that runs integ tests

eabatalov:
Commands that run integration tests in python-sdk currently use
pytest but none of the integration tests are using pytest, they
are using unittest framework instead. So all the commands used
for running tests in python-sdk dir are broken right now.

Changes:

1. Remove pytest from python-sdk project dependencies using poetry cli.
   1.1. The CLI update pyproject.toml and package lock file.
   1.2. Remove .coverage code coverage database. It was created a while ago
        when pytest was used by pytest-cov plugin.
2. Fix tests/run_tests.sh command. This command is not currently used in any
   of our GitHub actions and etc but in the future we'll make this a standard
   command for running integ tests in python SDK dir.
3. Update Makefile to run tests/run_tests.sh in test target.

Testing:

```
make test
```

```
tests/run_tests.sh
```
Exists with non zero exist status when any of the tests fail due to use of xargs.
The tests are running sequentially by xargs.


python-sdk/Makefile
```diff
@@ -18,15 +18,12 @@ lint:
 	@poetry run black --check .
 
 test:
-	@poetry run pytest --cov=./indexify
-
-watch:
-	@poetry run ptw
```
eabatalov, line 24: ptw is pytest watcher. We're not using pytest and thus pytest watcher anywhere in python-sdk dir.

diptanu:
@eabatalov please fix the build and add the make target to GitHub actions

eabatalov:
I'll add the make target to GH actions with a fix for integ tests because many of them are broken now. This will be in a separate CR. The lint check passes locally for me but I'll try to fix it in rev2 of this PR.

Issue #1067, opened 2024-11-26
server: create_tasks can panic if graph invocation ctx does not exist


seriousben:
We currently log an error without returning the error. After which we unwrap the ctx which panics.

PR #1068, opened 2024-11-26
No panic on create tasks if the graph ctx is none

seriousben:
## Context

Fixes https://github.com/tensorlakeai/indexify/issues/1067

## What

Removing unwraps in start_tasks and replay_compute_graph.

## Testing

Using the graph behavior test.

## Contribution Checklist

### DIFF
--- a/server/state_store/src/state_machine.rs
+++ b/server/state_store/src/state_machine.rs
@@ -639,13 +639,23 @@ pub(crate) fn create_tasks(
         "{}|{}|{}",
         req.namespace, req.compute_graph, req.invocation_id
     );
-    let graph_ctx = txn
-        .get_for_update_cf(
-            &IndexifyObjectsColumns::GraphInvocationCtx.cf_db(&db),
-            &ctx_key,
-            true,
-        )?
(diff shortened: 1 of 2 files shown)


diptanu (approved):

server/state_store/src/state_machine.rs
```diff
@@ -638,15 +638,14 @@ pub(crate) fn create_tasks(
...
-    let graph_ctx = txn.get_for_update_cf(
-        &IndexifyObjectsColumns::GraphInvocationCtx.cf_db(&db),
-        &ctx_key,
-        true,
-    )?;
-    if graph_ctx.is_none() {
-        error!("Graph context not found for graph: {}", req.compute_graph);
-    }
-    let mut graph_ctx: GraphInvocationCtx = JsonEncoder::decode(&graph_ctx.unwrap())?;
+    let graph_ctx = txn
```
diptanu, line 641: We can't return an error here if the graph context is not found. The graph might be deleted before this is being called here. 

We need to see if invocation is completed and return Ok(None) here. Since InvocationCompletion is meaningless if the invocation was deleted int he first place.

diptanu (changes requested):

PR #1069, opened 2024-11-26
Address various scheduler timing issues

seriousben:
## Context

Running test_graph_behavior.py continuously with as many as 9 executors results in various failures that do not show themselves with only one executor.

Error seen and addressed:
1. Executor stuck in an infinite loop when a scheduler loop processes 2 state changes.
2. Reducer function returning more than the expected single output when it a reducer finishes before the next parent output finishes.
3. Reducer function returning more than the expected single output when the ingest file for it happens right before a scheduler run loop.

## What

In this PR, on top of addressing the edge cases found we are also adding lots of traces and making sure a scheduler error will not block the loop to other state changes.

### Reducer problem 1

image

image

### Reducer problem 2

image

image

Known edge case to address in a future PR: the scheduler run loops expects to process state changes for a single compute graph at a time. This is an incorrect assumption and can results in edge cases.

## Testing

In order to test fixes and detect edge cases, I have detected errors by running the following:

```
TEST_MAX=500 INDEXIFY_URL=http://localhost:8900 command_stress_test poetry run python -u -m unittest tests/test_graph_behaviours.py 2>&1 | tee test-out.log
```

`command_stress_test` is https://github.com/seriousben/serious-nixos-config/blob/main/home-manager/files/command_stress_test.fish

### Before these changes:

After 480s (107/500 runs) ALL executors become very quickly stuck in a loop doing ingest_file for already finished tasks.

We can still infer these failures:
```
     12 FAIL: test_map_reduce_operation_1 (tests.test_graph_behaviours.TestGraphBehaviors.test_map_reduce_operation_1)
     11 FAIL: test_pipeline_1 (tests.test_graph_behaviours.TestGraphBehaviors.test_pipeline_1)
      8 FAIL: test_router_graph_behavior_1 (tests.test_graph_behaviours.TestGraphBehaviors.test_router_graph_behavior_1)
```

### After these changes:

```
===========================
Success Rate = 99.8%
===========================
Failures     = 1
Total        = 500
Elapsed      = 1566s
```

The single failure seen in this run is:

```
FAIL: test_map_reduce_operation_1 (tests.test_graph_behaviours.TestGraphBehaviors.test_map_reduce_operation_1)
  File "/Users/seriousben/Library/Caches/pypoetry/virtualenvs/indexify-dlsxfW2b-py3.11/lib/python3.11/site-packages/parameterized/parameterized.py", line 620, in standalone_func
    return func(*(a + p.args), **p.kwargs, **kw)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/seriousben/src/tensorlakeai/indexify/python-sdk/tests/test_graph_behaviours.py", line 318, in test_map_reduce_operation
    self.assertEqual(output_sum_sq, [Sum(val=5)])
AssertionError: Lists differ: [Sum(val=1)] != [Sum(val=5)]

First differing element 0:
Sum(val=1)
Sum(val=5)

- [Sum(val=1)]
?          ^

+ [Sum(val=5)]
?
```

Future work will look into this other edge case.

## Contribution Checklist


server/src/scheduler.rs
```diff
@@ -48,47 +48,35 @@ impl Scheduler {
...
-                    tasks: result.tasks,
-                };
-                create_task_requests.push(request);
-                new_reduction_tasks.extend(result.new_reduction_tasks);
-                processed_reduction_tasks.extend(result.processed_reduction_tasks);
             }
         }
         let mut new_allocations = vec![];
-        for state_change in &state_changes {
+        for state_change in &processed_state_changes {
```
seriousben, line 79: Only processing stage changes that were actually correctly processed.

server/src/scheduler.rs
```diff
@@ -48,47 +48,35 @@ impl Scheduler {
...
-                    Some(
-                        handle_task_finished(self.indexify_state.clone(), task, compute_graph)
-                            .await?,
-                    )
+            let span = span!(
+                tracing::Level::INFO,
+                "process_state_change",
+                state_change_id = state_change.id.to_string(),
+                change_type = ?state_change.change_type
+            );
```
seriousben, line 56: Not sure how helpful this is when debugging. At least will provide some fields on error.

server/task_scheduler/src/lib.rs
```diff
@@ -44,37 +49,68 @@ impl TaskScheduler {
...
-                    executor: executor_id.clone(),
-                });
+            let span = span!(
+                tracing::Level::INFO,
+                "scheduling_task",
+                task_id = task.id.to_string(),
+                namespace = task.namespace,
+                compute_graph = task.compute_graph_name,
+                compute_fn = task.compute_fn_name,
+            );
```
seriousben, line 62:
```
{"timestamp":"2024-11-26T21:48:30.717931Z","level":"INFO","message":"scheduling task TaskId(\"7b5dfb6c-7205-4286-985c-e0169b4890ee\")","target":"task_scheduler","span":{"compute_fn":"make_it_string","compute_graph":"simple_pipeline","namespace":"default","task_id":"7b5dfb6c-7205-4286-985c-e0169b4890ee","name":"scheduling_task"}}
{"timestamp":"2024-11-26T21:48:30.718015Z","level":"INFO","message":"assigning task TaskId(\"7b5dfb6c-7205-4286-985c-e0169b4890ee\") to executor ExecutorId(\"iYYptSyk7mS1Z3loBA7AN\")","target":"task_scheduler","span":{"compute_fn":"make_it_string","compute_graph":"simple_pipeline","namespace":"default","task_id":"7b5dfb6c-7205-4286-985c-e0169b4890ee","name":"scheduling_task"}}
```

server/src/scheduler.rs
seriousben:
```
{"timestamp":"2024-11-26T21:48:30.721452Z","level":"INFO","message":"compute graph simple_pipeline invocation finished: false","target":"task_scheduler::task_creator","span":{"change_type":"TaskFinished(TaskFinishedEvent { namespace: \"default\", compute_graph: \"simple_pipeline\", compute_fn: \"make_it_string\", invocation_id: \"5653129b88645fc5\", task_id: TaskId(\"7b5dfb6c-7205-4286-985c-e0169b4890ee\") })","state_change_id":"105","name":"process_state_change"}}
```

Actually quite helpful!

diptanu:
@seriousben Can you run some tests on the following scenarios - 
1. Create a graph, with no executor, invoke the graph(it will create tasks, but no allocations), delete the invocation, bring an executor(it will create allocation) -- expecting something to break here 
2. Create a graph, with an executor, invoke the graph with a function that takes 10 seconds to complete, meanwhile delete the graph, and let the task complete -- expecting something to break. 
3. Run the graph with a map reduce with a sequence of 100000000 and 15 executors running on docker locally. The map and reduce will run concurrently on all the machines -- expecting a lot to break here

server/src/scheduler.rs
```diff
@@ -111,11 +99,41 @@ impl Scheduler {
                 },
                 diagnostic_msgs,
             }),
-            state_changes_processed: processed_state_changes,
+            state_changes_processed: processed_state_changes.iter().map(|x| x.id).collect(),
```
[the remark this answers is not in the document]
seriousben: We are mapping the processes_state_changes to their id.

server/data_model/src/lib.rs
```diff
@@ -391,6 +392,23 @@ impl ComputeGraph {
 
         self
     }
+
+    pub fn get_compute_parent(&self, node_name: &str) -> Option<&str> {
```
diptanu, line 396: get_parerent_nodes

server/data_model/src/lib.rs
```diff
@@ -391,6 +392,23 @@ impl ComputeGraph {
...
+        // Find parent of the node
+        self.edges
```
diptanu, line 398: We could just precompute this in a hash map in the ComputeGraph object. But the logic seems fine.

server/src/scheduler.rs
```diff
@@ -48,57 +48,44 @@ impl Scheduler {
...
+            let span = span!(
+                tracing::Level::INFO,
+                "process_state_change",
+                state_change_id = state_change.id.to_string(),
+                change_type = state_change.change_type.as_ref(),
+            );
+            let _enter = span.enter();
+            match self.process_state_change(state_change).await {
+                Ok(result) => {
+                    processed_state_changes.push(state_change);
```
diptanu, line 60: Can you just add the id here so that you don't have to iterate again and map the list to ids ?

server/state_store/src/state_machine.rs
```diff
@@ -726,6 +729,41 @@ pub fn mark_task_completed(
...
+                &IndexifyObjectsColumns::TaskAllocations.cf_db(&db),
+                &task.make_allocation_key(&req.executor_id),
+            )?
+            .is_some()
+        {
+            error!(
+                task_key = task.key(),
+                "Task already completed but allocation still exists, deleting allocation",
+            );
+            txn.delete_cf(
```
diptanu, line 761: I don't think we should check this in. This feels like a bandaid. Let's investigate some more before we do this.

diptanu (changes requested):

server/state_store/src/state_machine.rs
seriousben: I fixed the root cause as part of this PR. But without this, we risk loosing executors stuck in a bad state.

I think if this happens in the future it should be an alert and we should debug it.

server/data_model/src/lib.rs
seriousben: Done

server/src/scheduler.rs
seriousben: Done

server/state_store/src/state_machine.rs
seriousben: Since the root cause is fixed and this will prevent outages in case a similar problem happens in the future, I would like to keep this and have an alert to get us to investigate and fix other root causes.

server/data_model/src/lib.rs
seriousben: For simplicity and because it is only needed for a edge case, I would like to postpone precomputing it. precomputing comes with challeneges like support for existing graphs that I would prefer not tackle in this PR.

server/src/scheduler.rs
```diff
@@ -43,89 +43,50 @@ impl Scheduler {
...
-            };
-            if let Some(task_placement_result) = task_placement_result {
-                new_allocations.extend(task_placement_result.task_placements);
-                diagnostic_msgs.extend(task_placement_result.diagnostic_msgs);
-            }
+        if requires_task_allocation {
+            let task_placement_result = self.task_allocator.schedule_unplaced_tasks()?;
+            new_allocations.extend(task_placement_result.task_placements);
+            diagnostic_msgs.extend(task_placement_result.diagnostic_msgs);
         }
```
seriousben, line 90: This is what could cause the same task to be allocated multiple times.

seriousben:
Merging to get rid of lots of timing issues. I am happy to make quick changes before next release as needed @diptanu.

PR #1076, opened 2024-12-02
Scheduler and Task Creation improvements

seriousben:
## Context

Fixes:
- If a reducer parent and a reducer task finish at the same time, the scheduler will not know about it which might result in more than one reduced output.
- If a reducer failed for an output, we should start scheduling new reducers while we wait for all parent tasks to complete. (This is will show up as `error processing state change: No outputs found for previous reducer task, should never happen`)

## What

- We skip creating or enqueuing a new reducer task, if the compute node has failed tasks.
  - New unit test covers this case.
- We only do process and persist one state change at a time to prevent timing issues.
  - Existing unit test was changed to test this.

## Contribution Checklist


server/task_scheduler/src/task_creator.rs
```diff
@@ -99,18 +100,18 @@ pub async fn handle_task_finished(
...
+            true
+        } else {
+            info!("task failed invocation finishing, waiting for outstanding tasks to finish");
+            false
+        };
         return Ok(TaskCreationResult::no_tasks(
             &task.namespace,
             &task.compute_graph_name,
             &task.invocation_id,
+            invocation_finished,
```
seriousben, line 114: This is fixing a regression introduced in https://github.com/tensorlakeai/indexify/pull/1072

PR #1078, opened 2024-12-04
chore(dead code): remove unused invocation_finished on scheduler TaskCreationResult

seriousben:
## Context

 
 CreationResult has an invocation_finished attribute that is misleading since it is not used even though we have lots of logic around it in the scheduler.

## What

Delete the attribute and simplify code setting it.

## Contribution Checklist

### DIFF
--- a/server/task_scheduler/src/lib.rs
+++ b/server/task_scheduler/src/lib.rs
@@ -15,22 +15,15 @@ pub struct TaskCreationResult {
     pub tasks: Vec<Task>,
     pub new_reduction_tasks: Vec<ReduceTask>,
     pub processed_reduction_tasks: Vec<String>,
-    pub invocation_finished: bool,
     pub invocation_id: String,
 }
 
 impl TaskCreationResult {
-    pub fn no_tasks(
-        namespace: &str,
-        compute_graph: &str,
-        invocation_id: &str,
-        invocation_finished: bool,
-    ) -> Self {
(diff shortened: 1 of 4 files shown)


server/task_scheduler/src/lib.rs
```diff
@@ -15,22 +15,15 @@ pub struct TaskCreationResult {
...
-    pub invocation_finished: bool,
     pub invocation_id: String,
 }
 
 impl TaskCreationResult {
-    pub fn no_tasks(
-        namespace: &str,
-        compute_graph: &str,
-        invocation_id: &str,
-        invocation_finished: bool,
```
diptanu, line 27: Why was this removed? There were cases an invocation was still running but we were not creating new tasks.

server/task_scheduler/src/task_creator.rs
```diff
@@ -30,7 +30,6 @@ pub async fn handle_invoke_compute_graph(
             compute_graph: event.compute_graph.clone(),
             new_reduction_tasks: vec![],
             processed_reduction_tasks: vec![],
-            invocation_finished: false,
```
diptanu, line 33: This should be invocation_finished: true no?

seriousben: That is what is going to happen, but that attribute was not read anywhere.

In Create tasks, it checks for outstanding tasks to finalize.

PR #1079, opened 2024-12-04
Implement Function Executor in a separate process

eabatalov:
A new indexify-cli command called function_executor is added. It
reads function inputs from host file system, runs the function
and saves function outputs in host file system so they can be
read by main Executor process once the function_executor process exits.

The command imitates an RPC server with request-response pattern
but no RPC framework with a network connection is used yet.
We're going to use it later. The request-response pattern helps
in long term while we're adding more features into host
Executor <-> Function Executor protocol.

Function Executor only passes binary data in the protocol messages
this allows to decouple host Executor from Python SDK and gives
a possibility to rewrite it in another non-Python language in the
future. Function Executor is the only component that actually needs
to import Python SDK because it load and runs customer functions
and works with their inputs and outputs represented as Python objects.

The protocol messages are currently encoded using jsonpickle.
This is the right option in short term as it allows to serialize
the protocol objects without any additional workarounds. E.g.
standart Python json module doesn't allow that - it can't serialize
bytes or custom objects. Inheriting protocol messaged from BaseModel
generates familiar constructors and allows to use .field notation
with protocol messages. .field notation is not available when using
TypedDict. In medium term we're going to migrate to an RPC framework
so all these data wiring considerations won't be relevant anymore.

This PR can be reviewed per commit

Testing:

make test
make fmt
make check

## Contribution Checklist

- [N/A] If the server was changed, please run `make fmt` in `server/`.


python-sdk/indexify/executor/agent.py
```diff
@@ -114,21 +90,30 @@ async def task_completion_reporter(self):
                     self._task_reporter.report_task_outcome(completed_task=task_outcome)
                 except Exception as e:
                     # The connection was dropped in the middle of the reporting, process, retry
-                    logging.error(
+                    logger.error(
                         "failed_to_report_task",
                         task_id=task_outcome.task.id,
-                        exception=f"exception: {type(e).__name__}({e})",
+                        exc_info=e,
```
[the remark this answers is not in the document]
seriousben: Sweet!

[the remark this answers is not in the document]
seriousben: I wonder what it looks like in JSON

python-sdk/indexify/function_executor/handlers/run_function/response_helper.py
```diff
@@ -0,0 +1,93 @@
...
+        self, message: str, stdout: str, stderr: str
+    ) -> RunFunctionResponse:
+        stderr = "\n".join([stderr, message])
+        return RunFunctionResponse(
+            task_id=self._task_id,
+            function_output=None,
+            router_output=None,
+            stdout=stdout,
+            stderr=stderr,
+            is_reducer=False,
```
seriousben, line 80: suggestion(non-blocking): Might make sense to keep set this correctly depending on the function to not create confusion.

I think the server should probably get rid of the need to pass is_reducer though, since it should know.

python-sdk/indexify/cli.py
```diff
@@ -6,22 +6,27 @@
...
+from indexify.executor.paths.host import HostPaths
+from indexify.function_executor.server import Server as FunctionExecutorServer
 from indexify.functions_sdk.image import (
     DEFAULT_IMAGE_3_10,
     DEFAULT_IMAGE_3_11,
     Image,
 )
+from indexify.http_client import IndexifyClient
+
+logger = structlog.get_logger(module=__name__)
```
j3m7, line 29: Yes please

seriousben (approved):
This is amazing!!

python-sdk/indexify/function_executor/handlers/run_function/response_helper.py
eabatalov: The reason why I set is_reducer here to false is because if loading the function into process address space fails then I won't be able to determine if it's a reducer or not, see how it's determined here:

```
def _func_is_reducer(func_wrapper: IndexifyFunctionWrapper) -> bool:
    return func_wrapper.indexify_function.accumulate is not None
```

Also I had a bug and I learned that Server currently uses is_reducer. If I put a wrong value their when function succeeds then the whole graph invocation produces wrong results.

eabatalov:
We discussed internally and it looks like we need to implement execution of multiple concurrent tasks in Function Executor. Otherwise it'd break some existing customers that use a shared in-process state to optimize loading of their models and etc.
I'll create a new PR with this functionality implemented + the functionality of this PR.

PR #1083, opened 2024-12-05
feat: include an optional list of tags when deploying a compute graph

miguelhrocha:
## Context

This PR comes from a need for a visual aid to understand which compute graph is being currently deployed on Indexify. 

This can be considered a short-term solution, as there is no constraint on creating graphs with the same tags collection. However, this can enable CI/CD to automatically generate tags based on commit SHAs.

## What

Adds an optional list of tags which are displayed on the Indexify UI.

### Bonus

Sorts the invocations of a compute graph by creation date because that was driving me crazy.

## Contribution Checklist


server/data_model/src/lib.rs
```diff
@@ -345,6 +345,7 @@ pub struct ComputeGraph {
     pub name: String,
     pub description: String,
     pub version: GraphVersion, // Version incremented with code update
+    pub tags: Vec<String>,
```
seriousben, line 348: Can you check the behavior when there are no tags in storage?

I believe this will need `#[serde(default)]`

server/data_model/src/lib.rs
```diff
@@ -345,6 +345,7 @@ pub struct ComputeGraph {
     pub name: String,
     pub description: String,
     pub version: GraphVersion, // Version incremented with code update
```
seriousben, line 347: Can you add a check in `test_compute_graph_update` to validate that tags can be updated (without versioning the graph).

server/data_model/src/lib.rs
seriousben: We want all existing graphs to default to an empty Vec without erroring out.

seriousben (approved):
Please make sure:

1. An older python-sdk can create and update graphs (it won't know about tags)
2. Graphs existing on an older server can be fetched with a default of no tags

This is to make sure we don't require wiping storage or force everyone to upgrade their SDK version on deploy.

python-sdk/indexify/functions_sdk/graph.py
```diff
@@ -65,14 +65,19 @@ def is_pydantic_model_from_annotation(type_annotation):
...
+        description: Optional[str] = None,
+        tags: Optional[List[str]] = None,
     ):
         self.name = name
         self.description = description
         self.nodes: Dict[str, Union[IndexifyFunction, IndexifyRouter]] = {}
         self.routers: Dict[str, List[str]] = defaultdict(list)
         self.edges: Dict[str, List[str]] = defaultdict(list)
         self.accumulator_zero_values: Dict[str, Any] = {}
+        self.tags = tags or []
```
stangirala, line 80: can we rename these to user tags or deployment tags?

seriousben: Can you describe why it would be preferable?

stangirala: I would like it to be very obvious what tags here is to avoid any confusion. Making the naming as obvious as possible helps.

seriousben: Maybe adding comments would help with removing any confusion.

diptanu (changes requested):

PR #1084, opened 2024-12-05
feat(sdk): improved dict function output and positional arguments support

### LINKED ISSUE #1081: JSON Encoding Issue
If I have a indexify function that returns a dict, for example,
```python
def run(self, raw_request: str) -> dict[str, object]:
```
and I use json encoding for the output, indexify will convert this dict into a named arg for the subsequent function and fail that subsequent function.

This happens here, https://github.com/tensorlakeai/indexify/blob/d2b817c46244a8b82e2d4b3f8b0af1bb8979b8e1/python-sdk/indexify/functions_sdk/indexify_functions.py#L280 when we eventually do `**kwargs`.

One way to fix this is to remove the special case treatment for dicts and only use variable positional args (ie. not named-args) and expect the user to align the positions. This might be a natural way to solve the problem. Most people are used to this pattern,

```python
a, b, c = run()

run2(a, b, c)
```

seriousben:
## Context

Fixes #1081

Currently dict outputs would always become kwargs which can be surprising.

## What

This PR makes it possible to
- return a tuple and correctly turn it in to args for the next function.
- return a dict and correctly turn it into an arg OR a kwargs depending on function signature.
- kwargs method of doing graph.run still works.

Working examples:

- New: Returning a multiple values / tuple
```py
@indexify_function()
def my_func(x: int) -> tuple:
    return 1, 2, 3

@indexify_function()
def my_func_2(x: int, y: int, z: int) -> int:
    return x + y + z
```

- New: Returning a multiple values / tuple (JSON)
```py
@indexify_function(input_encoder="json", output_encoder="json")
def my_func(x: int) -> tuple:
    return 1, 2, 3

@indexify_function(input_encoder="json", output_encoder="json")
def my_func_2(x: int, y: int, z: int) -> int:
    return x + y + z
```

- New: Returning a dict and args
```py
@indexify_function()
def my_func(x: int) -> dict:
    return dict(x=1, y=2, z=3)

@indexify_function()
def my_func_2(input: dict) -> int:
    return input["x"] + input["y"] + input["z"]
```

- New: Returning a dict and args (JSON)
```py
@indexify_function(input_encoder="json", output_encoder="json")
def my_func(x: int) -> dict:
    return dict(x=1, y=2, z=3)

@indexify_function(input_encoder="json", output_encoder="json")
def my_func_2(input: dict) -> int:
    return input["x"] + input["y"] + input["z"]
```

- Backward compatible: Returning a dict and kwargs
```py
@indexify_function()
def my_func(x: int) -> dict:
    return dict(x=1, y=2, z=3)

@indexify_function()
def my_func_2(x: int, y: int, z: int) -> int:
    return x + y + z
```

- Returning a dict and kwargs (JSON)
```py
@indexify_function(input_encoder="json", output_encoder="json")
def my_func(x: int) -> dict:
    return dict(x=1, y=2, z=3)

@indexify_function(input_encoder="json", output_encoder="json")
def my_func_2(x: int, y: int, z: int) -> int:
    return x + y + z
```

## Contribution Checklist

### DIFF
--- a/python-sdk/indexify/functions_sdk/graph.py
+++ b/python-sdk/indexify/functions_sdk/graph.py
@@ -147,6 +147,11 @@ class Graph:
         from_node: Union[Type[IndexifyFunction], Type[IndexifyRouter]],
         to_node: List[Union[Type[IndexifyFunction], Type[IndexifyRouter]]],
     ) -> "Graph":
+        if issubclass(from_node, IndexifyRouter):
+            raise ValueError(
+                "Cannot add edges from a router node, use route method instead"
+            )
+
         self.add_node(from_node)
         from_node_name = from_node.name
         for node in to_node:
--- a/python-sdk/indexify/functions_sdk/indexify_functions.py
+++ b/python-sdk/indexify/functions_sdk/indexify_functions.py
@@ -294,7 +294,6 @@ class IndexifyFunctionWrapper:
     def run_router(
         self, input: Union[Dict, Type[BaseModel]]
     ) -> Tuple[List[str], Optional[str]]:
-        # kwargs = input if isinstance(input, dict) else {"input": input}
         args = []
         kwargs = {}
         try:
--- a/python-sdk/indexify/functions_sdk/indexify_functions.py
+++ b/python-sdk/indexify/functions_sdk/indexify_functions.py
@@ -134,8 +134,8 @@ def _process_dict_arg(dict_arg: dict, sig: inspect.Signature):
     elif len(remaining_kwargs) > 0:
         # If there are remaining kwargs, add them as a single dict argument
         new_args.append(remaining_kwargs)
-
-    return new_args, new_kwargs
+    else:
+        return new_args, new_kwargs
 
 
 def indexify_router(
--- a/python-sdk/indexify/functions_sdk/indexify_functions.py
+++ b/python-sdk/indexify/functions_sdk/indexify_functions.py
@@ -117,7 +117,7 @@ from inspect import Parameter, signature
 from typing_extensions import get_type_hints
 
 
-def _process_dict_arg(dict_arg: dict, sig: inspect.Signature):
+def _process_dict_arg(dict_arg: dict, sig: inspect.Signature) -> Tuple[list, dict]:
     new_args = []
     new_kwargs = {}
     remaining_kwargs = dict_arg.copy()
@@ -134,8 +134,8 @@ def _process_dict_arg(dict_arg: dict, sig: inspect.Signature):
     elif len(remaining_kwargs) > 0:
         # If there are remaining kwargs, add them as a single dict argument
         new_args.append(remaining_kwargs)
-    else:
-        return new_args, new_kwargs
+
+    return new_args, new_kwargs
 
 
 def indexify_router(
--- a/python-sdk/indexify/functions_sdk/indexify_functions.py
+++ b/python-sdk/indexify/functions_sdk/indexify_functions.py
@@ -112,9 +112,30 @@ class IndexifyRouter:
         pass
 
 
-from inspect import signature
+from inspect import Parameter, signature
 
 from typing_extensions import get_type_hints
+
+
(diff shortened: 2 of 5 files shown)


seriousben:
This should be backward compatible at least based on existing test suite.

python-sdk/tests/test_graph_behaviours.py
```diff
@@ -282,28 +294,180 @@ def test_simple_function_multiple_inputs_json(self, is_remote=False):
...
+        graph = remote_or_local_graph(graph, is_remote)
+        invocation_id = graph.run(block_until_done=True, x=1)
+        output = graph.output(invocation_id, my_func_2.name)
+        self.assertEqual(len(output), 1)
+        self.assertEqual(output[0], 6)
+
+    @parameterized.expand([(False), (True)])
+    def test_return_dict_as_kwargs(self, is_remote):
+        @indexify_function()
+        def my_func(x: int) -> tuple:
```
stangirala, line 387: is this correct? we're returning a dict but the type hint is tuple?

Basically we're saying that if the encode mode is json then treat the dict as different. Is that right?

Basically backwards compatibility is maintained if we use `cloudpickle` but broken if we use `json`?

seriousben: I just copy paste without changing type hints

stangirala:
How do we document that we are breaking backwards compatibility for 

```python
@indexify_function()
def my_func(x: int) -> tuple:
    return dict(x=1, y=2, z=3)

@indexify_function()
def my_func_2(x: int, y: int, z: int) -> int:
    return x + y + z
```

This only works for cloudpickle

python-sdk/indexify/functions_sdk/graph.py
```diff
@@ -147,6 +147,11 @@ def add_edges(
         from_node: Union[Type[IndexifyFunction], Type[IndexifyRouter]],
         to_node: List[Union[Type[IndexifyFunction], Type[IndexifyRouter]]],
     ) -> "Graph":
+        if issubclass(from_node, IndexifyRouter):
```
stangirala, line 150: :doge

stangirala:
Can we add a test case where we return multiple dicts?

stangirala:
Can we also add another test case for pydantic models (this is a special case that I ran into).

```python

class P1(BaseModel):
  a: int
```

```python
@indexify_function(input_encoder="json", output_encoder="json")
def fn1(x: int) -> dict:
    return P1(a)

@indexify_function(input_encoder="json", output_encoder="json")
def my_func_2(p: P1) -> int:
    return p.a
```

and another extension
```python
@indexify_function(input_encoder="json", output_encoder="json")
def fn1(x: int) -> dict:
    return P1(a), P1(a+1), P1(a+2)

@indexify_function(input_encoder="json", output_encoder="json")
def my_func_2(a: P1, b: P1, c: P1) -> int:
    return a.a, b.a, c.a
```

python-sdk/tests/test_graph_behaviours.py
stangirala: talked offline - we are backwards compatible for all cases.

python-sdk/indexify/functions_sdk/indexify_functions.py
```diff
@@ -173,6 +201,20 @@ def construct(fn):
...
+                new_args_from_dict, new_kwargs = _process_dict_arg(dict_arg, sig)
+                new_args.extend(new_args_from_dict)
+                return fn(*new_args, **new_kwargs)
+            elif len(args) == 1 and isinstance(args[0], dict):
+                sig = inspect.signature(fn)
+                dict_arg = args[0]
+                new_args, new_kwargs = _process_dict_arg(dict_arg, sig)
+                return fn(*new_args, **new_kwargs)
+
             return fn(*args, **kwargs)
```
stangirala, line 218: might seem stylistic but could we add an else for this return block. Helps read the code that we process inputs with the three conditions.

seriousben: For sure :) 👍

PR #1091, opened 2024-12-09
Image build fixes

j3m7:
## Context

This PR makes some fixes and improvements to image building.  Specifically it matches the python version in built images with the version the builder is running on. This removes the need to keep track of the python version since as long as the workflow and the images are at the same time everything should match.

It also adds the SDK version to the runtime information of graphs. This is related to some future internal work.

## What

- Automatically select the python version in images to match builder version
- Add SDK version to RuntimeInfo
- Fix http_client.Graphs(), added namespace parameter.

## Contribution Checklist


python-sdk/indexify/cli.py
```diff
@@ -150,8 +150,7 @@ def build_image(
 
 @app.command(help="Build default image for indexify")
 def build_default_image():
-    _build_image(image=DEFAULT_IMAGE_3_10)
-    _build_image(image=DEFAULT_IMAGE_3_11)
+    _build_image(image=DEFAULT_IMAGE)
```
seriousben, line 153: Note: we need to continue building default images for open source usage and for the 310 and 311 executors.

diptanu:
@j3m7 the python version needs to match whatever the developer is using. We should detect it from the environment where the code is getting deployed from or introduce an argument in the function decorator that allows developers to specific that.

server/src/http_objects.rs
```diff
@@ -258,13 +258,15 @@ impl From<data_model::Node> for Node {
 pub struct RuntimeInformation {
     pub major_version: u8,
     pub minor_version: u8,
+    pub sdk_version: String,
```
seriousben, line 261: We should probably handle the case of an SDK version that creates a graph without passing the sdk_version.

We could either use `#[serde(default)]` to default to an empty string or we could make the whole field an Option. Maybe an empty string is acceptable for now since we expect all graphs to have an sdk version soon after the new SDK is shipped.

seriousben (approved):
Looks good but please address comments before merging.

j3m7:
> @j3m7 the python version needs to match whatever the developer is using. We should detect it from the environment where the code is getting deployed from or introduce an argument in the function decorator that allows developers to specific that.

This PR does both
- The DEFAULT_IMAGE variable is populated with the running interpreter. 
- When building default images the python version can be specified and we leverage that to build the images for CI.

server/data_model/src/lib.rs
```diff
@@ -98,6 +98,7 @@ pub struct ImageInformation {
     pub run_strs: Vec<String>,
     pub image_hash: String,
     pub version: ImageVersion, // this gets updated when the hash changes
+    pub image_uri: String,
```
seriousben, line 101: Since the image_uri is optional for open-source, I think we should make it optional so that the code has to handle it as optional. This will also make it so that deserializing from the state store an image infomation that does not have the image_uri will not fail.

```suggestion
    pub image_uri: Option<String>,
```

server/src/http_objects.rs
```diff
@@ -91,6 +91,8 @@ pub struct ImageInformation {
     pub tag: String,
     pub base_image: String,
     pub run_strs: Vec<String>,
+    #[serde(default)]
+    pub image_uri: String,
```
seriousben, line 95: To align with the data model, this would become optional.

server/data_model/src/lib.rs
seriousben: serde default might still be needed, I am unsure.

.github/workflows/publish_executor_containers.yaml
```diff
@@ -28,13 +28,20 @@ jobs:
           password: ${{ secrets.DOCKERHUB_TOKEN }}
       - run: |
           pip install indexify -U
-          indexify-cli build-default-image 
+          indexify-cli build-default-image --python-version 3.10
```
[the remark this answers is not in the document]
seriousben: Perfect!

server/data_model/src/lib.rs
diptanu: Do we need this image_uri at all? If you have the image hash, compute graph, namespace, and image name - we should be able to fetch the uri from a service right? 

For OSS - it's upto the operator to start them, and assuming we have code in the executor to report the image hash and name, the task creator would be able to assign functions to the container.

In case we are not building a service which will have the coordinates of the URI it makes sense to encode this in the public API.

server/src/http_objects.rs
seriousben: If we deployed the server, document ai would not be able to deploy new graph versions until it upgrades the SDK.

diptanu: Oh I see, right. That’s fine in this case, we can afford to keep up with new versions of the SDK at the moment. But I understand that it would be a backward incompatible change.

PR #1093, opened 2024-12-10
feat: ensure invocations stay on the same graph version until completion

seriousben:
## Context

While we tracked the graph version that existed when an invocation was created, during its lifetime the scheduler and the executor would start using the new version of the graph for it if an update was done.

## What

This PR ensures that a given invocation will stay on the same graph version until completion.

This PR can be reviewed commit by commit.

What changed:
- In order to stop using an internal endpoint to get the latest version of the code to deserialize outputs, the SDK now uses the output content type to deserialize.
- There are now tests ensuring graph versions are respected during the invocation lifetime.
-  ComputeGraphs have 1 or many ComputeGraphVersions.
    - Compute graph versions are created when a change on the compute graph requires a new version. 
    - Compute graph versions are deleted when the compute graph is deleted
    - Compute graph code is now garbage collected. It was not before.
    - We continue storing latest version data on the compute graph to maintain the same Public API. We can reevaluate if we believe it is necessary to expose specific versions in our Public API at future date.
- Scheduler uses the compute graph version based on the version in the invocation ctx or the task
- A new endpoint was added to retrieve the graph code for a given version of the code. The executor start leveraging it as well.
    - The older endpoint is kept around for now to not break compatibility. But can be removed within a week or so.

## Testing

Two new graph behavior tests were added:
- Change the existing function code while an invocation is running.
- Rename all function code while an invocation is running.

In these two cases, we also change the function code so that it becomes incompatible between versions.

## Contribution Checklist


python-sdk/indexify/http_client.py
```diff
@@ -64,7 +64,6 @@ def __init__(
         self._service_url = service_url
         self._timeout = kwargs.get("timeout")
         self._graphs: Dict[str, Graph] = {}
-        self._fns: Dict[str, IndexifyFunction] = {}
```
stangirala, line 67: what was this used for before that we don't need it now?

python-sdk/tests/test_graph_update.py
```diff
@@ -62,6 +67,134 @@ def test_graph_update(self):
         output = g.output(invocation_id, fn_name="update")
         self.assertEqual(output[0], Object(x="9b"))
 
+    def test_running_invocation_unaffected_diff_types(self):
```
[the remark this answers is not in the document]
stangirala: I'm asking this since we are progressively filling in the concept of graph version. Here from this change in the test it looks like we can swap id for graph version without breaking anything?

If so, then let's break this PR up and be sure about the SDK not breaking.

server/data_model/src/lib.rs
```diff
@@ -374,58 +375,88 @@ impl ComputeGraph {
...
         // self.created_at = other.created_at;
 
         self.description = update.description;
-        self.runtime_information = update.runtime_information;
         self.tags = update.tags;
 
+        let mut graph_version: Option<ComputeGraphVersion> = None;
+
         if self.code.sha256_hash != update.code.sha256_hash ||
+            self.runtime_information != update.runtime_information ||
```
stangirala, line 345: this would handle cases like updating a graph to say `python13.xxx`?

python-sdk/tests/test_graph_update.py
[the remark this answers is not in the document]
seriousben: While there are a few smaller commits I could move out.

This test is actually a failing test that gets fixed in the PR.

server/data_model/src/lib.rs
seriousben: Yeah, updating the SDK or the Python version would version the graph.

server/src/scheduler.rs
```diff
@@ -139,27 +139,35 @@ impl Scheduler {
...
+                        &task.compute_graph_name,
+                        task.graph_version,
+                    )
                     .map_err(|e| {
-                        error!("error getting compute graph: {:?}", e);
+                        error!("error getting compute graph version: {:?}", e);
                         e
                     })?;
-                if compute_graph.is_none() {
+                if compute_graph_version.is_none() {
```
stangirala, line 155: when would this is_none condition be true? If we write a graph don't we also write the version information?

stangirala: ie, wouldn't line 152 be enough for error handling?

server/src/scheduler.rs
```diff
@@ -139,27 +139,35 @@ impl Scheduler {
...
                     error!(
-                        "compute graph not found: {:?} {:?}",
+                        "compute graph version not found: {:?} {:?}",
                         task.namespace, task.compute_graph_name
                     );
                     return Ok(None);
                 }
-                let compute_graph = compute_graph.ok_or(anyhow!(
-                    "compute graph not found: {:?} {:?}",
+                let compute_graph_version = compute_graph_version.ok_or(anyhow!(
```
stangirala, line 162: could this be combined with line 155 as a match statement?

stangirala: like match None => {} and compute_graph_version => {}

stangirala: this might be a stylistic comment - I see the definition for `get_compute_graph_version` has Result<Option<..>>`

server/state_store/src/state_machine.rs
```diff
@@ -547,6 +547,27 @@ pub fn delete_compute_graph(
...
+    for iter in make_prefix_iterator(
+        txn,
+        &IndexifyObjectsColumns::ComputeGraphVersions.cf_db(&db),
+        prefix.as_bytes(),
+        &None,
+    ) {
+        let (key, value) = iter?;
+        let value = JsonEncoder::decode::<ComputeGraphVersion>(&value)?;
+
+        // mark all code urls for gc.
```
stangirala, line 559: what's the gc url?

stangirala:
> In order to stop using an internal endpoint to get the latest version of the code to deserialize outputs, the SDK now uses the output content type to deserialize.

What's the connection between latest version and the deserializer code? This should be independent right?

server/data_model/src/lib.rs
```diff
@@ -374,58 +375,88 @@ impl ComputeGraph {
...
             self.edges != update.edges ||
-            self.nodes != update.nodes ||
-            self.start_fn != update.start_fn
+            self.start_fn != update.start_fn ||
+            self.nodes.len() != update.nodes.len() ||
+            self.nodes.iter().any(|(k, v)| {
```
stangirala, line 349: 🙏

python-sdk/tests/test_graph_update.py
```diff
@@ -62,6 +67,134 @@ def test_graph_update(self):
...
+        graph_name = test_graph_name(self)
+
+        def initial_graph():
+            @indexify_function()
+            def start_node(x: int) -> int:
+                # sleeping in order to make sure the first invocation is running before the graph gets updated.
+                time.sleep(1)
```
stangirala, line 77: could we do something like graph.run(block=True) to ensure that the first invocation completes?

stangirala: or rely on line 116 below?

stangirala: I think the only way you can make this look better is if you check that the job has started running in line 116, and then rely on the first job to run for a "long" period of time (step on does sleep(10s)) or something.

python-sdk/tests/test_graph_update.py
```diff
@@ -62,6 +67,134 @@ def test_graph_update(self):
...
+        g = initial_graph()
+        g = RemoteGraph.deploy(g)
+        first_invocation_id = g.run(block_until_done=False, x=0)
+
+        g = second_graph()
+        # Sleeping in order to make sure the first invocation is running before the graph gets updated.
+        time.sleep(0.25)
+        g = RemoteGraph.deploy(g)
+        invocation_id = g.run(block_until_done=True, x=0)
+        output = g.output(invocation_id, fn_name="end_node")
```
stangirala, line 184: could we return the version information when we retrieve the outputs as well? This helps the client know which output was obtained. This information is implicitly encoded in the invocation_id but not easily readable by the user.

python-sdk/indexify/http_client.py
seriousben: It was only used to keep a cache of deserialized functions to know how to deserialize their output. Since we don't want the SDK to use an internal endpoint and because we can just use the content-type of the output to determine how to deserialiaze this is not needed.

python-sdk/tests/test_graph_update.py
seriousben: Unfortunately, I have'nt found a deterministic way of testing updating a graph while an already running invocation has started while running everything remotely.

The two sleep here make it so that we end up in the situation we are testing most of the time at least locally.

For now, I will keep it like this unless we find a better way to have more determinism.
What I'll do is clarify comments to help our future selves understand more how this is working.

server/src/scheduler.rs
seriousben: Great question, I'll add a comment in the code to make this more obvious.
But this is in the case where the graph is deleted while there is invocation running for it.

server/src/scheduler.rs
seriousben: This is related to the previous comment and to code style.

1) We actually want to return Ok(None) in this case since if the compute graph gets deleted, it is not an error if it is not found.
2) Since at this line compute_graph_version is an Option, we need a way to get the `Some` of the option. We could use a match statement here, but doing so would make all of our code more nested.

Line 152 is an actual error we got retrieving the graph version.
Line 155 is the graph version is not found we don't want to return an error
Line 162 is we just checked that the graph version cannot be None, in order to not use a match statement and have nested code, this time we return an error if it is None. (We could unwrap as well since we just checked and we know it cannot be None, but a panic in the scheduler is bad - and we try to not use unwrap).

server/state_store/src/state_machine.rs
seriousben: Garbage collection is what we call the background process that delete files from blob storage asynchronously.

server/task_scheduler/src/task_creator.rs
```diff
@@ -15,32 +22,93 @@ pub async fn handle_invoke_compute_graph(
     indexify_state: Arc<IndexifyState>,
     event: InvokeComputeGraphEvent,
 ) -> Result<TaskCreationResult> {
-    let compute_graph = indexify_state
+    // let compute_graph = indexify_state
```
seriousben, line 25: TODO: deadcode to delete

server/data_model/src/lib.rs
seriousben: haha

seriousben:
> > In order to stop using an internal endpoint to get the latest version of the code to deserialize outputs, the SDK now uses the output content type to deserialize.
> 
> What's the connection between latest version and the deserializer code? This should be independent right?

@stangirala yes it should be. We now simply look at the content-type of the output to know how to deserialize.

python-sdk/tests/test_graph_update.py
seriousben: I clarified comments and made it a parametrized test to highlight the difference between the two tests.

python-sdk/tests/test_graph_update.py
stangirala: make sense to no add features without thinking through.

> getting the invocation context using the invocation id would allow retrieving the graph version of that invocation.

does this already exist?

PR #1094, opened 2024-12-10
fix(server): make image_uri optional

seriousben:
## Context

Before we got:

```
indexify.error.ApiException: error deserializing from json bytes, missing field `image_uri` at line 1 column 874, value: "data_model::ComputeGraph"
```

When running on a previous state store.

## What

This PR fixes it by making image_uri an Option.

## Testing

Run test_graph_behavior on a previous version of the server and using the same state, run it on the main version.

## Contribution Checklist


stangirala:
+1

PR #1100, opened 2024-12-12
Host Executor uses Functions Executor processes to run customer functions

eabatalov:
Each function version gets a single Function executor process. All tasks for a function version are routed into the same Function Executor process.

Also removed redundant num_workers configuration and some other dead code. This configuration is not used and is now completely obsolete with the Function Executor in separate processes.

Function Executor processes take about 1.2 sec to start and setup a gRPC channel at about p50 with rare durations jumping up to 3 sec p100. I removed all locks from the code and it didn't result in any speedup. I measured that these delays are comming from library calls
`await asyncio.create_subprocess_exec` and
`asyncio.wait_for(channel.channel_ready())`. There's no plan to fix them for now as it'd take a significant effort to figure out what's going on in the libraries. Instead we bump the cold start duration limit in the test to 5 sec from 100 ms. This limit is still okay for the product as we don't have a hard requirement for cold start durations yet.

Looks like gRPC also added about 10 ms of latency to warm starts. This is a minor regression. Doesn't require the test limit change.

Before:

```
cold_start_duration: 0.009701967239379883 seconds
warm_start_duration: 0.006789207458496094 seconds
```

After:

```
cold_start_duration: 1.2409627437591553 seconds
warm_start_duration: 0.016773223876953125 seconds
```

Host Executor is using abstract FunctionExecutor objects so them and their Factory can be replaced in the future if e.g. we want to run Function Executors not as processes but as threads or containers.

This PR aims to have 0 difference in behavior from the current threaded Executor. The "one Function Executor per Function regardless of version" is implemented in the next much smaller PR.

Testing:

make test
make fmt
make check

## Contribution Checklist


python-sdk/indexify/executor/agent.py
```diff
@@ -335,16 +304,16 @@ async def run(self):
...
-        logging.info("shutting_down")
+        logger.info("shutting_down")
         self._should_run = False
+        await self._function_worker.shutdown()
         for task in asyncio.all_tasks(loop):
             task.cancel()
 
     def shutdown(self, loop):
-        self._function_worker.shutdown()
         loop.create_task(self._shutdown(loop))
```
eabatalov, line 319: I noticed lots of orphaned Function Executor processes getting collected in the CI log. This is because we call shutdown only for SIGINT. I'll update this PR tomorrow with SIGTERM and other reasonable signals to fix that.
I'd treat this as a blocking problem because normally daemons are not restarted via SIGINT.

seriousben: Ohh good find!

eabatalov:
Will create a new PR with 2 other commits added because they are important for this review.

PR #1101, opened 2024-12-12
fix(server): make graph versioning backward compatible by falling back on compute graph for latest

seriousben:
## Context

In https://github.com/tensorlakeai/indexify/commit/416ecd8edf42d5051164360788d31969e096b5b7, we introduced running invocations on the same graph version for its duration.

Doing so, we did not take into account backward compatibility support for invocations getting to graphs without a compute graph version. 

## What

In that case, we will rely on the version info stored on the compute graph.

## Contribution Checklist

### DIFF
--- a/server/state_store/src/scanner.rs
+++ b/server/state_store/src/scanner.rs
@@ -23,7 +23,7 @@ use metrics::Timer;
 use opentelemetry::KeyValue;
 use rocksdb::{Direction, IteratorMode, ReadOptions, TransactionDB};
 use serde::de::DeserializeOwned;
-use tracing::{info, instrument};
+use tracing::{debug, instrument};
 
 use super::state_machine::IndexifyObjectsColumns;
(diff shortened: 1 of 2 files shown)


server/state_store/src/scanner.rs
```diff
@@ -551,7 +551,19 @@ impl StateReader {
         let key = ComputeGraphVersion::key_from(namespace, name, version);
         let compute_graph_version =
             self.get_from_cf(&IndexifyObjectsColumns::ComputeGraphVersions, key)?;
-        Ok(compute_graph_version)
+        if compute_graph_version.is_some() {
+            return Ok(compute_graph_version);
+        }
+
+        info!(
```
diptanu, line 558: debug

diptanu (changes requested):

server/state_store/src/scanner.rs
seriousben: Thanks

PR #1103, opened 2024-12-12
Calculate and embed image hash

j3m7:
## Contribution Checklist


diptanu:
Fix lints and merge

server/src/http_objects.rs
```diff
@@ -92,6 +92,7 @@ pub struct ImageInformation {
     pub base_image: String,
     pub run_strs: Vec<String>,
     pub image_uri: Option<String>,
+    pub sdk_version: String,
```
seriousben, line 95: requires serde default to not break when it is missing.

server/data_model/src/lib.rs
```diff
@@ -99,15 +99,23 @@ pub struct ImageInformation {
     pub image_hash: String,
     pub version: ImageVersion, // this gets updated when the hash changes
     pub image_uri: Option<String>,
+    pub sdk_version: String,
```
seriousben, line 102: This will also need serde default to not break when missing from our DB.

server/data_model/src/lib.rs
```diff
@@ -99,15 +99,22 @@ pub struct ImageInformation {
...
+        base_image: String,
+        run_strs: Vec<String>,
+        sdk_version: Option<String>,
+    ) -> Self {
         let mut image_hasher = Sha256::new();
         image_hasher.update(image_name.clone());
-        image_hasher.update(tag.clone());
         image_hasher.update(base_image.clone());
         image_hasher.update(run_strs.clone().join(""));
+        image_hasher.update(sdk_version.clone().unwrap_or("".to_string())); // Igh.....
```
j3m7, line 117: @seriousben @diptanu is there a better way to do this?

j3m7: I'm sending it as is and moving on. We can clean this up later if there is a cleaner way to do this in rust.

PR #1106, opened 2024-12-13
feat(server): allocate by image hash instead of version

seriousben:
## Context

Image hash is deterministic and does not rely on the server to keep track of the next versions. This allows for a better OSS user experience since the hash can be put in the image automatically.

The image version could not be put on the image without more complexity like the image bootstrapper process that was started.

## What

- Change the scheduler selector code to pick an executor with the right image hash
- Change the executor to emit the image hash on registration
- Improved allocation diagnostic message

## Testing

- [x] Ensure default executor as an image work as expected

## Contribution Checklist

### DIFF
--- a/python-sdk/indexify/cli.py
+++ b/python-sdk/indexify/cli.py
@@ -191,10 +191,10 @@ def executor(
         "~/.indexify/executor_cache", help="Path to the executor cache directory"
     ),
     name_alias: Optional[str] = typer.Option(
-        None, help="Name alias for the executor if it's spun up with the base image"
+        None, help="Image name override for the executor"
     ),
-    image_version: Optional[int] = typer.Option(
-        "1", help="Requested Image Version for this executor"
+    image_hash: Optional[str] = typer.Option(
+        None, help="Image hash override for the executor"
     ),
 ):
     if not dev:
@@ -210,7 +210,7 @@ def executor(
         executor_version=executor_version,
         executor_cache=executor_cache,
         name_alias=name_alias,
-        image_version=image_version,
+        image_hash=image_hash,
         dev_mode=dev,
     )
 
@@ -227,7 +227,7 @@ def executor(
         config_path=config_path,
         code_path=executor_cache,
         name_alias=name_alias,
-        image_version=image_version,
+        image_hash=image_hash,
         development_mode=dev,
     )
 
--- a/python-sdk/indexify/executor/runtime_probes.py
+++ b/python-sdk/indexify/executor/runtime_probes.py
@@ -6,13 +6,13 @@ from typing import Any, Dict, Tuple
 from pydantic import BaseModel
 
 DEFAULT_EXECUTOR = "tensorlake/indexify-executor-default"
-DEFAULT_VERSION = 1
-DEFAULT_HASH = "deadbeefcafe"
+# Empty string is used as a default hash which tells the scheduler to accept any hash.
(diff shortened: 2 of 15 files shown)


server/data_model/src/lib.rs
```diff
@@ -241,10 +214,29 @@ impl Node {
...
+                "executor {}, image name: {} does not match function image name {}. Make sure the executor is running the latest image.",
+                executor.id,
+                executor.image_name,
+                self.image_name()
+            ));
+            return false;
         }
+
+        // Empty executor image hash means that the executor can accept any image
+        // version. This is needed for backwards compatibility.
```
seriousben, line 228: Also needed to support running a local executor.

python-sdk/indexify/executor/runtime_probes.py
```diff
@@ -6,13 +6,13 @@
 from pydantic import BaseModel
 
 DEFAULT_EXECUTOR = "tensorlake/indexify-executor-default"
-DEFAULT_VERSION = 1
-DEFAULT_HASH = "deadbeefcafe"
+# Empty string is used as a default hash which tells the scheduler to accept any hash.
```
j3m7, line 9: Definitely an improvement over dead cattle coffee.

server/data_model/src/lib.rs
j3m7: Agreeed

server/task_scheduler/src/lib.rs
```diff
@@ -161,16 +161,6 @@ impl TaskScheduler {
                 }
             }
 
-            if executor.image_name != node.image_name() {
```
j3m7, line 164: Are we messaging on task allocation failures due to hash mismatch?

j3m7 (approved):

python-sdk/indexify/executor/runtime_probes.py
seriousben: :D

server/task_scheduler/src/lib.rs
seriousben: Yes moved here: https://github.com/tensorlakeai/indexify/pull/1106/files/c3f688218b508d29d3d1e9c02b0ea471b0c5afdd#diff-a94d66af396f906968deadceb9f045fb823937877a7c7066f3429145f17b4527R229-R239

```rust
diagnostic_msgs.push(format!(
                "executor {}, image hash: {} does not match function image hash {}. Make sure the executor is running the latest image.",
                executor.id,
                executor.image_hash,
                self.image_hash()
            ));
```

j3m7: Perfect

python-sdk/indexify/cli.py
```diff
@@ -191,10 +191,10 @@ def executor(
         "~/.indexify/executor_cache", help="Path to the executor cache directory"
     ),
     name_alias: Optional[str] = typer.Option(
-        None, help="Name alias for the executor if it's spun up with the base image"
+        None, help="Image name override for the executor"
     ),
-    image_version: Optional[int] = typer.Option(
-        "1", help="Requested Image Version for this executor"
+    image_hash: Optional[str] = typer.Option(
```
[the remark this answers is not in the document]
seriousben: It lives within the image, I did not put too much thought on the override other than it facilitates testing and allows using images without a hash.

PR #1107, opened 2024-12-13
Move everything related to fetching tasks from server into its own class

eabatalov:
This is the first PR where I simplify agent.py by splitting it into components with dedicated responsibilities and interfaces. There's going to be more.

Testing:

make check
make test

## Contribution Checklist

### DIFF
--- a/python-sdk/indexify/executor/agent.py
+++ b/python-sdk/indexify/executor/agent.py
@@ -1,16 +1,10 @@
 import asyncio
-import json
-from importlib.metadata import version
 from pathlib import Path
 from typing import Dict, List, Optional
 
 import structlog
-from httpx_sse import aconnect_sse
 
-from indexify.common_util import get_httpx_client
-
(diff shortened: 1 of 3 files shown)


python-sdk/indexify/executor/agent.py
```diff
@@ -230,6 +226,20 @@ async def task_launcher(self):
                         self._task_store.complete(outcome=completed_task)
                         continue
 
+    async def _main_loop(self):
+        """Fetches incoming tasks from the server and starts their processing."""
+        self._should_run = True
+        while self._should_run:
+            try:
+                async for task in self._task_fetcher.run():
+                    self._task_store.add_tasks([task])
```
eabatalov, line 235: Adding tasks one by one might look inefficient but I'm going to remove task_store in the upcoming PRs. Also I measured duration of `make test` command and it didn't change with this PR. So should be good.

PR #1109, opened 2024-12-17
fix(server): fix support for unversioned code download

seriousben:
## Context

Previous version of the executor are not be able to get unversioned code.

## What

Fix backward compatibility support .

## Testing

- Last release server + executor + make test
- Run main server  + main executor + main make test

## Contribution Checklist


server/src/routes.rs
```diff
@@ -807,7 +807,14 @@ async fn delete_invocation(
     Ok(())
 }
 
-async fn get_code(
+async fn get_unversioned_code(
```
stangirala, line 810: can we remove this in the future?

PR #1111, opened 2024-12-17
fix(sdk): fix to support python 3.10

seriousben:
## Context

In python 3.10, we get `ImportError: cannot import name 'Self' from 'typing' (/usr/lib/python3.10/typing.py)`

## What

Uses the class name as the return type instead of the 3.11 Self feature. 
This uses String 

https://realpython.com/python-type-self string type hinting.

## Contribution Checklist


eabatalov (approved):
Thank you

PR #1114, opened 2024-12-17
chore(CI): testing releasing building on PRs

seriousben:
## Context

Issues preventing publishing of the server are only found at publishing time.

## What

This PR builds the release packages on PRs to detect issues early and aligns x86 with arm64.

It is possible to remove our docker images, cross still defaults to ubuntu 20.04 and there are some node 22 dependency conflicts when using ubuntu 20.04 with what is installed by cross.

See https://github.com/cross-rs/cross/pull/1580 and https://github.com/cross-rs/cross/pull/973

In the future, we should also build images on PR, but since our release image relies on deb, this is impractical for this PR.

## Contribution Checklist


server/Cross.toml
```diff
@@ -1,15 +1,14 @@
 [target.aarch64-unknown-linux-gnu]
 dockerfile = "dockerfiles/Dockerfile.builder_linux_aarch64"
 pre-build = [
-    "apt-get update && apt-get install --assume-yes build-essential npm curl protobuf-compiler",
     "dpkg --add-architecture $CROSS_DEB_ARCH",
     "mkdir -p /.npm",
-    "chown -R 1000:1000 /.npm"
+    "chown -R 1001:128 /.npm"
```
eabatalov, line 6: Looks like some pure magic. Is this Ubuntu specific uid, gid?

eabatalov: :)

eabatalov:
Very nice cleanup + ubuntu update.

PR #1122, opened 2024-12-18
Image copy support

j3m7:
## Context

This adds COPY support for images as well as setting up a foundation for supporting more build operations. 

## What

Refactors image building to execute on a list of BuildOps and ports all existing image building functionality. Also adds support for cloud bases image service.

## Testing

Run `indexify-cli build-default-image` and `indexify-cli build-image` with existing image definitions. Viable images should be built.

## Contribution Checklist


python-sdk/indexify/cli.py
```diff
@@ -161,6 +158,32 @@ def build_image(
                 _create_image(obj, python_sdk_path)
 
 
+@app.command(help="Build platform images for function names")
```
diptanu, line 161: @j3m7 this needs to be in the Tensorlake cli, and wrapped in Tensorlake deploy 

for OSS, we could translate .copy to COPY in the local build context

diptanu (changes requested):

python-sdk/indexify/cli.py
j3m7: Agreed, it's final home will be in the tensorlake cli. added it here so we could start leveraging this functionality on our internal build pipelines before writing the tensorlake cli. 

The .copy() calls are already rendered as COPY for both local and remote build contexts.

server/data_model/src/lib.rs
```diff
@@ -105,23 +105,29 @@ pub struct ImageInformation {
...
         base_image: String,
         run_strs: Vec<String>,
         sdk_version: Option<String>,
     ) -> Self {
-        let mut image_hasher = Sha256::new();
-        image_hasher.update(image_name.clone());
-        image_hasher.update(base_image.clone());
-        image_hasher.update(run_strs.clone().join(""));
-        image_hasher.update(sdk_version.clone().unwrap_or("".to_string())); // Igh.....
+        let mut compat_image_hash: String = "".to_string();
```
j3m7, line 114: @seriousben as discussed we preserve backwards compat while accepting blank image_hashes.

seriousben, line 114: You need to set this to the incoming image hash I believe.
```suggestion
        let mut compat_image_hash: String = image_hash;
```

python-sdk/indexify/functions_sdk/image.py
```diff
@@ -38,24 +106,90 @@ def base_image(self, base_image):
...
+            tarinfo = tarfile.TarInfo("Dockerfile")
+            tarinfo.size = len(dockerfile)
+
+            tf.addfile(tarinfo, BytesIO(dockerfile.encode()))
+
+    def _generate_dockerfile(self, python_sdk_path: Optional[str] = None):
+        docker_contents = [
+            f"FROM {self._base_image}",
+            "RUN mkdir -p ~/.indexify",
+            "RUN touch ~/.indexify/image_name",
```
seriousben, line 141: Looks like this is not needed based on the image_hash file working.
```suggestion
```

python-sdk/indexify/cli.py
```diff
@@ -266,72 +288,82 @@ def function_executor(
...
+        case "completed":
+            print(f"image {build.image_name}:{build.image_hash} is already built")
+        case "ready" | "building":
+            print(f"waiting for {build.image_name} image to build")
+            while build.status != "completed":
+                res = client.get(
+                    f"{service_endpoint}/builds/{build.id}", headers=headers
+                )
+                build = Build.model_validate(res.json())
+                time.sleep(5)
```
seriousben, line 338: Should the sleep not be reversed so that we sleep before getting the build by ID?

python-sdk/indexify/cli.py
```diff
@@ -266,72 +288,82 @@ def function_executor(
...
+    match build.status:
+        case "completed":
+            print(f"image {build.image_name}:{build.image_hash} is already built")
+        case "ready" | "building":
+            print(f"waiting for {build.image_name} image to build")
+            while build.status != "completed":
+                res = client.get(
+                    f"{service_endpoint}/builds/{build.id}", headers=headers
+                )
+                build = Build.model_validate(res.json())
```
seriousben, line 337: Is this where we'll be able to get the logs and display them?

python-sdk/indexify/cli.py
```diff
@@ -266,72 +288,82 @@ def function_executor(
...
+    match build.result:
+        case "success":
+            build_duration = build.push_completed_at - build.started_at
+            print(
+                f"Building completed in {build_duration}; image is stored in {build.uri}"
+            )
 
-    import docker.api.build
+        case "failed":
+            print(f"Building failed, please see logs for details")
```
seriousben, line 351: What is the current expectation when this fails? Is there a way to get the logs?

seriousben (approved):
LGTM, would be useful to know what you are planning to do for logs in the perspective that document ai image changes introduce an issue.

python-sdk/indexify/cli.py
j3m7: After the build is completed another call will need to be made to retrieve the logs, that is not implemented at this time.

python-sdk/indexify/cli.py
j3m7: Agreed. changing waiting before the GET

python-sdk/indexify/cli.py
j3m7: Not for the client side, they are logged by the builder task. A subsequent feature would be to store them in the blob store and add some API tissue to retrieve them.

PR #1146, opened 2025-01-05
Function allowlist for Executor and manual Graph management

diptanu:
Making the executors only run functions of specified compute graphs.
The functions are specified as
`indexify-cli executor --function <namespace>:<workflow>:<compute_func>:<version> --function ...`.

The allowlist is used with Executors in production setups. E.g. only run
a certain function in a Kubernetes POD with an Executor.
This is a default production setup because each function has different
resource usage and it’s easier to size containers per function.
Same with other container configs like secrets, roles, volumes.

`indexify-cli executor --dev` mode is still present. It makes server try
to run any function on the executor. This is convenient for development
of Indexify and for integration testing.

The new --function argument requires explicit control of function versions by the user.
Change Graph version type from u32 to string and allow user to set it in SDK.
This allows users to implement versioning of their graphs with any semantic they want.
Use a random uuid by default for the graph versions so if a user doesn't want to manage
versions manually we just always update the graph on each RemoteGraph.deploy().

Co-authored-by: Diptanu Gon Choudhury <diptanuc@gmail.com>
Co-authored-by: Eugene Batalov <eugene@tensorlake.ai>


python-sdk/indexify/executor/executor.py
```diff
@@ -24,11 +24,14 @@ def __init__(
         self,
         executor_id: str,
         code_path: Path,
+        compute_graph: str,
```
eabatalov, line 27: This needs to be optional.

python-sdk/indexify/executor/task_fetcher.py
```diff
@@ -28,11 +33,19 @@ def __init__(
         self.config_path = config_path
         self._logger = structlog.get_logger(module=__name__)
 
+        hostname = socket.gethostname()
+        ip_address = socket.gethostbyname(hostname)
```
eabatalov, line 37: This looks error prone as we don't guarantee that this IP address is reachable by server if we obtain it this way. It's better to pass it in CLI args into Executor. Because this depends on deployment environment. I'm also not sure why we need to pass this IP address to the Server because so far Server wasn't calling Executor but Executor was calling Server which is guaranteed to have a hostname reachable by Executors.

python-sdk/indexify/executor/task_fetcher.py
```diff
@@ -28,11 +33,19 @@ def __init__(
...
+
+        functions = []
+        if namespace is not None and compute_graph is not None and function is not None and graph_version is not None:
```
eabatalov, line 40: tbh I expected all these fields to be optional and passed directly to Server and it does filtering on its side with whatever provided values. E.g. if only namespace and compute_graph are set then server sees this in the ExecutorMetadata and then routes tasks for the namespace and the compute_graph to this Executor.

server/src/http_objects.rs
```diff
@@ -525,11 +524,21 @@ pub struct InvocationId {
...
+    // TODO Change this ot string
+    pub version: u32,
+}
+
 #[derive(Debug, Serialize, Deserialize, ToSchema)]
 pub struct ExecutorMetadata {
     pub id: String,
     pub executor_version: String,
     pub addr: String,
+    pub functions: Vec<FunctionContainer>,
```
eabatalov, line 541: These are not instantiated FunctionContainers/FunctionExecutors. Because of this I'd call these things as Executor filtering labels. Server will need to eventually track FunctionContainers/FunctionExecutors that really exist. Not mixing the filtering labels and existing FunctionContainers/FunctionExecutors would be nice imo.

executor/src/executor/task_fetcher.py
```diff
@@ -27,11 +33,31 @@ def __init__(
         self.config_path = config_path
         self._logger = structlog.get_logger(module=__name__)
 
+        hostname = socket.gethostname()
+        ip_address = socket.gethostbyname(hostname)
```
eabatalov, line 37: These lines also result in requiring to run Executor under sudo, fyi.

seriousben (approved):

PR #1148, opened 2025-01-07
Implicit namespace create

j3m7:
## Context

Simplify namespace creation to support serverless deployments.

## What

Adds an implicit namespace creation to the namespace middleware, the desired effect namespaces are dynamically created as functions are deployed.

Since namespaces are dynamically created we don't need a static definition of the default namespace so that has been removed.

## Testing

Start the server with an empty store and deploy any workspace.

## Contribution Checklist

### DIFF
--- a/python-sdk/src/indexify/functions_sdk/image.py
+++ b/python-sdk/src/indexify/functions_sdk/image.py
@@ -96,7 +96,7 @@ class Image:
         self._python_version = LOCAL_PYTHON_VERSION
         self._build_ops = []  # List of ImageOperation
         self._sdk_version = importlib.metadata.version("indexify-python-sdk")
-        self.uri = "" # For internal use
+        self.uri = ""  # For internal use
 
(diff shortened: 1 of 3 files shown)


python-sdk/src/indexify/functions_sdk/image.py
```diff
@@ -96,6 +96,7 @@ def __init__(self):
         self._python_version = LOCAL_PYTHON_VERSION
         self._build_ops = []  # List of ImageOperation
         self._sdk_version = importlib.metadata.version("indexify-python-sdk")
+        self.uri = ""  # For internal use
```
seriousben, line 99: was this meant to be included?

j3m7: Yes, piggy backing of this PR. Will be necessary for deploying.

seriousben (approved):
LGTM once linting passes

PR #1159, opened 2025-01-13
Migrate to tensorlake sdk

j3m7:
## Context

Use tensorlake SDK instead of indexify-python-sdk.

## Contribution Checklist


examples/contextual_rag/workflow.py
```diff
@@ -13,9 +13,9 @@
 from sentence_transformers import SentenceTransformer
 from pydantic import BaseModel
 
-from indexify.functions_sdk.graph import Graph
-from indexify.functions_sdk.image import Image
-from indexify.functions_sdk.indexify_functions import indexify_function, IndexifyFunction
+from tensorlake.functions_sdk.graph import Graph
+from tensorlake.functions_sdk.image import Image
+from tensorlake.functions_sdk.functions import tensorlake_function, TensorlakeCompute
```
eabatalov, line 18: Unrelated to this PR: these two terms "tensorlake_function" and "TensorlakeCompute" are not consistent while they are seemingly representing the same thing.

executor/src/executor/downloader.py
```diff
@@ -5,7 +5,7 @@
 import httpx
 import structlog
 from function_executor.proto.function_executor_pb2 import SerializedObject
-from python_utils.http_client import get_httpx_client
+from tensorlake.utils.http_client import get_httpx_client
```
eabatalov, line 8: Normally Executor shouldn't depend on Tensorlake SDK. Only Function Executor can do this.

function_executor/src/function_executor/handlers/run_function/handler.py
```diff
@@ -114,17 +113,17 @@ def _flush_logs(self) -> None:
...
 
-def _is_router(func_wrapper: IndexifyFunctionWrapper) -> bool:
+def _is_router(func_wrapper: TensorlakeFunctionWrapper) -> bool:
     """Determines if the function is a router.
 
     A function is a router if it is an instance of IndexifyRouter or if it is an IndexifyRouter class.
     """
     return str(
         type(func_wrapper.indexify_function)
     ) == "<class 'indexify.functions_sdk.indexify_functions.IndexifyRouter'>" or isinstance(
```
eabatalov, line 123: This string has to be updated.

function_executor/src/function_executor/main.py
```diff
@@ -1,4 +1,7 @@
-from python_utils.logging import configure_logging_early, configure_production_logging
+from tensorlake.utils.logging import (
```
eabatalov, line 1: Logging shouldn't be in Tensorlake SDK as it's logging functionality only used for services/daemons and Tensorlake SDK doesn't contain any services or daemons. So logging utils need to stay in Indexify repo.

function_executor/src/function_executor/handlers/run_function/handler.py
```diff
@@ -114,17 +113,17 @@ def _flush_logs(self) -> None:
...
-        func_wrapper.indexify_function, IndexifyRouter
+        func_wrapper.indexify_function, TensorlakeCompute
```
eabatalov, line 124: This should be Router not Compute (aka function)

PR #1164, opened 2025-01-15
Add tensorlake as submodule.

j3m7:
## Context
In order to do a full test before releasing tensorlake we need to have it available as a dev dependency. This is accomplished by adding tensorlake as a submodule. 

## What
- Adds tensorlake as a git submodule and directs all of the packages to use it a as a local dependency. 
- Swap dependency in packages
- Switch to python 3.9 for testing.
- Updated to use actions/checkout@v4 across all of CI
- In addition to testing indexify we also run tensorlake tests.

## Contribution Checklist


seriousben (approved):

eabatalov (approved):
LGTM

PR #1165, opened 2025-01-15
finalize and update graph ctx task analytics in the processor

seriousben:
## Context

Since graph invocation context influences task creation and that task creation happens in the processor, it enabled timing issues around parallel tasks where task creation could act on a previous task finalization event with data from a new task finalization which adds more complexity to the task creation logic and has added timing issues in the past.

## What

All dependencies of a processor should be mutated by it to remove any possibly of race conditions. This PR ensures that tasks will only get completed and its analytics updated by the processor.

## Testing

- from tensorlake: `TEST_TIMEOUT=2m TEST_MAX=100 TEST_FAIL_FAST=true INDEXIFY_URL=http://localhost:8900 command_stress_test poetry run python src/test_graph_behaviours.py`
- map reduce graph: `TEST_MAX=100 TEST_FAIL_FAST=true INDEXIFY_URL=http://localhost:8900 command_stress_test poetry run python graph.py`

## Contribution Checklist

### DIFF
--- a/server/data_model/src/lib.rs
+++ b/server/data_model/src/lib.rs
@@ -16,7 +16,6 @@ use filter::LabelsFilter;
 use indexify_utils::{default_creation_time, get_epoch_time_in_ms};
 use serde::{Deserialize, Serialize};
 use sha2::{Digest, Sha256};
-use strum::{AsRefStr, Display};
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct StateMachineMetadata {
@@ -377,7 +376,7 @@ impl ComputeGraph {
     }
 }
 
-#[derive(Debug, Display, PartialEq)]
+#[derive(Debug, strum::Display, PartialEq)]
 pub enum ComputeGraphError {
     VersionExists,
 }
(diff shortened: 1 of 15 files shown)


server/data_model/src/lib.rs
```diff
@@ -922,6 +934,18 @@ impl fmt::Display for TaskFinishedEvent {
     }
 }
 
+#[derive(Clone, Serialize, Deserialize, Debug, PartialEq)]
+pub struct TaskFinalizingEvent {
```
diptanu, line 938: Can we call it TaskOutputIngestedEvent or something to indicate clearly that this is sent out after we receive the task outputs, then the graph processor can do whatever it needs to.

server/state_store/src/state_changes.rs
```diff
@@ -97,7 +99,34 @@ pub fn tombstone_invocation(
     Ok(vec![state_change])
 }
 
-pub fn finalize_task(
+pub fn finalizing_task(
```
diptanu, line 102: pub fn set_task_outputs

diptanu (approved):
Some minor nits, looks good to me.

server/src/integration_test.rs
```diff
@@ -45,662 +44,357 @@ mod tests {
...
-        Ok(())
-    }
+    use crate::{
+        service::Service,
+        testing::{self, FinalizeTaskArgs},
+    };
 
     #[tokio::test]
-    async fn test_task_unassign() -> Result<()> {
+    async fn test_simple_graph_completion() -> Result<()> {
```
seriousben, line 53: Changing tests here since a lot of them relied on not having an executor which means the finalized task was still unallocated.

server/data_model/src/lib.rs
seriousben: done

server/state_store/src/state_changes.rs
seriousben: done

server/data_model/src/lib.rs
seriousben: went with TaskOutputsIngested

server/state_store/src/state_changes.rs
seriousben: went with task_outputs_ingested

PR #1173, opened 2025-01-17
Setup PDF Document Parsing example with indexify 0.3.1

stangirala:
This reverts commit e2f9776db50cacc56092d83f5c14e1086194ce4d.

## Context
Setup example to use the new indexify cli commands.

## What
Fix the docker compose file to setup the args correctly.

## Testing
Started containers locally and tested. We would have to publish new images and I can retest once to make sure everything is fine.

## Contribution Checklist

### DIFF
--- a/examples/pdf_document_extraction/workflow.py
+++ b/examples/pdf_document_extraction/workflow.py
@@ -55,7 +55,7 @@ if __name__ == "__main__":
     file = File(data=resp.content, mime_type="application/pdf")
 
     # uncomment to run locally
-    #invocation_id = graph.run(block_until_true=True, file=file)
+    #invocation_id = graph.run(file=file)
     #exit(0)
 
     import common_objects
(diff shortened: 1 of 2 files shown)


examples/pdf_document_extraction/workflow.py
```diff
@@ -55,7 +55,7 @@ def create_graph() -> Graph:
     file = File(data=resp.content, mime_type="application/pdf")
```
eabatalov, line 55: Do we need to set version="0.1" in create_graph() call?

diptanu: @eabatalov I think we do, otherwise the user will keep wondering what to pass in as arguments to the executor. They will have to look up the UI, get the UUID (graph version) and set up the executors.

With a known graph version, we can write the instructions to start the executor cleanly on the CLI.

stangirala: This hardcode is just to show how to use the feature.

eabatalov: I have the same impression @diptanu. Adding the version removes a step from the readme and makes it more clear how the things are related to each other (where the version is set).

PR #1179, opened 2025-01-17
feat(server): add invocation wait endpoint

seriousben:
## Context

Currently the only way to wait for invocation to finish is on creation. If the server or client goes down, there is no way to resume waiting for the invocation to complete.

## What

This PR adds a wait invocation endpoint that returns invocation task events to allow tracking the progress of an invocation.

Additionally, it adds lagging handling required by broadcast channels so that we never miss an invocation finished event.

## Testing

Tested in https://github.com/tensorlakeai/tensorlake/pull/14

## Contribution Checklist

### DIFF
--- a/server/src/routes.rs
+++ b/server/src/routes.rs
@@ -53,7 +53,12 @@ use download::{
     download_invocation_payload,
 };
 use internal_ingest::ingest_files_from_executor;
-use invoke::{invoke_with_file, invoke_with_object, replay_compute_graph};
+use invoke::{
+    invoke_with_file,
+    invoke_with_object,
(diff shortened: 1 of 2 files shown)


eabatalov:
One overall comment is that it'd be nice to add some unit tests to this functionality if there's no existing unit tests with a good coverage.

eabatalov (approved):

seriousben:
> One overall comment is that it'd be nice to add some unit tests to this functionality if there's no existing unit tests with a good coverage.

Unfortunately, we don't have good unit tests around the server API yet. For now, it is something I would like to invest it soon.

PR #1187, opened 2025-01-24
feat: implement endpoint for retrieving invocation's details

miguelhrocha:
## Context

Currently, there is no good way to check whether an invocation exists on a graph. This is a need because in lower environments graphs remove invocations and dependent systems are left hanging without a way to verify whether the invocation is there.

## What

GET operation to retrieve an invocation from a graph.

### DIFF
--- a/server/src/routes.rs
+++ b/server/src/routes.rs
@@ -8,8 +8,7 @@ use axum::{
     middleware::{self, Next},
     response::{sse::Event, Html, IntoResponse},
     routing::{delete, get, post},
-    Json,
-    Router,
+    Json, Router,
 };
 use axum_tracing_opentelemetry::{
     self,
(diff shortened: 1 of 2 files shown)


server/src/routes.rs
```diff
@@ -804,6 +808,30 @@ async fn list_outputs(
...
+)]
+async fn find_invocation(
+    Path((namespace, compute_graph, invocation_id)): Path<(String, String, String)>,
+    State(state): State<RouteState>,
+) -> Result<Json<Invocation>, IndexifyAPIError> {
+    let invocation_ctx = state
+        .indexify_state
+        .reader()
+        .invocation_ctx(&namespace, &compute_graph, &invocation_id)
+        .map_err(IndexifyAPIError::internal_error)?
```
eabatalov, line 829: It seems to me like we might turn any error in `.invocation_ctx(&namespace, &compute_graph, &invocation_id)` call here into internal error but then in the next ok_or call turning any error into not found.

Could you please double check that this is not the case?

miguelhrocha: `ok_or` is a method from a Rust `Option` 😄 `map_err` is a method from a Rust `Result`. The first `map_err` just asserts the data was fetched properly, and the `ok_or` asserts the data was there.

eabatalov: Nah, looks correct I guess. Cause ok_or is acting based on Some() or None. Not on the error.

PR #1201, opened 2025-01-30
Hack (don't merge unless really needed): Overwrite task version id in manual function executor management mode

eabatalov:
This is a hack that forces Function Executor to accept a task if the Function Executor was created for a different function version.


eabatalov:
This is not needed anymore.

PR #1218, opened 2025-02-12
Remove Executor image build that is not used in example workflows anymore

eabatalov:


eabatalov:
The image was removed in
https://github.com/tensorlakeai/indexify/commit/1f4debecbf2ab9ca4b2559404e959c8952aa4d9c

PR #1228, opened 2025-02-18
Move dependencie control to tensorlake.

j3m7:
## Context
Simplify dependency management by having the tensorlake package dictate dependencies. Tensorlake will always be a dependency and we should only specify indexify specific dependencies here. 

## What
Replace dependencies in pyproject with note that they are handled by tensorlake. Updated unpinned dependencies. 

## Contribution Checklist


eabatalov (approved):
lgtm

PR #1232, opened 2025-02-18
Remove py-inkwell package from examples

eabatalov:
This package is not available anymore and it's not needed in the examples.


stangirala:
+1

PR #1240, opened 2025-02-23
Tests for distribution of invocations between Executors by Server

eabatalov:
Testing a few typical scenarios that happen in production.

The tests depend on timings quite a lot. We'd delete them if they end up being flaky.
It's better to rewrite them in the future as deterministic server unit tests.

The test fails currently due to the known issue:

```
======================================================================
FAIL: test_server_redistributes_invocations_when_new_executor_joins (__main__.TestServer)
  File "/home/runner/work/indexify/indexify/indexify/tests/./cli/test_server.py", line 140, in test_server_redistributes_invocations_when_new_executor_joins
    self.assertLess(invocations_count, 150)
AssertionError: 200 not less than 150
Ran 3 tests in 127.851s
FAILED (failures=1)
```

### DIFF
--- a/indexify/tests/cli/test_disable_automatic_container_management.py
+++ b/indexify/tests/cli/test_disable_automatic_container_management.py
@@ -1,21 +1,15 @@
-import os
-import platform
 import subprocess
 import unittest
 
+import testing
 from tensorlake import Graph, tensorlake_function
 from tensorlake.remote_graph import RemoteGraph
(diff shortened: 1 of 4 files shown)


eabatalov:
The bug this PR is exposing is going to be fixed eventually and then we'll merge it.

seriousben:
Awesome!

PR #1249, opened 2025-02-26
migration adding status to tasks

seriousben:
## Context

We recently added a status (pending, running, completed) to tasks for display purposes. We have since started to use it to differentiate allocated or running tasks.

## What

This PR populates the tasks status so that it can be relied on.

- Set the task state to completed for task that are failed or successful.
- Set the task state to pending otherwise.
- Also fixing a stack overflow by using Box to allocate data on the heap.

## Testing

- [x] Use v0.2.32 and queued up a big reduce task without a live executor.
- [x] Those tasks are shown as pending after the migration on the UI.
- [x] Tasks that were Failed or Success are shown as completed on the UI.
- [x] Starting an executor starts processing all tasks successfully.
- [x] Show the migration logs
	```
	{"timestamp":"2025-02-26T16:36:12.139116Z","level":"INFO","message":"Migrated 2869 tasks from v1 to v2","target":"state_store::migrations","span":{"sm_meta":"StateMachineMetadata { db_version: 1, last_change_idx: 9143 }","name":"migrate_v1_to_v2"}}
	{"timestamp":"2025-02-26T16:36:12.202954Z","level":"INFO","message":"initialized state store with last state change id: 9143","target":"state_store"}
	{"timestamp":"2025-02-26T16:36:12.202980Z","level":"INFO","message":"db version discovered: 2","target":"state_store"}
	```

## Contribution Checklist

### DIFF
--- a/server/processor/src/graph_processor.rs
+++ b/server/processor/src/graph_processor.rs
@@ -170,9 +170,15 @@ impl GraphProcessor {
                     .invoke(&state_change.change_type, &mut indexes)
                     .await;
                 if let Ok(mut result) = scheduler_update {
-                    let placement_result = self
-                        .task_allocator
-                        .schedule_tasks(result.clone().updated_tasks, &mut indexes)?;
+                    let placement_result = self.task_allocator.schedule_tasks(
+                        result
+                            .clone()
+                            .updated_tasks
+                            .into_iter()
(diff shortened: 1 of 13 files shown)


diptanu:
Lets change the aspects of creating Allocations - Make them Boxed from beginning. Ship it.

PR #1254, opened 2025-02-27
gRPC definitions for new Server <-> Executor protocol

diptanu:
The new internal gRPC Task Scheduler service allows Server to
have full control over Function Executors lifecycle on Executors
and have full control about which task is running on each
Function Executor.

Depricated HTTP API is used for task inputs download and for
reporting task outcome. The existing task outcome reporting HTTP
handler can be used as is side by side with the new gRPC API.
There's no value in refactoring this handler right now.
We'll move task outcome reporting to gRPC when we move S3 uploads
to Executor.

### DIFF
--- a/server/internal_api/proto/task_scheduler.proto
+++ b/server/internal_api/proto/task_scheduler.proto
@@ -1,0 +1,144 @@
+syntax = "proto3";
+
+package task_scheduler_service;
+
+// ===== ReportExecutorState RPC =====
+
+enum GPUModel {
+    GPU_MODEL_UNKNOWN = 0;
+    GPU_MODEL_NVIDIA_TESLA_T4_16GB = 10;
+    GPU_MODEL_NVIDIA_TESLA_V100_16GB = 20;
+    GPU_MODEL_NVIDIA_A10_24GB = 30;
+    GPU_MODEL_NVIDIA_A6000_48GB = 40;
+    // A100 GPUs
+    GPU_MODEL_NVIDIA_A100_SXM4_40GB = 50;
+    GPU_MODEL_NVIDIA_A100_SXM4_80GB = 51;
+    GPU_MODEL_NVIDIA_A100_PCI_40GB = 52;
+    // H100 GPUs
+    GPU_MODEL_NVIDIA_H100_SXM5_80GB = 60;
+    GPU_MODEL_NVIDIA_H100_PCI_80GB = 61;
+    GPU_MODEL_NVIDIA_RTX_6000_24GB = 62;
+}
+
+// Free GPUs available at the Executor.
+message GPUResources {
+    optional uint32 count = 1;
+    optional GPUModel model = 2;
+}
+
+// Free host resources available at the Executor.
+message HostResources {
+    optional uint32 cpu_count = 1;
+    optional uint64 memory_bytes = 2;
+    optional uint64 disk_bytes = 3;
+    optional GPUResources gpu = 4;
+}
+
+// Specification of a single function that is allowed to be run on the Executor.
+message AllowedFunction {
+    optional string namespace = 1;
+    optional string graph_name = 2;
+    optional string function_name = 3;
+    // If none then any version of the graph is allowed to run on the Executor.
+    optional string graph_version = 4;
+}
+
+enum FunctionExecutorStatus {
+    FUNCTION_EXECUTOR_STATUS_UNKNOWN = 0;
+    FUNCTION_EXECUTOR_STATUS_STARTING = 1;
+    FUNCTION_EXECUTOR_STATUS_IDLE = 2;
+    FUNCTION_EXECUTOR_STATUS_RUNNING_TASK = 3;
+    FUNCTION_EXECUTOR_STATUS_SHUTTING_DOWN = 4;
+    FUNCTION_EXECUTOR_STATUS_UNHEALTHY = 5;
+}
+
+// Immutable information that identifies and describes a Function Executor.
+message FunctionExecutorDescription {
+    optional string id = 1;
+    optional string namespace = 2;
+    optional string graph_name = 3;
+    optional string graph_version = 4;
+    optional string function_name = 5;
+}
+
+message FunctionExecutorState {
+    optional FunctionExecutorDescription description = 1;
+    optional FunctionExecutorStatus status = 2;
+}
+
+enum ExecutorStatus {
+    EXECUTOR_STATUS_UNKNOWN = 0;
+    EXECUTOR_STATUS_STARTING = 1;
+    EXECUTOR_STATUS_RUNNING = 2;
+    EXECUTOR_STATUS_DRAINED = 3;
+    EXECUTOR_STATUS_SHUTTING_DOWN = 4;
+}
+
+message ExecutorState {
+    optional string executor_id = 1;
+    optional ExecutorStatus executor_status = 2;
+    optional HostResources host_resources = 3;
+    // Empty allowed_functions list means that any function can run on the Executor.
+    repeated AllowedFunction allowed_functions = 4;
+    repeated FunctionExecutorState function_executor_states = 5;
+}
+
+// A message sent by Executor to report its up to date state to Server.
+message ReportExecutorStateRequest {
+    optional ExecutorState executor_state = 1;
+}
+
+// A message sent by Server to Executor to acknowledge the receipt of Executor state.
+message ReportExecutorStateResponse {
+}
+
+// ===== GetDesiredExecutorStates RPC =====
+message Task {
+    optional string id = 1;
+    optional string namespace = 2;
+    optional string graph_name = 3;
+    optional string graph_version = 4;
+    optional string function_name = 5;
+    optional string graph_invocation_id = 6;
+    optional string input_key = 8;
+    optional string reducer_output_key = 9;
+    optional string image_uri = 10;
+}
+
+message TaskAllocation {
+    optional string function_executor_id = 1;
+    optional Task task = 2;
+}
+
+// A message sent by Executor to Server to open the stream of desired Executor States for the Executor.
+message GetDesiredExecutorStatesRequest {
+    optional string executor_id = 1;
+}
+
+// A message sent from Server to Executor that describes the desired state of the Executor at the moment.
+// Executor compares this state with its current state and make necessary changes to match the desired state.
+message DesiredExecutorState {
+    repeated FunctionExecutorDescription function_executors = 1;
+    repeated TaskAllocation task_allocations = 2;
+    // Server supplied clock value used to deduplicate messages. Executor records max clock value
+    // it observed and ignores all the messages with clock value <= the max observed value.
+    optional uint64 clock = 3;
+}
+
+// Internal API for scheduling and running tasks on Executors. Executors are acting as clients of this API.
+// Server is responsible for scheduling tasks on Executors and Executors are responsible for running the tasks.
+service TaskSchedulerService {
+    // Called by Executor every 5 seconds to report that it's still alive and provide its current state.
+    //
+    // Missing 3 reports will result in the Executor being deregistered by Server.
+    rpc ReportExecutorState(ReportExecutorStateRequest) returns (ReportExecutorStateResponse) {}
+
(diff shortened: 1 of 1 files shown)


protos/internal.proto
```diff
@@ -0,0 +1,114 @@
...
+message GPUResource {
+    uint32 gpu_count = 1;
+    string gpu_model = 2;
+}
+
+message HostResources {
+    uint32 cpu_count = 1;
+    uint64 memory_bytes = 2;
+    uint64 disk_bytes = 3;
+    GPUResource gpu_resource = 4;
```
eabatalov, line 12: Nit. gpu is enough here (without _resource suffix)

protos/internal.proto
```diff
@@ -0,0 +1,114 @@
...
+}
+
+message AllowedFunction {
+    string namespace = 1;
+    string compute_graph = 2;
+    string compute_graph_version = 3;
+    string function = 4;
```
eabatalov, line 19: I prefere function_name. It's more clear.

protos/internal.proto
```diff
@@ -0,0 +1,114 @@
...
+}
+
+message HeartbeatRequest {
+    string executor_id = 1;
+    uint64 timestamp = 2;
```
eabatalov, line 24: Let's clarify the timestamp format in a comment (like e.g. ms since epoch? timezone?)
Also how is it going to be used on Server? Like if it's a too old time stamp then it'll discard the message?

protos/internal.proto
```diff
@@ -0,0 +1,114 @@
...
+    uint32 cpu_count = 1;
+    uint64 memory_bytes = 2;
+    uint64 disk_bytes = 3;
+    GPUResource gpu_resource = 4;
+}
+
+message AllowedFunction {
+    string namespace = 1;
+    string compute_graph = 2;
+    string compute_graph_version = 3;
```
eabatalov, line 18: compute_graph_version is optional right?

protos/internal.proto
```diff
@@ -0,0 +1,114 @@
...
+    string function = 4;
+}
+
+message HeartbeatRequest {
+    string executor_id = 1;
+    uint64 timestamp = 2;
+    HostResources host_resources = 3;
+    repeated AllowedFunction allowed_functions = 4;
+    uint32 readiness = 5; // 0 -> not ready, 1 -> ready, 2 -> drain
```
eabatalov, line 27: Nice field, should we change this to `status` and use an enum for the values?

protos/internal.proto
```diff
@@ -0,0 +1,114 @@
...
+}
+
+message HeartbeatResponse {
+}
+
+message GetFunctionExecutorsRequest {
+    repeated string executor_ids = 1;
```
eabatalov, line 34: function_executor_ids

protos/internal.proto
```diff
@@ -0,0 +1,114 @@
...
+
+/// The executor will send this to the server to receive new task allocations.
+message GetTaskAllocationsRequest {
+    string executor_id = 1;
+}
+
+message TaskAllocation {
+    string namespace = 1;
+    string compute_graph = 2;
+    string compute_fn_name = 3;
```
eabatalov, line 86: The field names need to be consistent across the messages, e.g. see AllowedFunction

protos/internal.proto
```diff
@@ -0,0 +1,114 @@
...
+message TaskOutcome {
+    string namespace = 1;
+    string compute_graph = 2;
+    string compute_fn_name = 3;
+    string invocation_id = 4;
+    string task_id = 5;
+    uint64 outcome = 6; // 0 -> success, 1 -> failure, 2 -> timeout, 3 -> cancelled
+    uint64 received_at = 7;
+    uint64 started_at = 8;
+    uint64 completed_at = 9;
```
eabatalov, line 66: How are we going to provide task files: https://github.com/tensorlakeai/indexify/blob/main/indexify/src/indexify/executor/task_reporter.py#L86
(stdout, stderr, outputs)

protos/internal.proto
```diff
@@ -0,0 +1,114 @@
...
+    string function_executor_id = 6;
+}
+
+/// this will be sent to the executor at any point in time.
+/// this will have the list of functions that should be running on this executor.
+/// We are going to add new executors here, and keep them until they need to be removed.
+/// The executors need to diff the list of functions they are running and the list of functions they should be running.
+/// If they are not running the functions they should be running, they need to start running them.
+/// If they are running the functions they should not be running, they need to stop running them.
+/// If they are running the functions they should be running, they need to keep running them.
```
eabatalov, line 52: This block of comments doesn't look related to GetFunctionExecutorsResponse.
I'd move it to the RPC it's about

protos/internal.proto
eabatalov: You decided to not the content of GetFunctionExecutorsRequest into HeartbeatRequest, right? You have a scenario where server calls GetFunctionExecutors outside of the heartbeat exchange?

protos/internal.proto
```diff
@@ -0,0 +1,114 @@
...
+    string input_node_output_key = 7;
+    string reducer_output_id = 8;
+    uint64 clock = 9;
+}
+
+message GetTaskAllocationsResponse {
+    repeated TaskAllocation task_allocations = 1;
+}
+
+service HeartbeatService {
```
eabatalov, line 99: This service is Server, Executor is the client.

protos/internal.proto
```diff
@@ -0,0 +1,114 @@
...
+    /// Heartbeat is used to send a heartbeat to the server.
+    /// The server will use this to check if the executor is still alive. Missing 3 heartbeats will result in the executor being deregistered.
+    rpc Heartbeat(stream HeartbeatRequest) returns (stream HeartbeatResponse) {}
```
eabatalov, line 102: This is a simple request-response RPC. I don't see a need for more complex streaming here.

protos/internal.proto
```diff
@@ -0,0 +1,114 @@
...
+
+    /// GetFunctionExecutors is used to get the function executors for a given function.
```
eabatalov, line 104: Is the comment accurate?
It looks like an RPC from server to Executor to get state of Function Executors. Is it?
In this case it needs to be a bidirectional stream called FunctionExecutors with GetFunctionExecutorsRequest in returns part of it.

protos/internal.proto
eabatalov: Or is it executor_id?

protos/internal.proto
```diff
@@ -0,0 +1,114 @@
...
+    repeated string executor_ids = 1;
+}
+
+message FunctionExecutor {
+    string namespace = 1;
+    string compute_graph = 2;
+    string compute_graph_version = 3;
+    string function = 4;
+    string executor_id = 5;
+    string function_executor_id = 6;
```
eabatalov, line 43: How Graph is going to get downloaded and deserialized?

protos/internal.proto
```diff
@@ -0,0 +1,114 @@
...
+message TaskAllocation {
+    string namespace = 1;
+    string compute_graph = 2;
+    string compute_fn_name = 3;
+    string compute_graph_version = 4;
+    string invocation_id = 5;
+    string task_id = 6;
+    string input_node_output_key = 7;
+    string reducer_output_id = 8;
+    uint64 clock = 9;
```
eabatalov, line 92: How are we going to download and deserialize task inputs and reducer output id?
See example in function executor proto: https://github.com/tensorlakeai/tensorlake/blob/main/src/tensorlake/function_executor/proto/function_executor.proto#L29

protos/internal.proto
```diff
@@ -0,0 +1,114 @@
...
+    repeated TaskAllocation task_allocations = 1;
+}
+
+service HeartbeatService {
+    /// Heartbeat is used to send a heartbeat to the server.
+    /// The server will use this to check if the executor is still alive. Missing 3 heartbeats will result in the executor being deregistered.
+    rpc Heartbeat(stream HeartbeatRequest) returns (stream HeartbeatResponse) {}
+
+    /// GetFunctionExecutors is used to get the function executors for a given function.
+    rpc GetFunctionExecutors(GetFunctionExecutorsRequest) returns (stream GetFunctionExecutorsResponse) {}
```
eabatalov, line 105: I'm a bit confused about this RPC, let's discuss offline

protos/internal.proto
```diff
@@ -0,0 +1,114 @@
...
+}
+
+message FunctionExecutor {
+    string namespace = 1;
+    string compute_graph = 2;
+    string compute_graph_version = 3;
+    string function = 4;
+    string executor_id = 5;
+    string function_executor_id = 6;
+}
```
eabatalov, line 44: This needs a status field with enum:

- Idle
- Unhealthy
- Starting
- Destroying

protos/internal.proto
```diff
@@ -0,0 +1,114 @@
+syntax = "proto3";
+
+message GPUResource {
+    uint32 gpu_count = 1;
+    string gpu_model = 2;
```
eabatalov, line 5: What if we use an enum for these. GPU model string format not standardized, the enums will force us to have a common standard between Server and Executor.

protos/internal.proto
eabatalov: This RPC produces for Executor a stream of Function Executors it should be running at the moment,

protos/internal.proto
```diff
@@ -0,0 +1,114 @@
...
+}
+
+message AllowedFunction {
+    string namespace = 1;
+    string compute_graph = 2;
+    string compute_graph_version = 3;
+    string function = 4;
+}
+
+message HeartbeatRequest {
```
eabatalov, line 22: offline discussion:
+ function_executors repeated field that lists all the existing Function Executors on the Executor with the current status of each Function Executor.

protos/internal.proto
```diff
@@ -0,0 +1,114 @@
...
+message ReportTaskOutcomesResponse {
+}
+
+
+/// The executor will send this to the server to receive new task allocations.
+message GetTaskAllocationsRequest {
+    string executor_id = 1;
+}
+
+message TaskAllocation {
```
eabatalov, line 83: + function_executor_id

protos/internal.proto
eabatalov: offline discussion:
This is a server supplied clock value that Executor will use to deduplicate task allocation responses. Executor records latest clock it got from server in TaskAllocation responses and ignore all TaskAllocation responses with clock value less than or equal to the one recorded by Executor.

protos/internal.proto
```diff
@@ -0,0 +1,114 @@
...
+    rpc Heartbeat(stream HeartbeatRequest) returns (stream HeartbeatResponse) {}
+
+    /// GetFunctionExecutors is used to get the function executors for a given function.
+    rpc GetFunctionExecutors(GetFunctionExecutorsRequest) returns (stream GetFunctionExecutorsResponse) {}
+
+    /// GetTaskAllocations is used to get the task allocations for a given executor.
+    rpc GetTaskAllocations(GetTaskAllocationsRequest) returns (stream GetTaskAllocationsResponse) {}
+
+    /// ReportTaskOutcomes is used to report the task outcomes for a given executor.
+    rpc ReportTaskOutcomes(ReportTaskOutcomesRequest) returns (ReportTaskOutcomesResponse) {}
```
eabatalov, line 111: If this RPC fails then we need to keep retrying until the server successfully processes the request.

protos/internal.proto
eabatalov: offline discussion:
We don't really need this timestamp in HealthCheck request.

protos/internal.proto
```diff
@@ -0,0 +1,114 @@
...
+    repeated FunctionExecutor function_executors = 1;
+}
+
+message TaskOutcome {
+    string namespace = 1;
+    string compute_graph = 2;
+    string compute_fn_name = 3;
+    string invocation_id = 4;
+    string task_id = 5;
+    uint64 outcome = 6; // 0 -> success, 1 -> failure, 2 -> timeout, 3 -> cancelled
```
seriousben, line 63: Should be an enum here.

eabatalov: Working on it, yep

PR #1256, opened 2025-02-27
feat: limit allocations per executor to spread load

seriousben:
## Context

On a busy server, adding new executors would have very little to no impact on the P90 of the tasks's time-to-executor latency. This is because it only start being used to allocate new tasks but would not be used to place already allocated tasks. All the previous tasks would have already been allocated to the executor already present on the system without any limits.

## What

This PR limits the number of tasks that are allocated per executor at a time in order for the server to be able to take advantage of new executors joining the cluster.

The current limit is set to 20 which is double the number of tasks we queue (send via the task stream) on an executor at once.

## Testing

- Start server
- Queue up 30 invocations of many 1000 reducer graph
- Start executor 1
- Start executor 2
- Open /internal/allocations
- Notice allocations are spread very evenlly

## Contribution Checklist

### DIFF
--- a/server/processor/src/graph_processor.rs
+++ b/server/processor/src/graph_processor.rs
@@ -105,9 +105,13 @@ impl GraphProcessor {
             return Ok(());
         }
 
-        // 3. Fire a notification when caching multiple
-        // state changes in order to process them.
-        if cached_state_changes.len() > 1 {
+        // 3. Fire a notification when handling multiple
+        // state changes in order to fill in the cache when
+        // the next state change is processed.
+        //
+        // 0 = we fetched all current state changes
+        // > 1 = we are processing cached state changes, we need to fetch more when done
+        if cached_state_changes.len() > 0 {
             notify.notify_one();
         }
 
--- a/server/processor/src/graph_processor.rs
+++ b/server/processor/src/graph_processor.rs
@@ -173,22 +173,18 @@ impl GraphProcessor {
                     .invoke(&state_change.change_type, &mut indexes)
                     .await;
                 if let Ok(mut result) = scheduler_update {
-                    let placement_result = self.task_allocator.schedule_tasks(
-                        result
-                            .clone()
-                            .updated_tasks
-                            .into_iter()
-                            .map(Box::new)
-                            .collect(),
-                        &mut indexes,
-                    )?;
+                    result.updated_tasks.iter().for_each(|t| {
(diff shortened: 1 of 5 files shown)


server/processor/src/graph_processor.rs
```diff
@@ -173,22 +173,18 @@ impl GraphProcessor {
...
-                    let placement_result = self.task_allocator.schedule_tasks(
-                        result
-                            .clone()
-                            .updated_tasks
-                            .into_iter()
-                            .map(Box::new)
-                            .collect(),
-                        &mut indexes,
-                    )?;
+                    result.updated_tasks.iter().for_each(|t| {
```
diptanu, line 176: Lets move these into where they are created, so we know the copy of the in-memory indexes are always updated from the decision maker. It will also make it easy to write tests down the line for task creator.

server/processor/src/graph_processor.rs
```diff
@@ -173,22 +173,18 @@ impl GraphProcessor {
...
+                        indexes.tasks.insert(t.key(), Box::new(t.clone()));
+                        indexes.unallocated_tasks.insert(t.key(), [0; 0]);
+                    });
+                    let placement_result = self.task_allocator.allocate(&mut indexes)?;
```
diptanu, line 180: I believe the task_allocator updates the in memroy index. The comment above also ensures we have the same style of updating state everywehre.

diptanu (commented):
Leaving some comemnts but they are mostly nit.

server/processor/src/graph_processor.rs
seriousben: I moved it within the task_creator but because of how big the creator is. I kept it just in the invoke function for now.

server/processor/src/task_allocator.rs
```diff
@@ -129,8 +145,29 @@ impl TaskAllocationProcessor {
...
-            match self.allocate_task(&task, indexes) {
+
+            // get executors with allocation capacity
+            let executors = indexes
+                .executors
+                .iter()
+                .filter(|(k, _)| {
+                    let allocations = indexes.allocations_by_executor.get(*k);
+                    let allocation_count = allocations.map_or(0, |allocs| allocs.len());
+                    info!("executor {} has {} allocations", k, allocation_count);
```
seriousben, line 158: I need to remove this noisy log

PR #1259, opened 2025-03-02
Secondary index for invocations

diptanu:
Added secondary index for invocations because when we ingest 10s of 1000s of files, the UI becomes unresponsive. 

The list_invocations api returns the results in descending order of created_at attribute to give us what we need for the UI. It needs to be more configurable 

There is a bug here. The cursor should be null if we at the last page and iteration direction is backward, and on the first page if the iteration direction is forward.


adithyaakrishna:
@diptanu Updated the UI for pagination, here's the recoding of it

[screen-capture (14).webm](https://github.com/user-attachments/assets/894898a3-d200-456e-8658-999f597dacec)

PR #1260, opened 2025-03-04
fix: fix unallocated task ordering

seriousben:
## Context

We rely on Task Key ordering when iterating on unallocated tasks which is `<namespace>_<compute_graph_name>_<invocation_id>_<fn_name>_<task_id>`

This means that we allocate tasks from the lowest namespace first. Within a given namespace, we would favor lower invocation id and lower task id as well.
Task IDs are uuid v4 is random and do not incorporate a timestamp like uuid v7 does. Which would not prioritize older tasks.

Fixes https://github.com/tensorlakeai/compute-engine-internal/issues/66

## What

This PR introduces an `UnallocatedTaskId` which allows defining an ordering for unallocated tasks. From the `UnallocatedTaskId`, it is possible to get the Task Key to retrieve a task.

Note that since unallocated tasks never get saved in rocksdb, we do not have to make it a key. It is only used in memory in an am::OrdSet.

Since tasks are indexed by TaskKey and that task key require the whole `<namespace>_<compute_graph_name>_<invocation_id>_<fn_name>_<task_id>` fields to be present since we do range queries, we did not change how task are indexed in memory.

This PR also adds new metrics:

```
# HELP task_queuing_latency_seconds Time tasks wait in queue before allocation in seconds
# TYPE task_queuing_latency_seconds histogram
...
# HELP processor_processing_latency_seconds processor task processing latency in seconds
# TYPE processor_processing_latency_seconds histogram
...
# HELP state_transition_latency_seconds Latency of state transitions before processing in seconds
# TYPE state_transition_latency_seconds histogram
```

## Testing

- [x] Testing triggering a lot of distributed map invocations and expecting invocations to finish one by one as opposed to have them finish almost at the same time.

## Contribution Checklist

### DIFF
--- a/server/state_store/src/in_memory_state.rs
+++ b/server/state_store/src/in_memory_state.rs
@@ -10,9 +10,11 @@ use data_model::{
     ReduceTask,
     Task,
     TaskStatus,
+    UnallocatedTaskId,
 };
 use metrics::{StateStoreMetrics, Timer};
 use opentelemetry::{metrics::Gauge, KeyValue};
+use tracing::error;
 
 use crate::{
     requests::{RequestPayload, StateMachineUpdateRequest},
@@ -36,7 +38,7 @@ pub struct InMemoryState {
     pub allocations_by_executor: im::HashMap<String, im::Vector<Box<Allocation>>>,
 
     // TaskKey -> Task
-    pub unallocated_tasks: im::OrdMap<String, [u8; 0]>,
+    pub unallocated_tasks: im::OrdSet<UnallocatedTaskId>,
 
     // Task Key -> Task
     pub tasks: im::OrdMap<String, Box<Task>>,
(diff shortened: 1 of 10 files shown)


server/state_store/src/in_memory_state.rs
```diff
@@ -395,6 +406,8 @@ impl InMemoryState {
                         .retain(|a| a.task_id != allocation.task_id);
                 }
                 for executor_id in &req.remove_executors {
+                    self.active_allocations_gauge
```
diptanu, line 409: It sucks that we can't remove recordings

diptanu (changes requested):
Some minor nits

server/state_store/src/in_memory_state.rs
```diff
@@ -344,9 +346,9 @@ impl InMemoryState {
             RequestPayload::SchedulerUpdate(req) => {
                 for task in &req.updated_tasks {
                     if task.status == TaskStatus::Pending {
-                        self.unallocated_tasks.insert(task.key(), [0; 0]);
+                        self.unallocated_tasks.insert(task.unallocated_task_id());
                     } else {
-                        self.unallocated_tasks.remove(&task.key());
+                        self.unallocated_tasks.remove(&task.unallocated_task_id());
```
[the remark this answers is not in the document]
seriousben: Good point. I addressed the feedback.

server/state_store/src/in_memory_state.rs
seriousben: Agreed, at least having it as zero will help have better data. Otherwise we fallback to observable gauge for that metric.

PR #1273, opened 2025-03-12
feat: expose secrets meatadata in executor

seriousben:
## Context

 
 We need to allow executors to inject secrets in function executors.

## What

This PR exposes secret names in the `FunctionExecutorServerConfiguration` so that executors can implement their secret strategy.

This is made possible by:
- https://github.com/tensorlakeai/indexify/pull/1271
- https://github.com/tensorlakeai/tensorlake/pull/63

## Contribution Checklist


indexify/src/indexify/executor/function_executor/single_task_runner.py
```diff
@@ -133,7 +133,9 @@ async def _create_function_executor(self) -> None:
             FunctionExecutorServerConfiguration(
                 executor_id=self._executor_id,
                 function_executor_id=self._function_executor_state.id,
+                namespace=self._task_input.task.namespace,
                 image_uri=self._task_input.task.image_uri,
+                secret_names=self._task_input.task.secret_names or [],
```
eabatalov, line 138: Very Nit. Could just keep it optional (same as image_uri).

eabatalov (approved):
Nice, shipit :)

indexify/src/indexify/executor/function_executor/single_task_runner.py
seriousben: I am switching it to not be optional in the configuration to simplify downstream code basically.

PR #1285, opened 2025-03-15
update executor api service

diptanu:
WIP. 

* Move executor api into the server crate.
* Wiring up heartbeats with in-memory state and persistent state
* moving old code path for registration  with this


server/data_model/src/lib.rs
```diff
@@ -1154,14 +1154,114 @@ pub struct FunctionURI {
...
+    pub cpu_count: u32,
+    pub memory_bytes: u64,
+    pub disk_bytes: u64,
+    pub gpu: Option<GpuResources>,
+}
+
+impl Default for HostResources {
+    fn default() -> Self {
+        Self {
+            cpu_count: 1,
```
eabatalov, line 1174: I don't see any sensible default values that work in all cases.
I think by default all these should be None which matches OSS experience where no limits are assigned to functions both in SDK attributes and in Executor.

server/data_model/src/lib.rs
```diff
@@ -1154,14 +1154,114 @@ pub struct FunctionURI {
...
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Builder)]
+#[builder(build_fn(skip))]
+pub struct FunctionExecutor {
+    pub id: String,
+    pub namespace: String,
+    pub compute_graph_name: String,
+    pub compute_fn_name: String,
+    pub version: Option<GraphVersion>,
```
eabatalov, line 1214: The version can't be optional in FunctionExecutor, when FE is created it's created for a particular version which never changes over FE lifetime.

server/proto/executor_api.proto
```diff
@@ -85,12 +85,15 @@ message ExecutorState {
     optional string executor_id = 1;
     optional bool development_mode = 2;
     optional string hostname = 3;
+    optional string executor_version = 8;
```
eabatalov, line 88: Unless this field is going to be used purely for observability we might want to split it into:
enum executor_flavor = EXECUTOR_FLAVOR_PLATFORM | EXECUTOR_FLAVOR_INDEXIFY
string executor_version // version of the flavor, e.g. indexify package version or tensorlake-platform-executor package version

These two fields make it more clear what the version means and later Server can have some logic based on executor_flavor.

server/data_model/src/lib.rs
eabatalov: The version is optional only in FunctionURI (function allowlist items).

server/data_model/src/lib.rs
```diff
@@ -1154,14 +1154,113 @@ pub struct FunctionURI {
...
+        let version = self.version.clone().flatten();
+        let status = self.status.clone().ok_or(anyhow!("status is required"))?;
+        let mut hasher = DefaultHasher::new();
+        namespace.hash(&mut hasher);
+        compute_graph_name.hash(&mut hasher);
+        compute_fn_name.hash(&mut hasher);
+        if let Some(version) = &version {
+            version.0.hash(&mut hasher);
+        }
+        let id = format!("{:x}", hasher.finish());
```
eabatalov, line 1241: I don't see where this id generation is coming from. Function Executor ID is supplied from Server to Executor when Server asks Executor to create a Function Executor. Later Executor reports state of the FE with the ID to Server.

eabatalov: If we generate ID from hash then it won't work without adding a random value. This is because there will be many FEs with the same graph, namespace, graph_version and function.

server/data_model/src/lib.rs
diptanu: @eabatalov I am thinking to keep the scheduler logic free of allowed function Uris, and have it deal with only function executors. When we see an executor report allow list, the idea is to create the function executor objects immediately in the executor and let the scheduler execute work on them when the executors report that the executors have started and are healthy. This makes the code path uniform in all cases. So we need the versions to be optional here.

server/proto/executor_api.proto
diptanu: This is for observability, I don't foresee a cluster having both open source and closed source executors.

server/data_model/src/lib.rs
eabatalov: Yep but Executor will need to know graph version for FE before it can create it. In desired executor state we can't send FE with None version id.

eabatalov:
We progressed past what this PR is doing now, we can close it.

PR #1287, opened 2025-03-17
Diptanu pr rebased on main

eabatalov:
This is https://github.com/tensorlakeai/indexify/pull/1285 rebased onto main

### DIFF
--- a/server/Cargo.lock
+++ b/server/Cargo.lock
@@ -790,6 +790,7 @@ dependencies = [
  "anyhow",
  "derive_builder",
  "indexify_utils",
+ "nanoid",
  "rand",
  "serde",
  "serde_json",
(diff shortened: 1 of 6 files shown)


diptanu:
Let’s push this into a branch until we have this feature working end to end. I am a little concerned with making big changes like this when we are onboarding customers in the cloud platform. I would like this to bake in a dev branch and see the executors working end to end.

eabatalov:
I think it's better to keep it open and gradually ingest it into Ben's work on migration to heatbeats.

eabatalov:
I'm actually surprised that tests passed for this rebased PR. I didn't fix any of them and I think it's dangerous to merge it in main. I agree with the point of not merging it to main.

seriousben:
This code is merged now :)

PR #1291, opened 2025-03-18
feat: stop persisting executor metadata

seriousben:
## Context

As part of changing the protocol between executors and servers, it has become obvious that persisting executors is not required and adds complexity.

## What

This PR moves executor to only live in memory. Additionally, we look at persisted allocations to determine whether some allocations were abandoned when the server restarts for executors that did not reconnect within 30s.

The graph processor is also now running within its own thread.

## Contribution Checklist

### DIFF
--- a/server/processor/src/graph_processor.rs
+++ b/server/processor/src/graph_processor.rs
@@ -220,9 +220,6 @@ impl GraphProcessor {
                     result
                         .new_allocations
                         .extend(placement_result.new_allocations);
-                    result
-                        .remove_allocations
-                        .extend(placement_result.remove_allocations);
                     result.updated_tasks.extend(placement_result.updated_tasks);
                     Ok(StateMachineUpdateRequest {
                         payload: RequestPayload::SchedulerUpdate(Box::new(result)),
--- a/server/processor/src/graph_processor.rs
+++ b/server/processor/src/graph_processor.rs
@@ -235,8 +235,7 @@ impl GraphProcessor {
             }
             ChangeType::ExecutorAdded(_) |
             ChangeType::ExecutorRemoved(_) |
-            ChangeType::TombStoneExecutor(_) |
-            ChangeType::HandleAbandonedAllocations => {
+            ChangeType::TombStoneExecutor(_) => {
                 let scheduler_update = self
                     .task_allocator
                     .invoke(&state_change.change_type, &mut indexes);
--- a/server/processor/src/graph_processor.rs
+++ b/server/processor/src/graph_processor.rs
@@ -229,7 +229,7 @@ impl GraphProcessor {
                         processed_state_changes: vec![state_change.clone()],
                     })
                 } else {
-                    error!("error creating tasks: {:?}", scheduler_update.err());
+                    error!("error invoking task creator: {:?}", scheduler_update.err());
                     Ok(StateMachineUpdateRequest {
                         payload: RequestPayload::Noop,
                         processed_state_changes: vec![state_change.clone()],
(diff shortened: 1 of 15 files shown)


server/processor/src/task_allocator.rs
```diff
@@ -64,43 +64,72 @@ impl TaskAllocationProcessor {
...
-                                allocation.task_key(),
-                            );
-                        }
-                    }
-                }
-
-                return Ok(SchedulerUpdateRequest {
+            ChangeType::HandleAbandonedAllocations => {
+                // Get all executor IDs from allocation_by_executor that aren't in executor_ids
+                let missing_executors: Vec<String> = indexes
```
diptanu, line 69: You could probably collect all the allocations that needs to be removed in this iteration. Pass the allocation down to the unallocate function instead of the executor ids to remove 2 x iteration.

diptanu: There is the map called allocations_by_fn which is executor id -> fn_name -> Allocation  we should use that map instead of  allocations_by_executor and try to remove this map altogether. Its not used anywhere else I believe

diptanu (changes requested):

server/processor/src/task_allocator.rs
```diff
@@ -64,43 +64,72 @@ impl TaskAllocationProcessor {
...
             }
+            ChangeType::TombStoneExecutor(ev) => self.unallocate(ev.executor_id.clone(), indexes),
             _ => {
                 error!("unhandled change type: {:?}", change);
                 return Err(anyhow!("unhandled change type"));
             }
         }
     }
 
+    pub fn unallocate(
```
diptanu, line 98: We could also try to re-allocate the tasks if there are any executors available. This will save some back and forth between state machine and task allocator.

diptanu (changes requested):
We should consider using threadpools at some point to run the task allocator and task creators to not block the Tokio tasks.

https://docs.rs/tokio-threadpool/0.1.18/tokio_threadpool/struct.ThreadPool.html

server/processor/src/task_allocator.rs
seriousben: > Allocation we should use that map instead of allocations_by_executor and try to remove this map altogether
I removed allocations_by_executor in my PR. Thanks for mentioning it. the /internal/allocations also returns fn allocations now.

> Pass the allocation down to the unallocate function instead of the executor ids to remove 2 x iteration.
I refactored things a lot more to allocate on deregistration. Take a look. I kept the 2 loop for now, but it only happens on startup and allows us to keep the same code for deregistration and handling of abandoned allocations.

server/processor/src/task_allocator.rs
seriousben: Done! I also reduce the number of state changes happening for executor removal and also fixed an issue where updated_tasks could contain the same task more than once.

server/processor/src/graph_processor.rs
```diff
@@ -238,7 +235,8 @@ impl GraphProcessor {
             }
             ChangeType::ExecutorAdded(_) |
             ChangeType::ExecutorRemoved(_) |
-            ChangeType::TombStoneExecutor(_) => {
+            ChangeType::TombStoneExecutor(_) |
+            ChangeType::HandleAbandonedAllocations => {
```
diptanu, line 239: Do we need this new state here? If you did the reconciliation of which executors are not available anymore, in the executor manager, you could simply write new Deregister(or delete, forget the request name) Executor requests into the state machine, and that would follow the path of tombstone.

server/processor/src/task_allocator.rs
```diff
@@ -57,42 +57,28 @@ impl TaskAllocationProcessor {
...
-                        } else {
-                            error!(
-                                "task of allocation not found in indexes: {}",
-                                allocation.task_key(),
-                            );
-                        }
-                    }
+            ChangeType::HandleAbandonedAllocations => {
+                // Get all executor IDs of executors that haven't registered.
+                let missing_executor_ids: Vec<String> = indexes
```
diptanu, line 66: See the comment above. You would do this logic in executor manager.

diptanu (commented):
The PR is looking good. Left a comment about not requiring an additional state change type.

server/processor/src/graph_processor.rs
seriousben: Done!

seriousben: This is not batching anymore, but should  be fine.

PR #1294, opened 2025-03-19
Add GH test action with Server <-> Executor gRPC enabled

eabatalov:
This allows us to continuously test for regressions when the new gRPC protocol is enabled.

This action is a copy of existing tests.yaml with extra executor CLI flag that enables gRPC Executor state reporting
at the supplied gRPC Server endpoint.

### DIFF
--- a/.github/workflows/tests_grpc.yaml
+++ b/.github/workflows/tests_grpc.yaml
@@ -1,0 +1,180 @@
+name: Tests with gRPC enabled
+
+on:
+  push:
+    branches:
+      - 'main'
+  pull_request:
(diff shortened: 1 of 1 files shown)


diptanu:
@eabatalov why not always enable gRPC state reporting? We can get the new version of server deployed with the mock interfaces, then deploy the executor. We could skip some of these additional setup/infrastructure.

eabatalov:
@diptanu because OSS customers are running without gRPC enabled so we need to test if Indexify still works without gRPC. My idea was that we provide backward compatibility (running without gRPC enabled) until we fully implement gRPC. Then we ask customers to migrate to gRPC and make the Executor grpc flag required.

PR #1298, opened 2025-03-21
Making the server get the URLs of stored outputs from executor and returning URLs as inputs to functions 

diptanu:
Getting the server ready to accept function output urls from the executor. 

We will expect the executor to send JSON payload of the URLs where it stores the fucntion outputs - 

```
{
  "url": ..
   "checksum": ...
   "size":...
}

Which is same as `PutResult` we use in the server internally to track the outputs. Everything else more or less remains same. 

We are also sending the executor the DataPayload object which is the URL of the input which it needs to download

### DIFF
--- a/server/blob_store/src/lib.rs
+++ b/server/blob_store/src/lib.rs
@@ -50,7 +50,7 @@ impl Default for BlobStorageConfig {
     }
 }
 
-#[derive(Debug, Clone)]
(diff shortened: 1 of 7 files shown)


eabatalov:
This looks obsolete as we're not going to support direct Executor blob store operations in the deprecated SSE stream tasks protocol.

PR #1300, opened 2025-03-21
fix: deletion not deleting tasks and allocations

seriousben:
## Context

Deleting a CG or an allocation results in leaked tasks and allocations. This results in error without currently impacting the server.

## What

This PR delete tasks and allocations correctly when an invocation gets deleted.
Additionally, we add tests to ensure nothing is left behind anymore include what file gets GC'd.

This PR also adds a migration to delete orphaned tasks, orphaned allocations and update the allocation storage key.

## Testing

Tested the migration against a snapshot of the platform's data.

```
{"timestamp":"2025-03-21T21:14:25.752533Z","level":"INFO","message":"Migrated 1 allocations and deleted 0 orphaned allocations from 1 total allocations","target":"state_store::migrations","span":{"name":"migrate_v5_to_v6_migrate_allocations"}}
{"timestamp":"2025-03-21T21:14:46.238261Z","level":"INFO","message":"Deleted 21075 orphaned tasks out of 299284","target":"state_store::migrations","span":{"name":"migrate_v5_to_v6_clean_orphaned_tasks"}}
```

I am surprised only one allocation is found in that state. I suspect the cleanup I did around executors in memory is helping.

## Contribution Checklist


server/src/testing.rs
```diff
@@ -191,13 +201,38 @@ impl FinalizeTaskArgs {
...
     pub fn task_outcome(mut self, task_outcome: TaskOutcome) -> FinalizeTaskArgs {
         self.task_outcome = task_outcome;
         self
     }
+
+    pub fn diagnostics(mut self, stdout: bool, stderr: bool) -> FinalizeTaskArgs {
+        self.diagnostics = Some(TaskDiagnostics {
+            stdout: if stdout {
+                Some(DataPayload {
+                    path: "stdout".to_string(),
```
seriousben, line 217: This should use the task Id so it generates a different path each time.

Otherwise this makes all tasks have the same diagnostics.

Doing this should bump the gcURLs from 6 to 8 since we are missing 2 diagnostics.

diptanu (approved):
LGTM

PR #1302, opened 2025-03-22
Task reporting api

diptanu:
Added a new GRPC endpoint for task reporting. This will be used by the executor when it only sends task outcomes to the server, and handles uploading the output and logs to storage on it's own. 

Started to build this in the current HTTP endpoint for internal ingest and it was getting complex. This was much easier to do without breaking the existing code path.


indexify/pyproject.toml
```diff
@@ -23,7 +23,7 @@ python = "^3.9"
 aiohttp = "^3.11.0"
 prometheus-client = "^0.21.1"
 # Adds function-executor binary and utils lib.
-tensorlake = ">=0.1"
+tensorlake = "0.1.44"
```
eabatalov, line 26: With the current >= approach the effective tensorlake version is stored in indexify lock file. To update the lock file to the latest version just type `poetry update tensorlake`. This is the approach we're using in Indexify and internal repos. See some more details here: https://github.com/tensorlakeai/compute-engine-internal/blob/main/platform_executor/README.md

.github/workflows/tests.yaml
```diff
@@ -126,7 +126,7 @@ jobs:
         with:
           run: |
             cd indexify
-            poetry run indexify-cli executor --dev &
+            poetry run indexify-cli executor --grpc-server-addr localhost:8901 --dev &
```
eabatalov, line 129: I'll do this change now and remove the grpc only CI workflow as we discussed.

indexify/src/indexify/executor/executor.py
```diff
@@ -134,6 +129,18 @@ def __init__(
                 ExecutorStatus.EXECUTOR_STATUS_STARTING_UP
             )
 
+        # Looks like grpc doesn't like mixing streaming and non-streaming calls.
```
eabatalov, line 132: This should work without separate channel. Let's not create a separate channel, this is masking a bug.

indexify/src/indexify/executor/api_objects.py
```diff
@@ -49,5 +49,17 @@ class TaskResult(BaseModel):
...
+class DataPayload(BaseModel):
+    path: str
+    size: int
+    sha256_hash: str
+
+
+class IngestFnOutputsResponse(BaseModel):
+    data_payloads: List[DataPayload]
+    stdout: Optional[DataPayload] = None
+    stderr: Optional[DataPayload] = None
```
eabatalov, line 61: What's the reason for using BaseModel instead of gRPC protos?
I was expecting that most of/all api_objects.py goes away once we finish gRPC migration in full.

indexify/src/indexify/proto/executor_api.proto
```diff
@@ -154,6 +154,59 @@ message DesiredExecutorState {
     optional uint64 clock = 3;
 }
```
eabatalov, line 156: Nit. Please add
`// ===== report_task_outcome RPC =====`
for consistency :)

indexify/src/indexify/executor/api_objects.py
eabatalov: After looking more at the code this looks like a temporary set of api models for Executor uploading files to Server, right?

PR #1309, opened 2025-03-24
feat: implement server heartbeats

seriousben:
## Context

As part of switching executor protocol to grpc for dynamic scheduling support, we are moving to heartbeat based executor management.

## What

This PR:
- Add back removed tombstoned executor support
- Add a worker to removed executor that havent send an heartbeat for more than 30s
- Change executor registration to upsert logic to allow heartbeat and task stream to coexist for now
- Remove the task_stream deregister logic

The heartbeat mechanism reuses the same future which can be mutated by a watcher sender. 

## Contribution Checklist

### DIFF
--- a/server/src/executors.rs
+++ b/server/src/executors.rs
@@ -129,7 +129,7 @@ impl ExecutorManager {
         }
 
         // 5. Register the executor to upsert its metadata
-        let err = self.register_executor(executor.clone()).await;
+        let err = self.register_executor(executor.clone(), false).await;
         if let Err(e) = err {
             error!("failed to register executor {}: {:?}", executor.id.get(), e);
             return Err(e);
@@ -264,9 +264,16 @@ impl ExecutorManager {
         Ok(())
     }
 
-    pub async fn register_executor(&self, executor: ExecutorMetadata) -> Result<()> {
+    pub async fn register_executor(
+        &self,
+        executor: ExecutorMetadata,
+        for_task_stream: bool,
+    ) -> Result<()> {
         let sm_req = StateMachineUpdateRequest {
-            payload: RequestPayload::UpsertExecutor(UpsertExecutorRequest { executor }),
+            payload: RequestPayload::UpsertExecutor(UpsertExecutorRequest {
+                executor,
+                for_task_stream,
+            }),
             processed_state_changes: vec![],
         };
         self.indexify_state.write(sm_req).await
@@ -323,6 +330,11 @@ impl ExecutorManager {
 
 pub fn schedule_deregister(ex: Arc<ExecutorManager>, executor_id: ExecutorId, duration: Duration) {
     tokio::spawn(async move {
+        trace!(
+            executor_id = executor_id.get(),
+            deregister_after_s = duration.as_secs_f64(),
+            "Scheduling deregistration of executor"
+        );
         tokio::time::sleep(duration).await;
         let ret = ex.deregister_executor(executor_id.clone()).await;
         if let Err(e) = ret {
@@ -363,7 +375,7 @@ mod tests {
             state: Default::default(),
             tombstoned: false,
         };
-        executor_manager.register_executor(executor).await?;
+        executor_manager.register_executor(executor, true).await?;
 
         let executors = indexify_state
             .in_memory_state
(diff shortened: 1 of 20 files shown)


server/src/executors.rs
```diff
@@ -44,21 +90,150 @@ impl ExecutorManager {
...
+            // 2. Wait for the next deadline or a notification
+            match next_deadline {
+                // 2.1 If there is a next deadline, wait until it or a notification
+                Some(deadline) => {
+                    let now = Instant::now();
+                    if deadline > now {
+                        trace!("Waiting for next deadline");
+                        tokio::select! {
+                            // Wait until the deadline arrives
+                            _ = time::sleep_until(deadline) => {
```
diptanu, line 167: This will fire even if the deadline was resetted right? I think if you used a global timer and resetted it when an executor heartbeat things would be simpler.

seriousben: I simplified things for easier testing.

I don't quite understand what you mean though. Do you mean waking up every 15s to look at deadlines for example?

I initially played with the idea of waking up the task to reset the timer, but doing so defeats the purpose since it would reset every 5s and wake up the task. Instead, we are now waiting for the earliest deadline which means we wake up closer to every 30s except if there are no executors.

We could have a global timer, but that would requiring waking up more often to make sure we catch lapsed executor in a timely manner.

If there is a way to change a timer on the fly so that the task does not have to wake up, that would be ideal. But at the same time, the task does not do much typically.

What I am more annoyed by is the use of the RwLock.

diptanu: Timers have a reset API, you can change when they wake up on the fly. 

I am suggesting to reset when the timer fires up every time we get a new a heartbeat from the cluster. 

After re arranging the PQ when a new HB comes. Peak the PQ, whatever is at the top (the closest deadline), reset the global timer to that. Now you wake up only when a executor has to legitimately be de-registered

server/src/executors.rs
```diff
@@ -44,21 +92,188 @@ impl ExecutorManager {
...
+            was_empty
+        };
+
+        // 4. Notify of a first executor to wake up the monitor
+        if first_executor {
+            self.notify.notify_one();
+        }
+
+        // 5. Register the executor to upsert its metadata
+        let err = self.register_executor(executor.clone(), false).await;
```
diptanu, line 132: Looks like we always register the executor when it's heartbeated? We have a hash in the grpc request we could track to know if it needs upsert. I would avoid this step unless there is a change in the executor state, like a new FE coming up, etc.

server/src/executors.rs
```diff
@@ -44,21 +92,188 @@ impl ExecutorManager {
...
+                // Deadline reached - process lapsed executors
+                WaitResult::Deadline => {
+                    trace!("Received deadline");
+                    // Attempt to deregister executors that have missed their heartbeat
+                    if let Err(err) = self.process_lapsed_executors().await {
+                        error!("Failed to process lapsed executors: {:?}", err);
+                    }
+
+                    // Wait for 1 second to batch potential subsequent executor lapsing
+                    tokio::time::sleep(Duration::from_secs(1)).await;
```
diptanu, line 195: Do we need this? It's probably a good optimization but also slows down the scheduler by a second. Unless we have 1000s of executors timing out at the same time it's probably not very helpful

server/src/executors.rs
```diff
@@ -115,6 +330,11 @@ impl ExecutorManager {
 
 pub fn schedule_deregister(ex: Arc<ExecutorManager>, executor_id: ExecutorId, duration: Duration) {
```
diptanu, line 331: Looks like this is being used now only in tests? Might be good to just move this into the test module

diptanu (approved):

server/src/executors.rs
seriousben: Done! Got it working!

server/src/executors.rs
seriousben: Since this is in memory and we don't fire any state change, I will look at using this when we start creating other indices from the data. But right now it is simply an insert -> `O(log n)`

server/src/executors.rs
seriousben: This has been reworked. I still kept this but implemented in a more elegant way. Locally I was testing with 100s executor and it was evident that something like this was needed.

The way I implemented it is just as a min sleep of 1s.

server/src/executors.rs
seriousben: This is completely removed now since we are removing stream deregistration logic.

PR #1313, opened 2025-03-28
fix: only rely on grpc heartbeating for executor registration

seriousben:
## Context

Currently there is a discrepancy between the data we get on tasks stream creation and heartbeating. This results in some errors on task stream creation when passing a function uri.

## What

This PR:
1- Ensures some tests are failing if function uri are not working - This was failing silently before.
2- Stop registration of executors in the task stream. Instead the task stream relies on the heartbeat to happen before. If the executor tries to get the task stream before the server knows about the executor (via a heartbeat), the endpoint will just disconnect without sending anything in the stream. We do that after waiting for some time to prevent spurious re-connections from the executor.

## Contribution Checklist


server/state_store/src/lib.rs
```diff
@@ -431,7 +429,11 @@ pub fn task_stream(state: Arc<IndexifyState>, executor_id: ExecutorId) -> TaskSt
             rx
         } else {
             // Executor not found, return empty stream.
-            error!(executor_id=executor_id.get(), "executor not found, stopping task stream");
+            warn!(executor_id=executor_id.get(), "executor not found, stopping task stream");
+            // Prevent spurious task_stream creations for executors that are unknown the server.
+            // This can happen if the heartbeat is not sent before the task stream is created
+            // or if the executor is removed and it is still retrying to create a task stream.
+            tokio::time::sleep(std::time::Duration::from_secs(2)).await;
```
seriousben, line 436: Moving this in the executor. cc @eabatalov

diptanu (approved):

PR #1314, opened 2025-03-28
Implement Executor gRPC state reconciler with unit tests

eabatalov:
The unit tests are needed to prove that the reconciler works as expected. It's especially important for Python code that doesn't have compile time type safety.

The unit tests don't need to be maintained forever but at the very least we need them until Server starts using the reconciler so we can cover the reconciler using integration tests.

The unit tests cover all known important scenarios and can be used as an example on how to use the new protocol on Server side. Based on the tests it looks like the new protocol has all the required functionality.


eabatalov:
Rebased on main, the only failed test is
```
FAIL: test_server_distributes_invocations_fairly_between_two_executors
```

Which is unrelated to this PR.

PR #1319, opened 2025-03-31
fix: review all usage of rocksdb iterator prefix

seriousben:
## Context

Upgrading task versions would result in allocation not finding their compute graph version.

## What

This PR ensures that upgrading the version of a compute graph only updates tasks and invocation ctx that need to be updated.

## Contribution Checklist

### DIFF
--- a/server/state_store/src/scanner.rs
+++ b/server/state_store/src/scanner.rs
@@ -209,13 +209,11 @@ impl StateReader {
                 }
             }
         };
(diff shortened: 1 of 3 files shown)


server/state_store/src/scanner.rs
```diff
@@ -260,58 +260,6 @@ impl StateReader {
         })
     }
 
-    pub fn filter_cf<T, F>(
```
seriousben, line 263: deadcode

diptanu (approved):

PR #1337, opened 2025-04-07
Use data payloads for inputs and outputs in Executor gRPC protocol definitions

eabatalov:
The changes in this PR define how Executor is going to directly upload and download to/from S3.

The idea is to use the same DataPayload message for task inputs and outputs. The new Task field output_payload_uri_prefix tells Executor under which S3 path or filesystem dir to put output files.

Server side report task outcome RPC is updated to work with previous and new Executor behaviors when calling this RPC. The passing integration tests prove this because there are no Executor changes in this PR.

### DIFF
--- a/server/src/executor_api.rs
+++ b/server/src/executor_api.rs
@@ -26,6 +26,7 @@ use data_model::{
 use executor_api_pb::{
     executor_api_server::ExecutorApi,
     AllowedFunction,
+    DataPayloadEncoding,
     DesiredExecutorState,
     ExecutorState,
     ExecutorStatus,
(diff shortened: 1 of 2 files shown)


diptanu (approved):

server/src/executor_api.rs
```diff
@@ -383,7 +372,8 @@ impl ExecutorApi for ExecutorAPIService {
         for output in request.get_ref().fn_outputs.clone() {
             let path = output
                 .path
-                .ok_or(Status::invalid_argument("path is required"))?;
+                .or(output.uri)
```
diptanu, line 375: @eabatalov You could move out the encoding from data payloads to the top level struct. You can assume all the output encoding for a function is the same.

eabatalov:
I found some gaps in this PR, please don't merge it.

PR #1355, opened 2025-04-10
Server support for Executor side blob upload and download in HTTP mode

eabatalov:
One notable difference is that Server has to convert its internal blob paths
into full blob URIs when sending data payloads to Executor and Server.
Server also has to strip full blob URIs it gets from Executor to get blob paths.
The paths depends on blob store driver used by Server. This is documented
in code comments.

I tested this with Executor side changes and all tests are passing
both with S3 and local FS blob stores.

Sending this in a separate PR from Executor changes to ensure that
Executor still works withouts its changes. This is to ensure
backward compatibility.

### DIFF
--- a/server/blob_store/src/lib.rs
+++ b/server/blob_store/src/lib.rs
@@ -58,6 +58,8 @@ pub struct PutResult {
 
 pub struct BlobStorage {
     object_store: Arc<dyn ObjectStore>,
+    url_scheme: String,
+    url: String,
     path: Path,
     metrics: blob_storage::Metrics,
 }
@@ -69,6 +71,8 @@ impl BlobStorage {
         let (object_store, path) = Self::build_object_store(url)?;
         Ok(Self {
             object_store: Arc::new(object_store),
+            url_scheme: url.parse::<Url>()?.scheme().to_string(),
+            url: url.clone(),
             path,
             metrics: blob_storage::Metrics::new(),
         })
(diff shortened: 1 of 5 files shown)


diptanu (approved):
LGTM. @seriousben can you please review this as well?

server/blob_store/src/lib.rs
```diff
@@ -69,6 +71,8 @@ impl BlobStorage {
         let (object_store, path) = Self::build_object_store(url)?;
         Ok(Self {
             object_store: Arc::new(object_store),
+            url_schema: url.parse::<Url>()?.scheme().to_string(),
```
seriousben, line 74:
```suggestion
            url_scheme: url.parse::<Url>()?.scheme().to_string(),
```

server/blob_store/src/lib.rs
```diff
@@ -97,6 +101,14 @@ impl BlobStorage {
         self.object_store.clone()
     }
 
+    pub fn get_url(&self) -> String {
+        self.url.clone()
+    }
+
+    pub fn get_url_schema(&self) -> String {
```
seriousben, line 108:
```suggestion
    pub fn get_url_scheme(&self) -> String {
```

server/blob_store/src/lib.rs
eabatalov: Thank you. Renamed in the new revision of this PR.

PR #1361, opened 2025-04-16
Extend FunctionExecutorServerConfiguration with graph metadata

miguelhrocha:
## Context

The new image builder implementation centralizes the knowledge of image URIs into the image builder V2 service. This is due to security constraints implemented for pulling images from our container registry.

The signature to pull images in this new service requires the usage of the namespace, graph name, graph version, and function name. 

At the moment, these are not accessible to services looking to implement a `FunctionExecutorServerFactory` that relies on those values.

This PR extends `FunctionExecutorServerConfiguration` to hold those values.

## What

Extend FunctionExecutorServerConfiguration with the attributes:

- `graph_name`
- `function_name`
- `graph_version`

## Testing

Extend API, doesn't remove anything, safe change.

## Contribution Checklist


indexify/src/indexify/executor/function_executor/server/function_executor_server_factory.py
```diff
@@ -19,6 +19,9 @@ class FunctionExecutorServerConfiguration:
     executor_id: str
     function_executor_id: str
     namespace: str
+    graph: str
```
eabatalov, line 22: Please follow the existing naming convention in Executor, see example here https://github.com/tensorlakeai/indexify/blob/main/server/proto/executor_api.proto

indexify/src/indexify/executor/function_executor/server/function_executor_server_factory.py
```diff
@@ -19,6 +19,9 @@ class FunctionExecutorServerConfiguration:
...
+    function: str
+    version: Optional[str]
```
eabatalov, line 24: Version can't be optional. An FE can't be created for an unknown version.

eabatalov: If the source data for the version has it as Optional then it's a tech debt in the source of the data, please ignore it. It's never optional in reality.

eabatalov (changes requested):
Small stylistic issues, functionally everything is correct.

miguelhrocha:
Thank you for the review @eabatalov, I've addressed your comments

eabatalov (changes requested):
Let's not update the lock file as it might introduce bugs without any reason to do the updates right now.

eabatalov (approved):
This would probably results in other folks formatters formatting the code back but not a big deal. Thank you for addressing all the comments.

PR #1363, opened 2025-04-16
Propagate Function timeouts, retries, resources to Server and Executor

eabatalov:
The features are not implemented in this PR, but the data about them is available both in Server and Executor.

This is RFC for now as I don't have time to make all tests green. Also while testing manually I found a bug with function timeouts, they are always 5 minutes regardless of the value set in function attributes.


server/data_model/src/lib.rs
```diff
@@ -247,6 +247,62 @@ impl ImageInformation {
...
+        }
+    }
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct NodeRetryPolicy {
+    pub max_retries: u32,
+    pub initial_delay_ms: u32,
+    pub max_delay_ms: u32,
+    // The multiplier value is 1000x of the actual value to avoid working with floating point.
```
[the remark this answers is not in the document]
diptanu: This makes sense.

server/data_model/src/lib.rs
```diff
@@ -1011,6 +1103,12 @@ pub struct Task {
     pub graph_version: GraphVersion,
     pub image_uri: Option<String>,
     pub secret_names: Option<Vec<String>>,
+    #[serde(default)]
```
diptanu, line 1106: Do we need this here? We are setting retry policies to the functions, so we can get this metadata from the ComputeGraph object. When the executor first starts up a function executor it can pull the metadata about the compute graph and cache these values. This reduces some duplicate data across tasks of the same function.

suggesting moving these values to the ComputeGraphVersion object.

server/data_model/src/lib.rs
```diff
@@ -1141,6 +1239,19 @@ impl TaskBuilder {
         let duration = current_time.duration_since(UNIX_EPOCH).unwrap();
         let creation_time_ns = duration.as_nanos();
         let id = uuid::Uuid::new_v4().to_string();
+        let secret_names = self
```
diptanu, line 1242: I think we can do the same for secrets

server/data_model/src/lib.rs
eabatalov: Nice idea, I see a bunch of image uri and secret names updates code around in Server and it can be delete with this.

diptanu (approved):

PR #1369, opened 2025-04-21
Dynamic sched

diptanu:
Dynamic Scheduling WIP


indexify/src/indexify/executor/grpc/function_executor_controller.py
```diff
@@ -354,11 +354,11 @@ async def _create_function_executor(
     if function_executor_description.HasField("image_uri"):
         config.image_uri = function_executor_description.image_uri
     if function_executor_description.HasField("resources"):
-        resources: FunctionExecutorResources = function_executor_description.resources
-        config.cpu_ms_per_sec = resources.cpu_ms_per_sec
+        resources: HostResources = function_executor_description.resources
+        config.cpu_ms_per_sec = resources.cpu_count * 1000
```
eabatalov, line 358: There's a problem here. cpu_count is integer while customers can set fractions of CPU in their function resources. It looks like this is why FE Resources were introduced. There's going to be even more difference of HostResources and FE resources in the future. So I think we should keep FE resources struct.

PR #1380, opened 2025-04-24
Report total Executor host resources to Server

eabatalov:
Also includes minor Executor side fixes.

I dropped get free resources function as it's not used right now.

Heavily based on Executor side changes from dynamic scheduling PR https://github.com/tensorlakeai/indexify/pull/1369


diptanu (approved):
LGTM

PR #1383, opened 2025-04-25
Implement task output caching in Indexify

### LINKED ISSUE #1339: Cache Requests
Cache the I/O of functions if the code version and the Inputs are same. Provide flags to override cache by the user.

earhart:
## Context

The idea here is to cache function outputs and reuse them in subsequent function calls iff requested.

Fixes #1339 

## What

This PR allows the client to specify a caching key when specifying graph functions; this indicates to the server that it's allowed to cache the function applications.  A new TaskCache component observes output ingestion (creating cache entries), and intercepts tasks between creation and executor allocation; when it sees a cache hit, it skips the allocation, and augments the scheduler update with the cached task outputs; when the state machine observes the augmented scheduler update, it resolves the task as though it had just completed, generating output ingestion events to release downstream tasks.

## Testing

Currently, fairly basic: running a workload multiple times.  This PR needs further tests before it lands.

## Contribution Checklist

### DIFF
--- a/server/data_model/src/lib.rs
+++ b/server/data_model/src/lib.rs
@@ -1859,7 +1859,6 @@ pub struct InvokeComputeGraphEvent {
     pub invocation_id: String,
     pub namespace: String,
     pub compute_graph: String,
-    pub invocation_data_sha256_hash: String,
 }
 
 #[derive(Clone, Serialize, Deserialize, Debug, PartialEq)]
--- a/server/data_model/src/lib.rs
+++ b/server/data_model/src/lib.rs
@@ -336,6 +336,21 @@ pub struct DynamicEdgeRouter {
     pub retry_policy: NodeRetryPolicy,
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Hash, Eq)]
+pub struct CacheKey(String);
+
+impl CacheKey {
+    pub fn get(&self) -> &str {
+        &self.0
+    }
+}
+
+impl From<&str> for CacheKey {
+    fn from(value: &str) -> Self {
+        Self(value.to_string())
+    }
+}
+
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Default)]
 pub struct ComputeFn {
     pub name: String,
@@ -355,6 +370,7 @@ pub struct ComputeFn {
     pub resources: NodeResources,
     #[serde(default)]
     pub retry_policy: NodeRetryPolicy,
+    pub cache_key: Option<CacheKey>,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
@@ -442,6 +458,10 @@ impl Node {
             Node::Router(router) => router.name.clone(),
             Node::Compute(compute) => compute.name.clone(),
         };
+        let cache_key = match self {
+            Node::Router(_) => None,
+            Node::Compute(compute) => compute.cache_key.as_ref().and_then(|v| Some(v.clone())),
+        };
         let task = TaskBuilder::default()
             .namespace(namespace.to_string())
             .compute_fn_name(name)
@@ -450,6 +470,7 @@ impl Node {
             .input_node_output_key(input_key.to_string())
             .reducer_output_id(reducer_output_id)
             .graph_version(graph_version.clone())
+            .cache_key(cache_key)
             .build()?;
         Ok(task)
     }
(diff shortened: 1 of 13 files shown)


earhart:
This isn't quite ready to go in as-is, but if you'd like to give it an early review, the basic pieces are here.  The cache needs to respect namespaces and report cache hits to the client; we may also want to add cache-clearing and monitoring APIs in this PR.

server/data_model/src/lib.rs
```diff
@@ -335,6 +335,12 @@ pub struct DynamicEdgeRouter {
     pub retry_policy: NodeRetryPolicy,
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Hash, Eq)]
+pub struct CacheKey {
+    pub algorithm: String,
```
diptanu, line 340: Can this be an enum? If you want Enum <> String conversion somewhere else, then take a look at the crate called Strum.

server/data_model/src/lib.rs
```diff
@@ -436,11 +443,16 @@ impl Node {
         input_key: &str,
         reducer_output_id: Option<String>,
         graph_version: &GraphVersion,
+        input_sha256_hash: Option<String>,
```
diptanu, line 446: Do we need to store this value here? They can be retrieved from the NodeOutputs columnfamily using the input key. RocksDB has MultiGet to make lookups across multiple column family faster. The issue every field in Task adds to storage space.

diptanu (commented):
Need more time to review. Leaving some comments here.

server/data_model/src/lib.rs
earhart: My thinking here was that it's essentially an opaque string from the client; the server's just checking for equality.  We could also replace this struct with a single string, the way the [Content-Digest](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Digest) header works -- `"{algorithm}={hash}"`.  Thoughts?

server/data_model/src/lib.rs
earhart: Done

diptanu (approved):
@earhart Looks great, have some small comments. Thanks for working on this very patiently!

server/state_store/src/lib.rs
```diff
@@ -337,15 +339,27 @@ impl IndexifyState {
...
-                            invocation_events::TaskCreated {
-                                invocation_id: task.invocation_id.clone(),
-                                fn_name: task.compute_fn_name.clone(),
-                                task_id: task.id.to_string(),
-                            },
-                        ));
+                    if sched_update.cached_task_outputs.contains_key(&task.id) {
+                        let _ =
+                            self.task_event_tx
+                                .send(InvocationStateChangeEvent::TaskMatchedCache(
```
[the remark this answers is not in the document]
earhart: I kinda like having it as a different event: this way, TaskCreated always gets matched up with a corresponding TaskAssigned and TaskCompleted.  Also, this event name is very prominent in the text output from the cli, so it's very, very obvious to users what's going on.

server/state_store/src/requests.rs
```diff
@@ -55,11 +55,18 @@ impl FunctionExecutorIdWithExecutionId {
...
+    pub task: Task,
+    pub node_outputs: Vec<NodeOutput>,
+}
+
 #[derive(Debug, Clone, Default)]
 pub struct SchedulerUpdateRequest {
     pub new_allocations: Vec<Allocation>,
     pub remove_allocations: Vec<Allocation>,
     pub updated_tasks: HashMap<TaskId, Task>,
+    pub cached_task_outputs: HashMap<TaskId, CachedTaskOutput>,
```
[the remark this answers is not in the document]
earhart: Yes -- that actually happens in `task_cache.rs::try_allocate()`, where we insert the task into `cached_task_outputs`.

server/state_store/src/state_changes.rs
```diff
@@ -41,6 +41,7 @@ pub fn invoke_compute_graph(
             namespace: request.namespace.clone(),
             invocation_id: request.invocation_payload.id.clone(),
             compute_graph: request.compute_graph_name.clone(),
+            invocation_data_sha256_hash: request.invocation_payload.payload.sha256_hash.clone(),
```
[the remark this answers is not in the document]
earhart: Done

PR #1389, opened 2025-04-30
Skip timeout tests if SSE stream is used

eabatalov:
The timeouts are only implemented in full gRPC mode when SSE stream is not used. SSE stream is still in use because gRPC mode is not ready for production yet. So need to skip these tests in SSE stream Mode.


miguelhrocha (approved):

miguelhrocha (approved):

eabatalov:
All test failures are known unrelated issues

PR #1362, opened 2025-04-16
feat: get desired state implementation

seriousben:
## Contribution Checklist

- [ ] If a Python package was changed, please run `make fmt` in the package directory.
- [ ] If the server was changed, please run `make fmt` in `server/`.
- [ ] Make sure all PR Checks are passing.

PR #1391, opened 2025-05-01
Bring back --enable-grpc-state-reconciler Executor CLI flag (False by default)

eabatalov:
By default this flag is False which means that we'll use SSE stream by default. We need to be able to use the legacy SSE stream mode to run with the stable published Server that doesn't use gRPC yet and still uses SSE stream mode.

--enable-grpc-state-reconciler was originally removed in https://github.com/tensorlakeai/indexify/pull/1362/files#diff-e207854b655720e9130572ef90eb931ffe8460c318ddd8f3c916a8756a29c066 we're bringing it back here.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -127,7 +127,7 @@ jobs:
         with:
           run: |
             cd indexify
-            poetry run indexify-cli executor --grpc-server-addr localhost:8901 --dev &
+            poetry run indexify-cli executor --grpc-server-addr localhost:8901 --dev --enable-grpc-state-reconciler &
             echo $! > /tmp/indexify-executor.pid &
 
           wait-on: |
--- a/indexify/src/indexify/cli/cli.py
+++ b/indexify/src/indexify/cli/cli.py
@@ -131,6 +131,16 @@ def executor(
             "Specified as <key>=<value>",
         ),
     ] = [],
+    enable_grpc_state_reconciler: Annotated[
+        bool,
+        typer.Option(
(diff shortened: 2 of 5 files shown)


eabatalov:
The test failures are known issues not related to the PR:

```
======================================================================
FAIL: test_server_distributes_invocations_fairly_between_two_executors (__main__.TestServerTaskDistribution)
  File "/home/runner/work/indexify/indexify/indexify/tests/./cli/test_server_task_distribution.py", line 87, in test_server_distributes_invocations_fairly_between_two_executors
    self.assertLess(invocations_count, 125)
AssertionError: 200 not less than 125
======================================================================
FAIL: test_server_redistributes_invocations_when_new_executor_joins (__main__.TestServerTaskDistribution)
  File "/home/runner/work/indexify/indexify/indexify/tests/./cli/test_server_task_distribution.py", line 143, in test_server_redistributes_invocations_when_new_executor_joins
    self.assertLess(invocations_count, 150)
AssertionError: 200 not less than 150
---------------------------------------------------------------------
======================================================================
FAIL: test_tasks_routing_and_distribution (__main__.TestFunctionAllowlist)
  File "/home/runner/work/indexify/indexify/indexify/tests/./cli/test_function_allowlist.py", line 269, in test_tasks_routing_and_distribution
    self.assertTrue(
AssertionError: False is not true : Executor dev_mode invocation count (899) is not within 20% of expected count (400). Distribution: {'dev_mode': 899, 'function_a': 1, 'function_b': 300, 'function_c': 400}
```

miguelhrocha (approved):

PR #1396, opened 2025-05-06
Update Function resource defaults and min values

eabatalov:
Bump min CPU cores to 1 and min memory to 1 GB.
Set defaults to the same values.
See details about the new values for resource limits in https://github.com/tensorlakeai/tensorlake/commit/7eda092ddb03a78cb4f498dc4e7637968bce3c4f


eabatalov:
The test failure is due to flake Executor shutdown code on grpc path. This is not related to this PR and will be fixed later.

j3m7 (approved):

PR #1412, opened 2025-05-09
Simplify locking in task_allocator.rs

earhart:
## Context

The way we lock InMemoryState in task_allocator.rs is a bit tricky to analyze -- what changes while the lock is dropped and reacquired?

## What

This change breaks out the entrypoints to task_allocator.rs; the new entrypoints acquire a write lock on InMemoryState, and then call the existing entrypoints, which are modified to no longer acquire the locks (since they're unneeded and unavailable). 

## Testing

Ran the Indexify tests against a server built with the change.

## Contribution Checklist

- [NA] If a Python package was changed, please run `make fmt` in the package directory.

### DIFF
--- a/server/processor/src/graph_processor.rs
+++ b/server/processor/src/graph_processor.rs
@@ -17,7 +17,7 @@ use state_store::{
 use tokio::sync::Notify;
 use tracing::{debug, error, info, trace};
 
-use crate::{task_allocator::TaskAllocationProcessor, task_cache, task_creator};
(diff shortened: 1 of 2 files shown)


diptanu (approved):

earhart:
AFAICT, the test that's failing is triggering some sort of edge condition / flaky behavior; I can run it in a tight loop on my machine, succeeding over and over again, until it fails.  (And then once it's in the failing state, it continues failing until the executor is restarted).

PR #1416, opened 2025-05-09
Bring Executor dev mode reporting back

eabatalov:
Not reporting dev mode in Executor breaks all CI with stable Server that uses SSE stream.

Also comment out known failing test in CI


eabatalov:
The test failure is not related to this PR

j3m7 (approved):

PR #1423, opened 2025-05-14
Remove SSE stream code from Executor

eabatalov:
Also remove deprecated fields from Executor proto API and make previously optional proto fields required cause now Server provides them and we can remove all the if branches from Executor.
Other more major refactoring will be done later.


indexify/poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 2.1.2 and should not be changed by hand.
+# This file is automatically @generated by Poetry 2.0.0 and should not be changed by hand.
```
earhart, line 1: Older version of Poetry?

indexify/src/indexify/executor/function_executor/server/subprocess_function_executor_server_factory.py
```diff
@@ -12,14 +12,17 @@
 
 
 def get_free_tcp_port(iface_name="localhost") -> int:
-    tcp = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-    tcp.bind((iface_name, 0))
-    _, port = tcp.getsockname()
-    tcp.close()
-    return port
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
```
earhart, line 15: Not that this needs to be changed in this PR, but is there some reason we're not using an AF_UNIX socket?

server/proto/executor_api.proto
earhart, line 1: I'm still not a big fan of having two copies of this file in the repo -- makes me want to either convert everything to Rust or use a general build tool.  Any thoughts on Bazel?  :-)

earhart (approved):
LGTM, although someone more familiar with the codebase might spot something I missed.

indexify/poetry.lock
eabatalov: It just depends on the dev machine, we need it to be > 2.0.0

indexify/src/indexify/executor/function_executor/server/subprocess_function_executor_server_factory.py
eabatalov: UNIX is indeed sufficient for OSS. Now as we don't support Windows, this should work.
Though with FEs in VMs this won't work so TCP is still the main use case.

server/proto/executor_api.proto
eabatalov: Thoughts: don't touch what works already :) 
The reason the file exists is purely due to missing feature in Python gRPC. See https://github.com/tensorlakeai/indexify/blob/main/indexify/Makefile#L11

Also I didn't particularly like Brazil back in the days.

indexify/src/indexify/executor/function_executor/server/subprocess_function_executor_server_factory.py
eabatalov: Also it's much less convenient to setup file system sharing between PE and FE containers than just using TCP port forward in Docker.

PR #1425, opened 2025-05-15
Report task outcomes with detailed outcome codes from state reporter

eabatalov:
Also schedule Executor state reports asap when Executor state changes. This reduces latency in the system.

Now task outcomes are reported in Executor State Reporter. This makes it very easy to merge state report RPC and task outcome report RPCs in a follow-up PR.


indexify/src/indexify/executor/grpc/state_reporter.py
```diff
@@ -67,77 +68,132 @@ def __init__(
...
-            # We need to fetch total resources only once, because they are not changing.
-            total_host_resources: HostResources = (
-                await self._host_resources_provider.total_host_resources(self._logger)
-            )
-            total_function_executor_resources: HostResources = (
-                await self._host_resources_provider.total_function_executor_resources(
-                    self._logger
+        # TODO: Remove this code once all task outcomes are reported via ExecutorState message.
+        # Warning: this code needs to go before the state report RPC, because currently Server marks tasks
+        # as failed once we report a terminated FE. As a result task outputs are stored in Server state store.
```
eabatalov, line 182: Typo: are not stored

earhart (approved):
I'm still not completely sold on TaskOutcome -- does the server need more than "Success", "Failed, but it's worth trying this again", and "Failed, don't retry"?

Otherwise, LGTM.

server/proto/executor_api.proto
```diff
@@ -197,11 +187,13 @@ message DesiredExecutorState {
 enum TaskOutcome {
     TASK_OUTCOME_UNKNOWN = 0;
     TASK_OUTCOME_SUCCESS = 1;
-    TASK_OUTCOME_FAILURE = 2;
+    TASK_OUTCOME_FAILURE_CUSTOMER_CODE_ERROR = 2;
```
diptanu, line 190: The right way to do this would be to leave this as Unknown, Success and Failure. And introduce a new field called FailureReason. We can branch the code to a failure handling subroutine if the outcome if failed

diptanu (changes requested):

server/proto/executor_api.proto
eabatalov: Done, this is cleaner indeed

PR #1427, opened 2025-05-16
Implement missing dynamic scheduling features on Executor side

eabatalov:
They are implemented on top of the current overcomplicated Executor code base. I'll refactor all of this soon and address the gap in this PR where I run FE health checks after each task even if a task didn't fail.

Two commits:

```
Check FE health after running each task

If a task failed because FE is unhealthy we learn about this asap
so Server stop schedule tasks on this FE and we reduce overall
latency and rate of failures in the system.

Because this is a hacky implementation we're not checking if the
task failed right now. It's more complecated, will be done later
after refactoring.
    
Now test_function_process_crash.py from Tensorklake SDK tests
passes even without the 10 sec delay.
```

and

```
Report task outcome when it didn't run or finish
    
FE terminates if it failed to startup or if it got unhealthy.
Report the task outcome in this case so it's easier for Server
to reason what happened.
    
Also report task as cancelled if it was removed from desired
Executor state by Server before the task finished.
```


earhart (approved):

earhart (approved):

eabatalov:
There are a few test failures, I notices they happen with this PR more than in the past. They are however racy, most probably due to Server side timings and change when Executor reports task outcomes. These failures need to get investigated and fixed.

PR #1442, opened 2025-06-04
Fix function_allowlist test broken by ZIP graphs

eabatalov:
The test was broken because it was using a global variable that was previously saved by CloudPickle which doesn't happen with ZIP graphs.


eabatalov:
There's a mistake here

PR #1443, opened 2025-06-04
Fix parts of test_function_allowlist broken by ZIP graphs

eabatalov:
The @tensorlake_function can't use global variable anymore cause it's not restored when ZIP graphs are used.


eabatalov:
All the test failures are known issues

PR #1448, opened 2025-06-05
Allow routing to an empty edge list

### LINKED ISSUE #1445: Distinguish "no routes" and "routes not specified" in executor API
We should distinguish between these cases in the executor API, so that we're not special-casing the case where we have an empty route list.

earhart:
## Context

Previous to this change, an empty `next_functions` list in the executor `TaskResult` would cause the server to route outputs to all possible downstream functions specified in the graph definition.  We'd like routers to be able to route to no downstream functions, ending graph execution along that path.

Fixes #1445 

## What

This change adds a new field to the executor `TaskResult`: `routing`.  If present, this field overrides `next_functions`.  If the next functions indicated by a present `routing` is empty, then it's left as empty; we only use the graph definition if `routing` is absent and if `next_functions` is empty, preserving the existing behavior for un-updated clients.

Once all clients have been updated, `next_functions` may be removed from the `TaskResult` proto.

## Testing

Indexify unit tests.

## Contribution Checklist


server/data_model/src/lib.rs
```diff
@@ -648,7 +654,7 @@ pub struct NodeOutput {
     pub compute_fn_name: String,
     pub invocation_id: String,
     pub payloads: Vec<DataPayload>,
-    pub edges: Vec<String>,
+    pub routing: Routing,
```
diptanu, line 657: Do we need this here? If the edges are present then they act as overrides to the static definitions.

server/processor/src/task_creator.rs
```diff
@@ -362,19 +363,13 @@ impl TaskCreator {
...
-        if edges.is_empty() {
-            edges = compute_graph_version
-                .edges
-                .get(&task.compute_fn_name)
-                .cloned()
-                .unwrap_or_default();
-        }
 
-        if edges.is_empty() {
+        let edges = match &node_output.routing {
```
diptanu, line 367: This is pretty much what the same logic as before.

earhart:
> I think you can get away without the server side changes here?

I _think_ we do need the server-side changes in order to be able to supply an empty routing list and have it mean "Don't route to anywhere"; is there another way to do it?

diptanu (approved):

PR #1449, opened 2025-06-06
Miscelenious fixes and getting ready for Retries 

diptanu:
* Fixed a scheduler bug which was not sending the right desired state changes to the executor 
* Added a scheduler integration test which checks wether we are shutting down executors properly when we have to vacuum.
* Added the retry fields in Task and Allocator 
* Added some fields in the HTTP API to expose free resources, function executors in executors, etc for UI to consume.


diptanu:
All the tests related to startup time is failing because of the delay we add because of dynamic scheduling in executor

PR #1465, opened 2025-06-13
Add support for graph and invocation errors

earhart:
## Context

When things go wrong, we want to help users understand what happened; when a task can recognize that the invocation cannot succeed or the graph is fundamentally broken for some reason, we want to latch those conditions as well.

## What

This change adds better error handling throughout the Indexify system, and allows tasks to return errors that break the current graph and stop the current invocation.  (NB Task retries aren't implemented in the server as of this change, so the current invocation will stop anyway, but this change adds enough information that the retry logic will be able to figure out whether it's worth retrying.)

## Testing

Tensorlake tests.

## Contribution Checklist


indexify/src/indexify/proto/executor_api.proto
```diff
@@ -224,6 +226,12 @@ message ResultRouting {
     repeated string next_functions = 1;
 }
 
+message FailureInfo {
```
eabatalov, line 229: What's the use case for this message?

server/src/http_objects.rs
```diff
@@ -631,6 +631,35 @@ impl From<data_model::RuntimeInformation> for RuntimeInformation {
     }
 }
 
+#[derive(Debug, Serialize, Deserialize, ToSchema, Default)]
+pub struct ComputeGraphState {
+    pub status: String,
+    pub failed_invocation_id: Option<String>,
+    pub failed_compute_fn: Option<String>,
+    pub failure_cls: Option<String>,
```
eabatalov, line 639: This is all customer data, do we need to expose it instead of just pointing at the failed task stderr?

server/src/executor_api.rs
```diff
@@ -507,8 +514,38 @@ impl ExecutorAPIService {
                     task.outcome = TaskOutcome::Success;
                 }
                 executor_api_pb::TaskOutcomeCode::Failure => {
-                    // TODO: Handle all the failure reasons.
-                    task.outcome = TaskOutcome::Failure;
+                    let reason = match failure_reason {
```
eabatalov, line 517: It's better to move this mapping into a separate function (typically done using Into of From traits)

indexify/src/indexify/executor/function_executor_controller/run_task.py
```diff
@@ -161,6 +163,15 @@ def _task_output_from_function_executor_response(
         metrics.counters = dict(response.metrics.counters)
         metrics.timers = dict(response.metrics.timers)
 
+    failure_reason = None
+    if not response.success:
+        failure_reason = TaskFailureReason.TASK_FAILURE_REASON_FUNCTION_ERROR
```
eabatalov, line 168: let's stop using .success field and introduce a function that maps failure FE proto scope  to TaskFailureReason. Just to keep things clean.

indexify/src/indexify/proto/executor_api.proto
```diff
@@ -254,6 +261,8 @@ message TaskResult {
     // `next_functions`, so that if `routing` is not present, the
     // graph's routing definitions will always be used.
     ResultRouting routing = 14;
+
+    optional FailureInfo failure = 15;  // Human-readable diagnostic info
```
diptanu, line 265: Can we just re-use the DataPayload (function_outputs)? And use the TaskFailureReason and TaskOutcomeCode to indicate InvocationError? 

That might make it easier to deal with storage too, as it's already handled.

eabatalov:
This PR was replaced by https://github.com/tensorlakeai/indexify/pull/1499

PR #1467, opened 2025-06-13
Save Function Executor diagnostics (startup stdout, startup stderr)

eabatalov:
FE now sends its startup stdout, stderr data payloads to Server
and Server records them.

Currently we don't capture full FE stdout, stderr on Executor side
yet but this can be done later without changing the Executor <->
Server protocol.

Executor can stream logs to the BLOBs it sends in the data payloads
so this change is still usable when we start logs streaming.
We also need to store logs of finished functions/FE startups anyway
even when we start streaming them.

We store the FE startup logs payloads to new persistent state store
FE diagnostics table. FEs are ephemeral but FE Diagnostics is not,
it's stored long term, so e.g. we can show to user FE startup logs
if their allocation failed due to FE startup error.
The table can be extended later to store more long term data about
previously existing FEs.

The FE diagnostics table is keyd by:
namespace|graph_name|function_name|graph_version|fe_id

This allows to prefix search for all FE startup logs per graph
or graph version.

New http path is added to Server to allow FE startup logs download.
Checked that it works:

```
curl http://localhost:8900/namespaces/default/compute_graphs/__main___TestFunctionTimeouts_test_initilization/compute_functions/FunctionThatSleepsForeverOnInitialization/versions/C43FjQ55uja55mJrN80MK/function_executors/hk9c6LUAAYghcw_gSkuIP/startup_logs/stderr
Function initialization exceeded its configured timeout of 2.000 sec.%
```

Also Add ExecutorUpdate message to Executor grpc API
    
This is a mechanism for Executor to report updates to Server so
Server doesn't have to diff these updates out of the existing
ExecutorState. This also allows to not grow ExecutorState for things
that Executor needs to report successfully only once.
    
Currently ExecutorUpdate includes:

* Completed tasks
* FE startup logs

Also pass real allocation ID to executor in desired state instead of allocation key.
This is to assist with end to end tracing and consistent log tags.

Also refactor Executor to pass around TaskAllocation instead of Task
    
This allows to log allocation_id tag correctly in task related
logs and also we don't need to pass allocation_id as a separate
argument in many functions.


server/data_model/src/lib.rs
```diff
@@ -1809,6 +1819,7 @@ pub struct ExecutorMetadata {
     pub addr: String,
     pub labels: HashMap<String, serde_json::Value>,
     pub function_executors: HashMap<FunctionExecutorId, FunctionExecutor>,
+    pub function_executor_diagnostics: HashMap<FunctionExecutorId, FunctionExecutorDiagnostics>,
```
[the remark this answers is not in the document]
eabatalov: function_executor_diagnostics vector is going to be huge, this is why FE diagnostics are stored in a separate state store table. This is why not adding them to ComputeGraphVersion.

server/state_store/src/scanner.rs
```diff
@@ -697,6 +741,20 @@ impl StateReader {
         Ok(allocation)
     }
 
+    pub fn get_function_executor_diagnostics(
```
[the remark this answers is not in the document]
eabatalov: Done, see the full FE diagnostics key in the updated PR description.

server/state_store/src/scanner.rs
```diff
@@ -697,6 +738,18 @@ impl StateReader {
...
```
diptanu, line 741: I don't see this function being used anywhere else. Can we merge this with the *payload function?

server/src/routes/logs.rs
```diff
@@ -64,3 +64,57 @@ pub async fn download_allocation_logs(
...
+            &file,
+        )
+        .map_err(|e| {
+            IndexifyAPIError::internal_error(anyhow!(
+                "failed to download function executor diagnostic payload: {}",
+                e
+            ))
+        })?;
+
+    if payload.is_none() {
```
diptanu, line 101: Can this code can be refactored and shared between the alloc logs download function? Both of these are getting a data payload and then this code which reads data is common.

server/src/executor_api.rs
```diff
@@ -662,17 +728,29 @@ impl ExecutorApi for ExecutorAPIService {
...
 
         let task_results = request.get_ref().task_results.clone();
-        self.handle_task_outcomes(executor_id, task_results)
+        self.handle_task_outcomes(executor_id.clone(), task_results)
             .await
             .map_err(|e| Status::internal(e.to_string()))?;
+
+        let duration_sec = start.elapsed().as_secs_f64();
+        if duration_sec >= 1.0 {
+            warn!(
```
diptanu, line 748: I think you might want some more granular diagnostics here. Get the functions which you are calling to return some sort of diagnostic payload back but I think it's fine to do this as a follow up. At least if we see some warnings here we will know there are issues in the server and not anything related to network.

diptanu (commented):
Looks good overall. Just asked a few clarifying questions.

diptanu (approved):

PR #1468, opened 2025-06-13
Verify that cache-hit task is present in task DB

earhart:
## Context

It's ungood to issue an AllocationOutputsIngestedEvent for a task that doesn't exist.

## What

With this change, in the state machine logic, we load the task from the task db, and bail out (calling error!) if we can't find it.

## Testing

Ran caching tests to make sure the code path was exercised.

## Contribution Checklist


diptanu (approved):
Perfect

PR #1482, opened 2025-06-18
Handle duplicate task updates

diptanu:
There are two different paths for handling failures in tasks - 
1. When an executor reports tasks have failed
2. When an executor reports FEs have failed. 
In some cases executors will report both. 

We were updating task status in both the code paths, which could lead to duplicate allocations. Not updating tasks in the case of FE failures or task failures if the task is set to PENDING (which means unallocated) or COMPLETED (which means it has been deemed failed or successful) by either of the code paths. 

Long term we need to implement CAS based updates to objects where state of objects shouldn't be updated if they were changed by different code paths.


server/state_store/src/in_memory_state.rs
```diff
@@ -973,6 +973,28 @@ impl InMemoryState {
         Ok(changed_executors)
     }
 
+    pub fn fe_resource_for_task(&self, task: &Task) -> Result<FunctionResources> {
```
eabatalov, line 976: This is not quite correct naming. We have FunctionResources which is coming from function definition (includes list of gpu configs). And we have FE Resources which is particular resources allocated on an Executor for a FE (has a single GPU config).

eabatalov (approved):

PR #1492, opened 2025-06-20
Change vacuum logic

diptanu:
Changes the vacuum logic to clean up functions which don't have any pending tasks even if invocations are active. This will make slower functions scale out and get resources from idle functions and thus increase throughput. 

The downside is that when new invocations are created we will wait for slower functions running tasks from older invocations to finish. 

The future improvement is to limit the % of cluster we allocate to any functions such that slow functions don't eat up the whole cluster. 

The other change is - if we can't free up enough resources to accommodate a function we don't free anything since it's not productive to terminate functions unless they can be used for something else.

### DIFF
--- a/server/state_store/src/in_memory_state.rs
+++ b/server/state_store/src/in_memory_state.rs
@@ -1170,8 +1170,6 @@ impl InMemoryState {
         &self,
         fe_resource: &FunctionResources,
     ) -> Result<Vec<Box<FunctionExecutorServerMetadata>>> {
-        let mut function_executors_to_remove: Vec<Box<FunctionExecutorServerMetadata>> = Vec::new();
-
         // For each executor in the system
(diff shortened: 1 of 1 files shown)


server/state_store/src/in_memory_state.rs
```diff
@@ -1201,8 +1206,6 @@ impl InMemoryState {
 
                 let fe = &fe_metadata.function_executor;
                 let Some(executor) = self.executors.get(executor_id) else {
```
eabatalov, line 1208: This line looks like a bug to me. What is this about?

The previous comment that was removed in this PR:

> // If the executor is not found, we can remove the FE
>                    // since it's not running any more

Suggests that we need to check for None here?

eabatalov (approved):

PR #1499, opened 2025-06-25
Invocation errors and refactoring of Function Executor and Executor protocols

eabatalov:
If InvocationError is raised by customer function code then
then its current invocation fails. InvocationError requires
a string message argument that can be retrieved via Server API
and Tensrolake CLI.

Use DataPayload for InvocationError message/payload
This makes it much more flexible wrt the size of the InvocationError
message and lets SDK own what can be stored there.
Users can now add any size of string message to InvocationError
and in the future we can easily allow DataModels there and etc
if needed.
    
This also removes the need to store the InvocationError message
in Server state store. All we need is just put DataPayload there.

Also refactor FunctionExecutor and Executor grpc protocols
because we don't need backward compatibility right now - it's
a good window to do this.

The CI tests will fail until we deploy the required changes
in Indexify (due to FE and Executor grpc protocol changes).

Now Function Executor always provides next_functions in its run task
response. The next functions are either produced by the function
or edges from the graph definition object. This allows FE to fully
control next_functions and server just following with the decisions.
This moves the next_functions list ownership to a single place (FE/SDK)
and also simplifies the FE and Executor protocols.


server/processor/src/task_creator.rs
```diff
@@ -439,11 +449,19 @@ impl TaskCreator {
 
         // Create new tasks for the edges of the node of the current task.
         let mut new_reduction_tasks = vec![];
-        for edge in edges.unwrap().iter() {
-            let edge_compute_node = compute_graph_version
-                .nodes
-                .get(edge)
-                .ok_or(anyhow!("compute node not found: {:?}", edge))?;
+        for edge in node_output.next_functions.iter() {
```
j3m7, line 452: Is this where we would validate and enforce limits on dynamic routing? e.g. limits for reducer functions and avoiding cyclical routes?

server/data_model/src/lib.rs
```diff
@@ -991,6 +1046,52 @@ impl Display for TaskOutcome {
...
+}
+
+impl TaskFailureReason {
+    pub fn is_retriable(&self) -> bool {
+        matches!(
+            self,
+            TaskFailureReason::InternalError
+            | TaskFailureReason::FunctionError
+            | TaskFailureReason::FunctionTimeout
+            // | TaskFailureReason::InvocationError - fails the invocation and the task
```
diptanu, line 1088: Why is this commented out?

server/data_model/src/lib.rs
```diff
@@ -991,6 +1046,52 @@ impl Display for TaskOutcome {
     }
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub enum TaskFailureReason {
```
diptanu, line 1050: Can you add some comments on these fields please? I forget what TaskCancelled and FunctionExecutorTerminated was for. Need to understand since these are being used in the allocation is retriable method

server/processor/src/task_creator.rs
eabatalov: Yes. This was never implement though.

server/data_model/src/lib.rs
eabatalov: Yep, np.

eabatalov: Added the comments here and also commented all the enums in the executor api proto file. It's important to have clear comments there cause it's a contract between executor and server.

server/data_model/src/lib.rs
eabatalov: Now it is:

```
// Only InvocationError is not retriable because it fails the invocation
// permanently
```

PR #1510, opened 2025-06-27
Don't count platform issues against task attempts

earhart:
## Context

Tasks may be cancelled by the system for reasons outside the task's control; these shouldn't count against the task's retry attempts.

Fixes #1509

## What

Add a function to determine whether a `TaskFailureReason` should count against task retry attempts, and use it.

## Testing

No good tests for this yet -- we think we may have seen this come up when the desired state stream is disrupted, but if so it's tricky to reproduce.

## Contribution Checklist

### DIFF
--- a/server/data_model/src/lib.rs
+++ b/server/data_model/src/lib.rs
@@ -1104,6 +1104,15 @@ impl TaskFailureReason {
                 TaskFailureReason::FunctionExecutorTerminated
         )
     }
+
+    pub fn should_count_against_task_retry_attempts(&self) -> bool {
+        // Platform/infrastructure failures shouldn't count against retry attempts
+        // since they're not legitimate task execution failures
+        matches!(
+            self,
+            TaskFailureReason::FunctionError | TaskFailureReason::FunctionTimeout
+        )
+    }
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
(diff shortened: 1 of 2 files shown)


server/data_model/src/lib.rs
```diff
@@ -1104,6 +1104,15 @@ impl TaskFailureReason {
                 TaskFailureReason::FunctionExecutorTerminated
         )
     }
+
+    pub fn should_count_against_task_retry_attempts(&self) -> bool {
+        // Platform/infrastructure failures shouldn't count against retry attempts
+        // since they're not legitimate task execution failures
+        matches!(
+            self,
+            TaskFailureReason::FunctionError | TaskFailureReason::FunctionTimeout
```
diptanu, line 1113: @earhart Un-related to this PR. TaskFailureReason should be `Option<TaskFailureReason>` on the Allocation object since tasks don't fail always. You think you can squeeze that in this PR? Doing it in a follow up is also fine.

eabatalov: There's a corner case here. FunctionExecutorTerminated can be due to customer code error or timeout on FE startup. In this case we should imo increment the attempts counter.

eabatalov: I'm not correct.  We don't know if this particular task was the cause of the FE termination. If we increment the counter here for FE terminations then a single FE termination will increment the counter for all tasks assigned to it which is not correct.

PR #1511, opened 2025-06-28
Refactor graph processor

earhart:
## Context

task_allocator.rs was doing two things: task allocation and function executor management.  That's one thing too many.

## What

This change breaks the function executor management out to a separate file, hoists the `InMemoryState` locking up to `graph_processor.rs` (eliminating some potential race conditions), and replaces some`sync::RwLock` and `sync::Mutex` uses with their `tokio` equivalents.

## Testing

Standard PR regression tests.

## Contribution Checklist


server/processor/src/function_executor_manager.rs
```diff
@@ -0,0 +1,493 @@
...
+    TaskStatus,
+};
+use rand::seq::IndexedRandom;
+use state_store::{
+    in_memory_state::InMemoryState,
+    requests::{RequestPayload, SchedulerUpdateRequest},
+};
+use tracing::{error, info, info_span};
+
+pub struct FunctionExecutorManager {
```
diptanu, line 25: Can we write some comments here on what this does, so that its easier for other developers to understand what they should put in here and what not.

server/processor/src/function_executor_manager.rs
```diff
@@ -0,0 +1,493 @@
...
+            in_memory_state,
+            &mut executor_server_metadata,
+            &function_executors_to_remove,
+        )
+    }
+
+    /// Selects or creates a function executor for the given task
+    /// Returns the allocation target and any scheduler updates
+    #[tracing::instrument(skip(self, in_memory_state, task))]
+    pub fn select_or_create_function_executor(
```
diptanu, line 398: What do you think about leaving the responsibility of selecting the FE in task allocator? And then changing this method to get some candidate FEs where the task can be allocated. 

I am thinking the algorithm for selecting the FE is going to get complex over time as we would be ranking the FEs based on how much tasks are already allocated on them, any observed latency overhead because of where they are placed, etc. That type of logic can sit on top of the FE manager which simply adds/removes/filters FEs.

server/processor/src/function_executor_manager.rs
earhart: Done

diptanu (approved):

PR #1513, opened 2025-06-30
Fold outcomes and failure reasons

earhart:
## Context

The representation for outcomes and failure reasons has them as separate fields, which isn't quite right -- we only have a failure reason when there's a failure outcome.

## What

This change modifies the failure outcome enum values to include the failure reason, and removes the separate failure reason.

## Testing

`cargo test`

## Contribution Checklist


server/data_model/src/lib.rs
```diff
@@ -751,23 +748,25 @@ impl InvocationPayloadBuilder {
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub enum GraphInvocationOutcome {
-    Undefined,
+    InProgress,
```
eabatalov, line 751: This doesn't look like an outcome of invocation. What does it mean?

server/data_model/src/lib.rs
```diff
@@ -751,23 +748,25 @@ impl InvocationPayloadBuilder {
...
     Success,
-    Failure,
+    Failure(GraphInvocationFailureReason),
 }
 
 impl Default for GraphInvocationOutcome {
     fn default() -> Self {
-        Self::Undefined
+        Self::InProgress
```
eabatalov, line 758: I'd expected something like "not started" or unknown as a default value. Why is it in progress?

eabatalov (changes requested):
I'm not sure why we add in progress as outcome and why it's default value.
Also I think that we need to fix bugs right now instead of refactoring with a chance to add more bugs.

server/data_model/src/lib.rs
earhart: "We don't know what the outcome is yet"

server/data_model/src/lib.rs
```diff
@@ -1040,21 +1037,21 @@ impl ReduceTask {
...
 }
 
 impl Display for TaskOutcome {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         let str_val = match self {
-            TaskOutcome::Success => "Success",
-            TaskOutcome::Failure => "Failure",
             TaskOutcome::Unknown => "Unknown",
+            TaskOutcome::Success => "Success",
+            TaskOutcome::Failure(_) => "Failure",
```
eabatalov, line 1054: I'd add the failure details in the display

server/processor/src/task_creator.rs
```diff
@@ -373,8 +369,10 @@ impl TaskCreator {
                     });
                 }
             }
-            invocation_ctx.failure_reason = task.failure_reason.into();
-            invocation_ctx.complete_invocation(true, GraphInvocationOutcome::Failure);
+            invocation_ctx.complete_invocation(
+                true,
+                GraphInvocationOutcome::Failure(failure_reason.clone().into()),
```
eabatalov, line 374: I'd add a convertor for TaskOutcome -> GraphInvocationOutcome so we can just
task.outcome.into() here

server/src/http_objects.rs
```diff
@@ -880,8 +880,11 @@ impl From<GraphInvocationCtx> for Invocation {
         Self {
             id: value.invocation_id.to_string(),
             completed: value.completed,
-            outcome: value.outcome.into(),
-            failure_reason: value.failure_reason.into(),
+            outcome: value.outcome.clone().into(),
+            failure_reason: match &value.outcome {
+                GraphInvocationOutcome::Failure(reason) => reason.clone().into(),
+                _ => GraphInvocationFailureReason::Unknown.into(),
```
eabatalov, line 886: This doesn't look like a good refactoring tbh

eabatalov (changes requested):
The test are failing.

server/data_model/src/lib.rs
earhart: Nice idea -- done

server/data_model/src/lib.rs
earhart: Changed to "Unknown"

PR #1526, opened 2025-07-01
Exercise retry test edge cases

earhart:
## Context

There were some edge cases with retryable tasks and errors that don't consume retry attempts in the face of tasks that start with no retry attempts available.

## What

This change fixes those edge cases.

## Testing

Added tests to cover these edge cases.

## Contribution Checklist


server/data_model/src/lib.rs
```diff
@@ -549,11 +549,11 @@ impl ComputeGraphVersion {
             .collect()
     }
 
-    pub fn should_retry_task(&self, task: &Task) -> bool {
+    pub fn should_retry_task(&self, task: &Task, uses_attempt: bool) -> bool {
         let Some(node) = self.nodes.get(&task.compute_fn_name) else {
             return false;
         };
-        task.attempt_number < node.retry_policy.max_retries
+        task.attempt_number < node.retry_policy.max_retries || !uses_attempt
```
eabatalov, line 556: I think we need to split retry counters for user defined policy driven retries and for "internal error/infra" driven retries. Then we won't need to pass this uses_attempt bool and call should retry task if we're on "internal error/infra" driven retry path.

This is extra refactoring and up for discussion, so not a blocker for approving this PR.

eabatalov: As discussed for now we'll use the same retry counter for all retriable failures and implement it better later.

Issue #1520, opened 2025-06-30
Retry Function Executor failures


eabatalov:
Currently this test fails cleanely https://github.com/tensorlakeai/indexify/blob/main/indexify/tests/features/test_function_retries.py#L89
Expectation: the function succeeds once it's possible to create a Function Executor.

We need a Server integration test to repro this and fix.

We need to handle all FE failure reasons both these happening during FE startup and while FE is running.

Once done please remove the `@unittest.skip("Function Executor startup retries feature has a bug")` attribute from the Python integration test.

eabatalov:
Added more integration tests to catch errors in implementation https://github.com/tensorlakeai/indexify/pull/1521

eabatalov:
I think we addressed this. Will see if there are any bugs after Dev and prod testing.

PR #1530, opened 2025-07-02
Handle FE termination reasons

earhart:
## Context

We're not handling FE termination reasons correctly (retrying the task, and docking it a retry attempt when we're not sure it's our fault).

## What

We've talked through which FE termination reasons should dock retry attempts; this PR implements what we've discussed, mostly fixing #1520 (although we also have to get the retry timeouts correct, which this PR does not do yet).

## Testing

Added tests for all the various FE termination reasons

## Contribution Checklist

### DIFF
--- a/indexify/tests/features/test_function_retries.py
+++ b/indexify/tests/features/test_function_retries.py
@@ -85,8 +85,11 @@ class TestFunctionRetries(unittest.TestCase):
 
 
 class FunctionWithFailingConstructor(TensorlakeCompute):
-    name = "FunctionWithFailingConstructor"
     FILE_PATH = "/tmp/FunctionWithFailingConstructor_fail"
+    MAX_RETRIES = 3
+
+    name = "FunctionWithFailingConstructor"
+    retries = Retries(max_retries=MAX_RETRIES, max_delay=1.0)
 
     def __init__(self):
         super().__init__()
@@ -109,9 +112,12 @@ class FunctionWithFailingConstructor(TensorlakeCompute):
 
 
 class FunctionWithTimingOutConstructor(TensorlakeCompute):
+    FILE_PATH = "/tmp/FunctionWithTimingOutConstructor_timeout"
+    MAX_RETRIES = 3
+
     name = "FunctionWithTimingOutConstructor"
     timeout = 1
-    FILE_PATH = "/tmp/FunctionWithTimingOutConstructor_timeout"
+    retries = Retries(max_retries=MAX_RETRIES, max_delay=1.0)
 
     def __init__(self):
         super().__init__()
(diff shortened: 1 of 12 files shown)


indexify/tests/features/test_function_retries.py
```diff
@@ -133,7 +139,92 @@ def timeout_constructor(cls):
...
+    name = "FunctionWithRetryCountingTimeoutConstructor"
+    timeout = 1
+    retries = Retries(max_retries=MAX_RETRIES, max_delay=1.0)
+
+    def __init__(self):
+        super().__init__()
+        # Read current retry count, increment it, and write back
+        try:
+            with open(self.COUNTER_FILE_PATH, "r") as f:
+                retry_count = int(f.read().strip())
```
eabatalov, line 197: Call self.get_retry_count() here?

indexify/tests/features/test_function_retries.py
```diff
@@ -170,5 +261,63 @@ def untimeout_constructor_with_delay():
...
+
+        # Run the graph - it should succeed after exactly MAX_RETRIES + 1 attempts
+        invocation_id = graph.run(block_until_done=True, x=1)
+
+        # Verify the output
+        outputs = graph.output(invocation_id, FunctionWithRetryCountingConstructor.name)
+        self.assertEqual(outputs, ["success after retries"])
+
+        # Verify the retry count matches expected retries + initial attempt
+        retry_count = FunctionWithRetryCountingConstructor.get_retry_count()
```
eabatalov, line 284: Nit, these functions are not counting retries, they count contructor runs.

server/data_model/src/lib.rs
```diff
@@ -1708,13 +1707,32 @@ pub enum FunctionExecutorState {
...
+    FunctionError,
+    FunctionTimeout,
+    FunctionCancelled,
+}
+
+impl FunctionExecutorTerminationReason {
+    pub fn should_count_against_task_retry_attempts(self) -> bool {
+        matches!(
+            self,
+            Self::Unknown |
```
eabatalov, line 1726: As was discussed yesterday on Slack ExecutorShutdown and DesiredStateRemoved is not set by Executor in reported FE state and I'm going to remove it. As a result are we using Unknown on Executor shutdown and normal FE removal by Server? In this case regular infra operations like executor shutdown and FE removal by Server will count against retry attempts.

What do you think?

server/processor/src/function_executor_manager.rs
```diff
@@ -303,36 +304,34 @@ impl FunctionExecutorManager {
...
-                        FunctionExecutorTerminationReason::DesiredStateRemoved,
-                    ) => {
-                        task.status = TaskStatus::Pending;
-                        task.attempt_number = task.attempt_number + 1;
-                    }
+
+                if_chain! {
+                        if let Ok(compute_graph_version) = in_memory_state.get_existing_compute_graph_version(&task);
+                        if let Some(max_retries) = compute_graph_version.task_max_retries(&task);
+                        if matches!(fe.state, FunctionExecutorState::Terminated(termination_reason) if termination_reason.should_count_against_task_retry_attempts());
```
eabatalov, line 311: We'd need to set Removed From Desired state or something like that to termination_reason when we see FE is not getting reported by Executor anymore.

Also we'd need to set ExecutorShutdown to termination_reason when we detect that Executor is shutdown.

What do you think?

server/processor/src/function_executor_manager.rs
```diff
@@ -303,36 +304,34 @@ impl FunctionExecutorManager {
...
+                                // attempts.
+                                if task.attempt_number < max_retries {
+                                    // The task can be retried.
+                                    task.attempt_number += 1;
+                                    task.status = TaskStatus::Pending;
+                                } else {
+                                    // The task cannot be retried.
+                                    task.status = TaskStatus::Completed;
+                                    task.outcome =
+                                        TaskOutcome::Failure(TaskFailureReason::FunctionError);
```
eabatalov, line 323: iirc for failed tasks we are setting task.outcome = alloc.outcome.into(). So the task outcome is derived from its last alloc outcome.

The same works for failed invocations:
invocation.outcome = last_task.outcome.into()

server/processor/src/task_creator.rs
```diff
@@ -200,16 +182,18 @@ impl TaskCreator {
 
             if let TaskOutcome::Failure(failure_reason) = &allocation.outcome {
                 let uses_attempt = failure_reason.should_count_against_task_retry_attempts();
```
eabatalov, line 184: Nit, can this bunch of if statements be simplified? Just for readability?

server/processor/src/function_executor_manager.rs
```diff
@@ -303,36 +304,34 @@ impl FunctionExecutorManager {
...
+                                    failed_tasks += 1;
+                                }
+                        }
+                else {
+                            // Either we weren't able to get the info we
```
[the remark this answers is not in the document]
eabatalov: Though it seems like this pattern is used in many places right now, so not a big deal.

server/src/testing.rs
```diff
@@ -345,14 +345,14 @@ impl TestExecutor<'_> {
         Ok(())
     }
 
-    pub async fn mark_function_executors_as_running(&self) -> Result<()> {
+    pub async fn mark_function_executors(&self, fes: FunctionExecutorState) -> Result<()> {
```
eabatalov, line 348: fes look like a collection of FEs, let's use fe_state name

server/processor/src/function_executor_manager.rs
eabatalov: I think we need to map FE failure_reason to TaskFailure reason.
There's a clear mapping there. Also we need to only consume
retries for allocations that caused the FE termination
and just realloc the rest.

server/data_model/src/lib.rs
eabatalov: Checked existing code that we're not relying on ExecutorShutdown and DesiredStateRemoved  delivered by Executor. So not a blocker for this PR.

server/processor/src/function_executor_manager.rs
earhart: Heh, I love AI hallucinations.  :-)
`MaxRetriesExceeded` - it's a little odd in the current failure reason paradigm, since presumably it's part of the closed set of reasons that might be reported by the executor.  I'm guessing that it shouldn't consume a retry attempt?  :-). Other than that, I kinda like it -- there's some nice user clarity there.  Done.

I'm thinking we'll add the alloc list in a subsequent PR.

server/processor/src/function_executor_manager.rs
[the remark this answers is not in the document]
earhart: If we can't get the info we need, task creation can deal with it.

server/processor/src/task_creator.rs
earhart: Done

server/src/testing.rs
earhart: Done

server/data_model/src/lib.rs
```diff
@@ -1125,8 +1128,10 @@ impl TaskFailureReason {
 
     pub fn should_count_against_task_retry_attempts(self) -> bool {
         // Explicit platform decisions and provable infrastructure
-        // failures don't count against retry attempts.  Everything
-        // else counts against retry attempts.
+        // failures don't count against retry attempts; we also don't
+        // bother treating MaxRetriesExceeded as counting against
+        // retry attempts.  Everything else counts against retry
+        // attempts.
         matches!(
```
eabatalov, line 1135: fyi not adding Unknown here seems correct because Unknown would be used when FE is not terminated by Executor decision. But this doesn't seem documented anywhere in the comments which is not great, cause hard to figure out.

eabatalov: I think this is right to not add unknown because in task outcome case we can only get unknown from executor which is an inconsistent protocol error, we shouldn't really handle it anywhere.

eabatalov (approved):

PR #1540, opened 2025-07-02
Only charge retries to allocs indicated by executor

### LINKED ISSUE #1539: Only charge retries to tasks indicated by executor

earhart:
## Context

When an FE terminates, we only want to charge retry attempts against allocs indicated by the executor.

Fixes #1539 

## What

This change plumbs through the alloc list from the executor, using it as a filter for retries.

## Testing

`cargo test --workspace -- -- test-threads 1`

## Contribution Checklist

### DIFF
--- a/server/data_model/src/lib.rs
+++ b/server/data_model/src/lib.rs
@@ -1668,17 +1668,7 @@ impl Default for FunctionExecutorId {
 }
 
 #[derive(
-    Debug,
-    Clone,
-    Copy,
-    Serialize,
-    Deserialize,
-    PartialEq,
-    Default,
-    strum::AsRefStr,
-    Display,
-    Eq,
-    Hash,
+    Debug, Clone, Serialize, Deserialize, PartialEq, Default, strum::AsRefStr, Display, Eq, Hash,
 )]
 pub enum FunctionExecutorState {
     #[default]
@@ -1688,7 +1678,10 @@ pub enum FunctionExecutorState {
     // Function Executor is running and ready to accept tasks.
     Running,
     // Function Executor is terminated, all resources are freed.
-    Terminated(FunctionExecutorTerminationReason),
+    Terminated {
+        reason: FunctionExecutorTerminationReason,
+        allocs: Vec<String>,
+    },
 }
 
 #[derive(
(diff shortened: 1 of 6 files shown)


server/data_model/src/lib.rs
```diff
@@ -1688,7 +1678,10 @@ pub enum FunctionExecutorState {
     // Function Executor is running and ready to accept tasks.
     Running,
     // Function Executor is terminated, all resources are freed.
-    Terminated(FunctionExecutorTerminationReason),
+    Terminated {
+        reason: FunctionExecutorTerminationReason,
+        allocs: Vec<String>,
```
eabatalov, line 1683: Nit. This field might benefit from a comment or are more explicit name that these are alloc IDs and these are causes of FE termination.

server/processor/src/function_executor_manager.rs
```diff
@@ -306,33 +307,34 @@ impl FunctionExecutorManager {
...
+                    then {
+                        // The alloc's task should be retried iff it has remaining retry
+                        // attempts.
+                        if task.attempt_number < max_retries {
+                            // The task can be retried.
+                            task.attempt_number += 1;
                             task.status = TaskStatus::Pending;
-                }
+                        } else {
+                            // The task cannot be retried.
```
eabatalov, line 323: I'm not getting this part. Why do we fail alloc that didn't cause FE termination?

I think if the alloc didn't cause FE termination then we shouldn't charge retries for it and just put it to pending again.

earhart: That's what we're doing, I think?  If _all_ of these `if-` checks succeed, we wind up in the `then` block, so we've been told to blame this alloc -- charging it a retry attempt and completing it if it doesn't have any.  The `else` block runs if any of the conditions fail -- so if the alloc isn't in the blame list, it gets set to `Pending`.

eabatalov: True, thank you. This is correct.

server/processor/src/function_executor_manager.rs
```diff
@@ -306,33 +307,34 @@ impl FunctionExecutorManager {
...
+                            task.status = TaskStatus::Completed;
+                            task.outcome =
+                                TaskOutcome::Failure(TaskFailureReason::FunctionError);
```
eabatalov, line 326: This is out of scope for this PR but this is still not correct. For consistency with other code paths we need to derive task.outcome from FE termination reason that caused the failure.

PR #1541, opened 2025-07-03
Log FE state in termination

earhart:
## Context

It came up in debugging that we didn't have info logs for the FE termination reason.

## What

Add the FE termination reason to the logs, adding a separate termination reason when the entire executor goes away.

## Testing

Also, ran a test that threw an exception from a constructor, and verified that we see the correct info logs.

## Contribution Checklist

### DIFF
--- a/server/processor/src/function_executor_manager.rs
+++ b/server/processor/src/function_executor_manager.rs
@@ -284,11 +284,12 @@ impl FunctionExecutorManager {
         // Handle allocations for FEs to be removed and update tasks
         for fe in function_executors_to_remove {
             info!(
-                namespace = fe.namespace,
-                compute_graph = fe.compute_graph_name,
-                compute_fn = fe.compute_fn_name,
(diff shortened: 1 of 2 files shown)


diptanu (approved):

server/processor/src/function_executor_manager.rs
```diff
@@ -284,11 +284,12 @@ impl FunctionExecutorManager {
         // Handle allocations for FEs to be removed and update tasks
         for fe in function_executors_to_remove {
             info!(
-                namespace = fe.namespace,
-                compute_graph = fe.compute_graph_name,
-                compute_fn = fe.compute_fn_name,
-                fn_executor_id = fe.id.get(),
-                "removing function executor from executor",
+                fe.namespace,
```
eabatalov, line 287: Does this preserve consistent log tags that we need for troubleshooting across Server and Executor?
See https://github.com/tensorlakeai/internal-docs/blob/main/architecture/tags.md

PR #1542, opened 2025-07-03
fix: task ingest

diptanu:
We shouldn't be updating tasks when we ingest allocation output. By the time the tasks are getting updated, they might be getting updated by the scheduler as well. 

Scenario - 
1. FE fails, and we get FE failure update from the executor. The scheduler updates the task outcome and state
2. If we get a task result simultaenously, if it updates the task also it can collide with (1)

### DIFF
--- a/server/src/executor_api.rs
+++ b/server/src/executor_api.rs
@@ -691,7 +691,6 @@ impl ExecutorAPIService {
                 compute_graph: compute_graph.compute_graph_name.clone(),
                 compute_fn: compute_fn.to_string(),
                 invocation_id: invocation_id.to_string(),
-                task: task.clone(),
                 node_output,
                 executor_id: executor_id.clone(),
                 allocation,
--- a/server/state_store/src/in_memory_state.rs
+++ b/server/state_store/src/in_memory_state.rs
@@ -627,19 +627,6 @@ impl InMemoryState {
                     .insert(req.ctx.key(), Box::new(req.ctx.clone()));
             }
             RequestPayload::IngestTaskOutputs(req) => {
-                // Update task
-                let invocation_ctx_key = GraphInvocationCtx::key_from(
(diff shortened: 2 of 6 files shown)


server/src/testing.rs
```diff
@@ -108,12 +109,13 @@ impl TestService {
...
     pub async fn create_executor(&self, executor: ExecutorMetadata) -> Result<TestExecutor> {
-        let e = TestExecutor {
+        let mut e = TestExecutor {
             executor_id: executor.id.clone(),
+            executor_metadata: executor.clone(),
             test_service: self,
         };
 
-        e.heartbeat(executor).await?;
+        e.heartbeat(executor.clone()).await?;
```
earhart, line 118: Why the `.clone()`?

server/src/testing.rs
```diff
@@ -316,21 +318,41 @@ impl FinalizeTaskArgs {
...
-        self.test_service
+    pub async fn heartbeat(&mut self, executor: ExecutorMetadata) -> Result<()> {
+        let executor_state_changed = self
+            .test_service
             .service
             .executor_manager
-            .heartbeat(executor, function_executor_diagnostics)
+            .heartbeat(&executor)
+            .await?;
+        self.executor_metadata = executor.clone();
```
earhart, line 333: Likewise, do you need the `.clone()`?

server/src/testing.rs
```diff
@@ -340,12 +362,15 @@ impl TestExecutor<'_> {
 
         // Update state hash and send heartbeat
         executor.state_hash = nanoid!();
-        self.heartbeat(executor).await?;
+        self.heartbeat(executor.clone()).await?;
```
earhart, line 365: Another `.clone()`?

earhart (approved):

earhart (approved):

PR #1543, opened 2025-07-03
Use std logging tags for FE term logging

earhart:
## Context

We have [standard logging tags](https://github.com/tensorlakeai/internal-docs/blob/main/architecture/tags.md); we should use them.  :-)

## What

Change to use the standard logging tags where available.

## Testing

Ran a local test; verified that the info message uses the standard fields.

## Contribution Checklist

### DIFF
--- a/server/processor/src/function_executor_manager.rs
+++ b/server/processor/src/function_executor_manager.rs
@@ -284,11 +284,11 @@ impl FunctionExecutorManager {
         // Handle allocations for FEs to be removed and update tasks
         for fe in function_executors_to_remove {
             info!(
-                fe.namespace,
-                fe.compute_graph = fe.compute_graph_name,
-                fe.compute_fn = fe.compute_fn_name,
-                fe.executor_id = fe.id.get(),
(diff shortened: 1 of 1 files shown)


eabatalov (approved):

server/processor/src/function_executor_manager.rs
```diff
@@ -284,11 +284,11 @@ impl FunctionExecutorManager {
         // Handle allocations for FEs to be removed and update tasks
         for fe in function_executors_to_remove {
             info!(
-                fe.namespace,
-                fe.compute_graph = fe.compute_graph_name,
-                fe.compute_fn = fe.compute_fn_name,
-                fe.executor_id = fe.id.get(),
-                ?fe.state,
+                namespace = fe.namespace,
+                graph = fe.compute_graph_name,
```
diptanu, line 288: I think we are using a different tag for graph. Please see other places we are logging info logs.

diptanu (changes requested):

diptanu (approved):

PR #1579, opened 2025-07-10
Add processor debug log support

### LINKED ISSUE #1576: Save scheduler events to a file for further investigation
In https://github.com/tensorlakeai/indexify/commit/64617ee6f5ecf33722db5a0665b795b660f930a8 we switched lots of scheduler event logs to debug log level. This makes it not possible to investigate why scheduler did a certain decision in prod at certain point.
Let's create a log file on Server startup for scheduler events and log them there so the events are available to us. The writes to the file are append only so we can rotate it (later if needed). The logging format is the same structured logging format we're currently using.
Let's also add a Server configuration value for max log file size (e.g. 10 MB by default) and run a thread in Server that periodically checks the size and truncates the file so it doesn't grow without boundaries.

earhart:
## Context

We recently turned down the verbosity of the logs we're sending to the collector, but we still want to save detailed scheduler logs locally for debugging purposes.

Fixes #1576 

## What

This change adds a processor module debug log option, with daily rotation.

## Testing

Ran the server; verified that it does not log below INFO to its stdout, but that it does log events DEBUG and lower in the processor module to the requested log file.

## Contribution Checklist

### DIFF
--- a/server/sample_config.yaml
+++ b/server/sample_config.yaml
@@ -1,8 +1,5 @@
 state_store_path: indexify_server_state 
 listen_addr: 0.0.0.0:8900
-blob_storage:
-  path: "s3://indexifyblobs"
-  dynamodb_table: kvs_cas_table 
 
 telemetry:
   enable_tracing: false
@@ -12,3 +9,5 @@ telemetry:
   endpoint: "http://localhost:4317"
   # Metrics export interval in seconds (defaults to 10 if not specified)
   metrics_interval: 5
+  # Optional path to write processor debug logs to a rotating file
+  processor_debug_log_file: "/tmp/indexify/processor.log"
--- a/server/src/main.rs
+++ b/server/src/main.rs
@@ -1,4 +1,7 @@
-use std::path::PathBuf;
+use std::{
+    ffi::OsStr,
+    path::{Path, PathBuf},
+};
 
 use anyhow::{Ok, Result};
 use clap::Parser;
@@ -7,7 +10,8 @@ use opentelemetry::global;
 use opentelemetry_otlp::{SpanExporter, WithExportConfig};
 use opentelemetry_sdk::trace::{SdkTracerProvider, TracerProviderBuilder};
 use service::Service;
-use tracing::error;
+use tracing::{error, Level};
+use tracing_appender;
 use tracing_subscriber::{
     fmt::{
         self,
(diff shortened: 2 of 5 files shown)


server/src/main.rs
```diff
@@ -77,13 +81,47 @@ where
...
+    );
+    let (non_blocking, _guard) = tracing_appender::non_blocking(file_appender);
+    let file_layer = tracing_subscriber::fmt::layer()
+        .with_writer(non_blocking)
+        .with_ansi(false) // No ANSI colors for file output
+        .with_filter(tracing_subscriber::filter::FilterFn::new(|metadata| {
+            let Some(path) = metadata.module_path() else {
+                return false;
+            };
+            *metadata.level() <= Level::DEBUG && path.starts_with("processor::")
```
eabatalov, line 108: Let's make it easy to add new targets besides the "processor::". Also is their any way we can prevent renaming of the processor module resulting in logs not being added to the file?

server/src/main.rs
```diff
@@ -77,13 +81,47 @@ where
...
+        return Box::new(tracing_subscriber::layer::Identity::new());
+    };
+
+    let file_appender = tracing_appender::rolling::daily(
+        Path::new(log_file_path).parent().unwrap_or(Path::new(".")),
+        Path::new(log_file_path)
+            .file_name()
+            .unwrap_or(OsStr::new("processor.log")),
+    );
+    let (non_blocking, _guard) = tracing_appender::non_blocking(file_appender);
```
eabatalov, line 100: Non blocking means buffered? Do we know what's the buffer size?

server/src/main.rs
earhart: Sure - I'll switch to using a scheduler tag

server/src/main.rs
```diff
@@ -77,13 +81,47 @@ where
     Box::new(tracing_subscriber::fmt::layer().compact())
 }
 
+fn get_processor_debug_layer<S>(config: &ServerConfig) -> Box<dyn Layer<S> + Send + Sync + 'static>
```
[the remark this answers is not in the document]
earhart: Yeah, I kinda miss C++'s "auto" for things like this.  We could elide these traits by inlining the functions and letting type inference do the right thing...

server/processor/src/function_executor_manager.rs
```diff
@@ -50,6 +50,7 @@ impl FunctionExecutorManager {
             in_memory_state.vacuum_function_executors_candidates(fe_resource)?;
 
         debug!(
+            target: "scheduler",
```
eabatalov, line 53: What if we don't use hardcoded string here and instead go with: `local_log: SCHEDULER_EVENT`.
It's more clear to me that this is a scheduler event and I know that we want scheduler events to be in local log and that it actually goes to local log. Just for me as a code reader.

server/sample_config.yaml
```diff
@@ -12,3 +9,8 @@ telemetry:
   endpoint: "http://localhost:4317"
   # Metrics export interval in seconds (defaults to 10 if not specified)
   metrics_interval: 5
+  # Optional path to write local logs to a rotating file
+  local_log_file: "/tmp/indexify/local.log"
+  # List of targets and their log levels for local logging
+  local_log_targets:
```
eabatalov, line 15: I'd remove this from dynamic runtime configuration and wire it all in code tbh. So we get this all setup in compile time and we only configure if we enable local log or not and the log file path.
Also as we know have `local_log` or `target` tag do we really need to care about log level? Let's just put everything properly tagged to local log regardless of log level?

eabatalov (approved):
I think the UX could be made more clear, see the tag rename suggestion. Also it seems like filetering by log level is excessive in local log.
I like the local log term. Really clear.

server/src/main.rs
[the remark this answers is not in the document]
diptanu: I think it's fine to use these features IMO within reason :) I learn a thing or two when someone uses something that makes performance better.

server/processor/src/function_executor_manager.rs
earhart: I think we can use a constant, but I don't think this code should know that the logs are going to a local file; that's an orthogonal decision. (Also, info! logs will *also* be going to the telemetry collector :-)

eabatalov: Maybe then `target=SCHEDULER_EVENT`. This makes it clear that this not a regular log line but a scheduler event which is logged locally (can be changed later).
Cause I'm not even sure tbh what a term "target" means when it's value is scheduler? We're not sending a log line to a scheduler here.

eabatalov: Or `source=SCHEDULER`. This is clear.

server/src/main.rs
```diff
@@ -77,13 +82,52 @@ where
     Box::new(tracing_subscriber::fmt::layer().compact())
 }
 
+fn get_local_log_layer<S>(config: &ServerConfig) -> Box<dyn Layer<S> + Send + Sync>
```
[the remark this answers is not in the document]
earhart: Done :-)

earhart:
(Merging for now so I can get on to the next PR, but I'm definitely open to revisiting these definitions later.)

PR #1593, opened 2025-07-13
Switch to using a single crate for Indexify

earhart:
## Context

We've been building Indexify as multiple crates, but that was only because we thought it built faster that way; we did some measurements and discovered that it doesn't.

## What

This change converts our multi-crate build into a single-crate build, rearranging the source code as needed.  It also removes / refactors all of the newly discovered dead code in the system.

## Testing

`cargo build`
`cargo check --all-targets --workspace`
`cargo test`
Ran Indexify integration tests locally.

Verified that all builds are clean (no errors or warnings), and that tests pass.

## Contribution Checklist


diptanu (approved):
Marking this as approved as I am assuming you have just moved things around.

Issue #1592, opened 2025-07-12
Indexify APIs


diptanu:
Indexify APIs are not versioned, and before they are used in Tensorlake Cloud we need to version them, and it's an opportunity to update them. Proposal 

###### Create Graph
```
POST /v1/namespaces/{namespace}/computegraphs

Body - 
{
 // Compute Graph Payload
}
```

###### List Graphs
```
LIST  /v1/namespaces/{namespace}/computegraphs?from=<curosr>

Response - 
[
{
 // Compute Graph Payload
}
cursor: "..."
]
```

###### Get Graph 
```
GET  /v1/namespaces/{namespace}/computegraphs/{graph}

Response - 
{
 // Compute Graph Payload
}
```

###### Delete Graph 
```
GET  /v1/namespaces/{namespace}/computegraphs/{graph}
```

###### Create Request 
```
POST  /v1/namespaces/{namespace}/computegraphs/{graph}

Body -

{
  // function parameter names and values 
  // values can be JSON or bytes with mimetype. This allows us to upload files to functions, or call them with regular python objects or just values
}

Return 
{
"request_id": "..."
}
```

##### Get Response of requests
```
GET  /v1/namespaces/{namespace}/computegraphs/{graph}/request/{request_id}

Response -

{
"status": "...",
    "created_at": "....",
    "error": "....",
     "outcome": "....",
    "outputs": [
       {
         "function_name": "o
       }
    ]
}
```

###### List Requests
```
POST  /v1/namespaces/{namespace}/computegraphs/{graph}/requests

Body -
[
{
   "request_id": "....",
    "status": "...",
    "created_at": "....",
    "error": "....",
     "outcome": "....",
}
]
```

###### List Tasks
```
POST  /v1/namespaces/{namespace}/computegraphs/{graph}/request/{request}/tasks

Body -
[
{
   "task_id": "....",
    "status": "...",
    "created_at": "....",
     "outcome": "....",
     "allocations" : [
       {
         "allocation_id": "....",
         "status": ".....",
          "outcome": "...."
       }
     ]
}
]
```

###### Cancel Requests
```
DELETE  /v1/namespaces/{namespace}/computegraphs/{graph}/request/{request}
```

###### Get Output of a function
```
GET  /v1/namespaces/{namespace}/computegraphs/{graph}/request/{request}/function/{function}

returns the output binary 
```

###### Get Output from one of the instance of function

[...]

PR #1606, opened 2025-07-15
Add /v1 namespaces & advertise as defaults

earhart:
## Context

Currently, the public Indexify URLs are all under `/namespaces`; we'd like them to be under `/v1/namespaces`.

This is the first step of implementing #1592.

## What

This change adds `/v1` as an alternative prefix for requests, and modifies our published Swagger API to to include that prefix.  Once a server with this prefix is released, we can update the paths in the client library, and then remove server support for `/namespaces`.

## Testing

Ran a local executor and local Indexify tests
Manually verified that `/namespaces` and `/v1/namespaces` both work
Checked the Swagger UI to make sure the `v1` endpoints are exported.

## Contribution Checklist

### DIFF
--- a/server/src/routes.rs
+++ b/server/src/routes.rs
@@ -173,11 +173,23 @@ pub fn create_routes(route_state: RouteState) -> Router {
             get(namespaces).with_state(route_state.clone()),
         )
         .route(
+            "/v1/namespaces",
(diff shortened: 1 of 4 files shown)


diptanu (changes requested):
Just marking this PR as not ready for review yet. We need the APIs here match what we have in #1592

PR #1609, opened 2025-07-16
fix: postcss and brace-expansion versioning

ilbum:
- Update of "postcss" package, this is okay as we're not using its features (postcss). See vulnerability here https://github.com/tensorlakeai/indexify/security/dependabot/136
- Update of "brace-expansion" package. See vulnerability here https://github.com/tensorlakeai/indexify/security/dependabot/212

### DIFF
--- a/server/ui/package.json
+++ b/server/ui/package.json
@@ -40,12 +40,12 @@
     "web-vitals": "^2.1.4"
   },
   "overrides": {
-    "postcss": "<8.4.31",
-    "brace-expansion": "<=1.1.11",
+    "postcss": "^8.4.31",
+    "brace-expansion": "^1.1.11",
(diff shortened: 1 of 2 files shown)


server/ui/package.json
```diff
@@ -40,8 +40,8 @@
     "web-vitals": "^2.1.4"
   },
   "overrides": {
-    "postcss": "<8.4.31",
-    "brace-expansion": "<=1.1.11",
+    "postcss": "^8.4.31",
+    "brace-expansion": "^1.1.11",
```
eabatalov, line 44: Nit (not important): we need to update to 1.1.12 to mitigate the vulnerability but the changes in the lock file show that npm pull the latest version 1.1.12 which has the fix.

eabatalov (approved):

PR #1611, opened 2025-07-16
Log Server->FE alloc assignment deltas

earhart:
## Context

We'd like some logging for the FE assignment deltas to aid in system debugging.

## What

This change saves the desired executor states as they're being sent, and computes a delta from the current state and the previous state, logging any changes.

## Testing

Ran local integration tests, and watched the log messages.

## Contribution Checklist

### DIFF
--- a/server/src/executor_api.rs
+++ b/server/src/executor_api.rs
@@ -19,12 +19,13 @@ use executor_api_pb::{
     HostResources,
     ReportExecutorStateRequest,
     ReportExecutorStateResponse,
+    TaskAllocation,
     TaskResult,
 };
 use tokio::sync::watch;
 use tokio_stream::{wrappers::WatchStream, Stream};
 use tonic::{Request, Response, Status};
-use tracing::{debug, error, info, trace, warn};
+use tracing::{debug, error, info, info_span, trace, warn};
 
 use crate::{
     blob_store::{self, BlobStorage},
@@ -829,11 +830,14 @@ impl ExecutorApi for ExecutorAPIService {
         }));
         let executor_manager = self.executor_manager.clone();
         tokio::spawn(async move {
+            let loop_span = info_span!("Executor update loop", executor_id = executor_id.get());
+
             // 1. Mark the state as changed to trigger the first change notification to the
             //    executor. This is important because between the report_executor_state and
             //    the get_desired_executor_states requests, the executor state may received
             //    a new desired state.
             state_rx.mark_changed();
+            let mut last_sent_state = DesiredExecutorState::default();
             loop {
                 // 2. Wait for a state change for this executor
                 if let Err(err) = state_rx.changed().await {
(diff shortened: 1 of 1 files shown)


server/src/executor_api.rs
```diff
@@ -865,14 +869,61 @@ impl ExecutorApi for ExecutorAPIService {
...
+                        if let (Some(fn_executor_id), Some(allocation_id)) =
+                            (fn_executor_id_option, allocation_id_option)
+                        {
+                            match last_assignments.get(fn_executor_id) {
+                                Some(last_allocation_id) => {
+                                    if allocation_id != last_allocation_id {
+                                        info!(
+                                            fn_executor_id,
+                                            allocation_id, last_allocation_id, "re-assigning FE"
+                                        )
```
diptanu, line 900: These are nice logs, I think for what we needed just seeing the list of FEs that we are sending in the desired state would have been enough. At max there is going to be 8-10 FEs on most machines, if we looked at a series of logs, we would have figured out this logic. Just wondering if it's worth doing the difference here for what we need to know right now.

server/src/executor_api.rs
```diff
@@ -865,14 +869,61 @@ impl ExecutorApi for ExecutorAPIService {
                     "get_desired_executor_states: got desired state: {:#?}", desired_state
                 );
 
+                // Log state differences
```
eabatalov, line 872: Why don't we use local log for this?
This looks like exact use case.

We can see what Server assigned in its allocations API. So the only gap we have is to know what server sent.
If we use local log here we can just dump the whole desired state or its subset without even diffing anything.

Also I think this function is getting really huge and this logging really needs to be extracted out of it.

server/src/executor_api.rs
```diff
@@ -856,33 +907,25 @@ impl ExecutorApi for ExecutorAPIService {
                 }
 
                 // 3. Get the latest state
-                let desired_state: DesiredExecutorState = match executor_manager
-                    .get_executor_state(&executor_id)
-                    .await
-                    .try_into()
```
eabatalov, line 862: This try_into is not needed?

server/src/executor_api.rs
```diff
@@ -856,33 +907,25 @@ impl ExecutorApi for ExecutorAPIService {
...
+                let desired_state = executor_manager.get_executor_state(&executor_id).await;
+
                 {
```
eabatalov, line 912: Do we need this `{`? Maybe I misunderstand something.

eabatalov (dismissed):
Just minor comments, lgtm

server/src/executor_api.rs
```diff
@@ -711,6 +712,53 @@ impl ExecutorAPIService {
     }
 }
 
+fn log_desired_executor_state(
```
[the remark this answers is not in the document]
earhart: Done

server/src/executor_api.rs
earhart: Nope - turns out we're already getting a `DesiredExecutorState` here; it just converts into itself.

server/src/executor_api.rs
```diff
@@ -711,6 +712,105 @@ impl ExecutorAPIService {
...
+        }
+    }
+
+    for (fn_executor_id, last_allocation_id) in last_assignments {
+        info!(%fn_executor_id, %last_allocation_id, "idling FE")
+    }
+}
+
+#[instrument(skip_all, fields(executor_id = %executor_id))]
+async fn executor_update_loop(
```
eabatalov, line 760: Nice function extraction refactoring

eabatalov (approved):
ty

PR #1612, opened 2025-07-17
Changing the grpc stream to use mpsc instead of watch.

diptanu:
Changing the grpc stream to use mpsc instead of watch. We want the receiver to drop when the sender is dropped. From the docs of tokio watch its not quite clear when the receiver is dropped. Plus watch semantics doesn't guarantee that the producer sees all the values the sender is sending.

### DIFF
--- a/server/src/executor_api.rs
+++ b/server/src/executor_api.rs
@@ -21,8 +21,8 @@ use executor_api_pb::{
     ReportExecutorStateResponse,
     TaskResult,
 };
-use tokio::sync::watch;
-use tokio_stream::{wrappers::WatchStream, Stream};
+use tokio::sync::mpsc;
+use tokio_stream::{wrappers::ReceiverStream, Stream};
 use tonic::{Request, Response, Status};
 use tracing::{debug, error, info, trace, warn};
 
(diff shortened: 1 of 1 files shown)


eabatalov (changes requested):
I looked at this code and I think that existing behavior is correct.
If Executor disconnects and then connects back and there are multiple desired states in the server queue then we should only send the latest desired state to the Executor.

diptanu:
> I looked at this code and I think that existing behavior is correct. If Executor disconnects and then connects back and there are multiple desired states in the server queue then we should only send the latest desired state to the Executor.

@eabatalov This will not happen. We don't queue desired state events when an executor's desired state stream is interrupted. When the executor registers for the first time, we start sending state changes to the executor. Every state change reflects the current desired state. It doesn't change the behavior from what I have seen. I couldn't find any documentation around the semantics of what happens to the receiver when the transmitter is dropped. Every tonic example for bidirectional stream uses mpsc channel.

eabatalov:
I'm not sure about the following: if Executor slows down and consume desired states slower than Server adds them then what's going to happen?

PR #1622, opened 2025-07-18
Update tensorlake and pin httpx version

eabatalov:
Tensorlake relaxes versioning requirements for httpx and
pydantic packages. We need to pin httpx and pydantic in indexify
to not get them updated without us knowing about that.

Also remove code that uses pydantic in Executor where we don't
serilize anything.

This is complementary to https://github.com/tensorlakeai/tensorlake/pull/262


ilbum (approved):
lgtm

PR #1625, opened 2025-07-21
Execution duration

diptanu:
Added information about execution duration of functions


server/proto/executor_api.proto
```diff
@@ -272,6 +272,9 @@ message TaskResult {
     optional DataPayload stderr = 14;
     // User payload for invocation error if task failed with invocation error.
     optional DataPayload invocation_error_output = 15;
+
+    // The time taken to run the function in seconds.
+    optional uint64 execution_duration_s = 16;
```
eabatalov, line 277: Let's use ms here, it is going to be a long term thing. Serverless products use ms billing.
Also please use _ms or _sec suffix here to be consistent with the rest.

PR #1634, opened 2025-07-22
Add top-level .justfile

earhart:
## Context

It's useful to have a set of repo commands.  :-)

## What

This PR adds a .justfile with a few useful commands.  NB This doesn't conflict with other possible ways to define repo commands, like adding a top-level Makefile (which we might also want to do).

## Testing

Ran a few of the commands; they seem to do the right thing.

## Contribution Checklist

### DIFF
--- a/.justfile
+++ b/.justfile
@@ -1,0 +1,67 @@
+# https://just.systems
+
+[private]
+default:
+    @just --list
+
+[private]
+nightly-toolchain:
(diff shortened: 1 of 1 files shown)


eabatalov (approved):
We're copy-pasting commands from the Makefiles there which results in us needing to maintain more copy-pasted code. I don't really get why but I don't mind as it's a small change

PR #1642, opened 2025-07-25
Use async RPCs for FE task management

earhart:
## Context

We want to use asynchronous RPCs for FE task management, so that we can stream results and get progress updates from the FE.

## What

This change implements the executor side of using asynchronous streaming RPCs for FE task management.  NB: It should not go in until https://github.com/tensorlakeai/tensorlake/pull/284 lands.

## Testing

Tensorlake tests, Indexify Python tests

## Contribution Checklist

- [X] If a Python package was changed, please run `make fmt` in the package directory.

### DIFF
--- a/indexify/src/indexify/executor/function_executor_controller/run_task.py
+++ b/indexify/src/indexify/executor/function_executor_controller/run_task.py
@@ -88,7 +88,6 @@ async def run_task_on_function_executor(
 
     metric_function_executor_run_task_rpcs.inc()
     metric_function_executor_run_task_rpcs_in_progress.inc()
-    start_time = time.monotonic()
     # Not None if the Function Executor should be terminated after running the task.
     function_executor_termination_reason: Optional[
         FunctionExecutorTerminationReason
@@ -101,7 +100,7 @@ async def run_task_on_function_executor(
     # let the AioRpcError to be raised here.
     timeout_sec = task_info.allocation.task.timeout_ms / 1000.0
     try:
-        task_result = await _run_task(task, function_executor, timeout_sec)
+        task_result = await _run_task_rpcs(task, function_executor, timeout_sec)
 
         task_info.output = _task_output_from_function_executor_result(
             allocation=task_info.allocation,
@@ -111,7 +110,8 @@ async def run_task_on_function_executor(
             logger=logger,
         )
     except asyncio.TimeoutError:
-        # The task is still running in FE, we only cancelled the client-side RPC.
+        # This is an await_task() RPC timeout - we're not getting
+        # progress messages or a task completion.
         function_executor_termination_reason = (
             FunctionExecutorTerminationReason.FUNCTION_EXECUTOR_TERMINATION_REASON_FUNCTION_TIMEOUT
         )
@@ -122,13 +122,26 @@ async def run_task_on_function_executor(
             execution_end_time=time.monotonic(),
         )
     except grpc.aio.AioRpcError as e:
-        metric_function_executor_run_task_rpc_errors.inc()
-        logger.error("task execution failed", exc_info=e)
-        task_info.output = TaskOutput.internal_error(
-            allocation=task_info.allocation,
-            execution_start_time=execution_start_time,
-            execution_end_time=time.monotonic(),
-        )
+        if e.code() == grpc.StatusCode.DEADLINE_EXCEEDED:
+            # This is either a create_task() RPC timeout or a
+            # delete_task() RPC timeout; either suggests that the FE
+            # is unhealthy.
+            function_executor_termination_reason = (
+                FunctionExecutorTerminationReason.FUNCTION_EXECUTOR_TERMINATION_REASON_UNHEALTHY
+            )
+            task_info.output = TaskOutput.function_executor_unresponsive(
+                allocation=task_info.allocation,
+                execution_start_time=execution_start_time,
+                execution_end_time=time.monotonic(),
+            )
+        else:
+            metric_function_executor_run_task_rpc_errors.inc()
+            logger.error("task execution failed", exc_info=e)
+            task_info.output = TaskOutput.internal_error(
+                allocation=task_info.allocation,
+                execution_start_time=execution_start_time,
+                execution_end_time=time.monotonic(),
+            )
     except asyncio.CancelledError:
         # The task is still running in FE, we only cancelled the client-side RPC.
         function_executor_termination_reason = (
@@ -148,7 +161,9 @@ async def run_task_on_function_executor(
             execution_end_time=time.monotonic(),
         )
 
-    metric_function_executor_run_task_rpc_latency.observe(time.monotonic() - start_time)
+    metric_function_executor_run_task_rpc_latency.observe(
+        time.monotonic() - execution_start_time
+    )
     metric_function_executor_run_task_rpcs_in_progress.dec()
 
     function_executor.invocation_state_client().remove_task_to_invocation_id_entry(
@@ -178,48 +193,62 @@ async def run_task_on_function_executor(
     )
 
 
-async def _run_task(
+async def _run_task_rpcs(
     task: Task, function_executor: FunctionExecutor, timeout_sec: float
-):
-    """Runs the task, returning the result, reporting errors via exceptions."""
+) -> TaskResult:
+    """Runs the task, returning the result, reporting errors via exceptions.
 
-    last_response: Optional[AwaitTaskProgress] = None
+    Exceptions:
+
+        asyncio.TimeoutError: Timeout while waiting for progress
+            messages from the function executor.
+
+        grpc.aio.AioRpcError(grpc.StatusCode.DEADLINE_EXCEEDED):
+            Timeout in the create_task() or delete_task() RPCs,
+            indicating either an error in the implementation or
+            badly-behaved user code.
+
+        Anything else: Other system failures (incorrect function
+            executor response, unable to connect to the function
+            executor, &c).
+    """
+
+    response: AwaitTaskProgress
     channel: grpc.aio.Channel = function_executor.channel()
     fe_stub = FunctionExecutorStub(channel)
 
-    # Use a single timeout across all three GRPC calls, restarting when await_task() returns responses
-    current_rpc = None
+    # Create task with timeout
+    await fe_stub.create_task(
+        CreateTaskRequest(task=task), timeout=_CREATE_TASK_TIMEOUT_SECS
+    )
 
-    try:
-        # Create task with timeout
-        current_rpc = fe_stub.create_task(CreateTaskRequest(task=task))
-        await asyncio.wait_for(current_rpc, timeout=_CREATE_TASK_TIMEOUT_SECS)
-        current_rpc = None
+    # Await task with timeout resets on each response
+    await_rpc = fe_stub.await_task(AwaitTaskRequest(task_id=task.task_id))
 
-        # Await task with timeout resets on each response
-        current_rpc = fe_stub.await_task(AwaitTaskRequest(task_id=task.task_id))
+    while True:
+        # Wait for next response with fresh timeout each time
+        response = await asyncio.wait_for(await_rpc.read(), timeout=timeout_sec)
+        if response.WhichOneof("response") == "task_result":
+            # We're done waiting; cancel the outstanding RPC to ensure
+            # any resources in use are cleaned up.
+            await_rpc.cancel()
+            break
 
-        while True:
-            # Wait for next response with fresh timeout each time
-            response = await asyncio.wait_for(current_rpc.read(), timeout=timeout_sec)
-            if response == grpc.aio.EOF:
-                current_rpc = None
-                break
-            last_response = response
+        # NB: We don't actually check for other message types
+        # here; any message from the FE is treated as an
+        # indication that it's making forward progress.
 
-        # Delete task with timeout
-        current_rpc = fe_stub.delete_task(DeleteTaskRequest(task_id=task.task_id))
-        await asyncio.wait_for(current_rpc, timeout=_DELETE_TASK_TIMEOUT_SECS)
+        if response == grpc.aio.EOF:
+            raise Exception(
+                "Function Executor didn't return function/task alloc response"
+            )
 
-    finally:
-        # Cancel any outstanding RPCs to clean up our client-side resources
-        if current_rpc is not None:
-            current_rpc.cancel()
+    # Delete task with timeout
+    await fe_stub.delete_task(
+        DeleteTaskRequest(task_id=task.task_id), timeout=_DELETE_TASK_TIMEOUT_SECS
+    )
 
-    if not last_response or last_response.WhichOneof("response") != "task_result":
-        raise Exception("Expected a final task result")
-
-    return last_response.task_result
+    return response.task_result
 
 
 def _task_output_from_function_executor_result(
--- a/indexify/src/indexify/executor/function_executor_controller/run_task.py
+++ b/indexify/src/indexify/executor/function_executor_controller/run_task.py
@@ -93,6 +93,12 @@ async def run_task_on_function_executor(
         FunctionExecutorTerminationReason
     ] = None
 
+    # NB: We start this timer before invoking the first RPC, since
+    # user code should be executing by the time the create_task() RPC
+    # returns, so not attributing the task management RPC overhead to
+    # the user would open a possibility for abuse. (This is somewhat
+    # mitigated by the fact that these RPCs should have a very low
+    # overhead.)
     execution_start_time: Optional[float] = time.monotonic()
 
     # If this RPC failed due to customer code crashing the server we won't be
@@ -122,26 +128,35 @@ async def run_task_on_function_executor(
             execution_end_time=time.monotonic(),
         )
     except grpc.aio.AioRpcError as e:
+        # This indicates some sort of problem communicating with the FE.
+        #
+        # NB: We charge the user in these situations: code within the
+        # FE is not isolated, so not charging would enable abuse.
+        #
+        # This is an unexpected situation, though, so we make sure to
+        # log the situation for further investigation.
+
+        function_executor_termination_reason = (
+            FunctionExecutorTerminationReason.FUNCTION_EXECUTOR_TERMINATION_REASON_UNHEALTHY
+        )
+        metric_function_executor_run_task_rpc_errors.inc()
+
         if e.code() == grpc.StatusCode.DEADLINE_EXCEEDED:
             # This is either a create_task() RPC timeout or a
             # delete_task() RPC timeout; either suggests that the FE
             # is unhealthy.
-            function_executor_termination_reason = (
-                FunctionExecutorTerminationReason.FUNCTION_EXECUTOR_TERMINATION_REASON_UNHEALTHY
-            )
-            task_info.output = TaskOutput.function_executor_unresponsive(
-                allocation=task_info.allocation,
-                execution_start_time=execution_start_time,
-                execution_end_time=time.monotonic(),
-            )
+            logger.error("task management RPC execution deadline exceeded", exc_info=e)
         else:
-            metric_function_executor_run_task_rpc_errors.inc()
-            logger.error("task execution failed", exc_info=e)
-            task_info.output = TaskOutput.internal_error(
-                allocation=task_info.allocation,
-                execution_start_time=execution_start_time,
-                execution_end_time=time.monotonic(),
-            )
+            # This is a status from an unsuccessful RPC; this
+            # shouldn't happen, but we handle it.
+            logger.error("task management RPC failed", exc_info=e)
+
+        task_info.output = TaskOutput.function_executor_unresponsive(
+            allocation=task_info.allocation,
+            execution_start_time=execution_start_time,
+            execution_end_time=time.monotonic(),
+        )
+
     except asyncio.CancelledError:
         # The task is still running in FE, we only cancelled the client-side RPC.
         function_executor_termination_reason = (
@@ -153,8 +168,11 @@ async def run_task_on_function_executor(
             execution_end_time=time.monotonic(),
         )
     except Exception as e:
-        metric_function_executor_run_task_rpc_errors.inc()
-        logger.error("task execution failed", exc_info=e)
+        # This is an unexpected exception; we believe that this
+        # indicates an internal error.
+        logger.error(
+            "Unexpected internal error during task lifecycle RPC sequence", exc_info=e
+        )
         task_info.output = TaskOutput.internal_error(
             allocation=task_info.allocation,
             execution_start_time=execution_start_time,
@@ -196,22 +214,7 @@ async def run_task_on_function_executor(
 async def _run_task_rpcs(
     task: Task, function_executor: FunctionExecutor, timeout_sec: float
 ) -> TaskResult:
-    """Runs the task, returning the result, reporting errors via exceptions.
-
-    Exceptions:
-
-        asyncio.TimeoutError: Timeout while waiting for progress
-            messages from the function executor.
-
-        grpc.aio.AioRpcError(grpc.StatusCode.DEADLINE_EXCEEDED):
-            Timeout in the create_task() or delete_task() RPCs,
-            indicating either an error in the implementation or
-            badly-behaved user code.
-
-        Anything else: Other system failures (incorrect function
-            executor response, unable to connect to the function
-            executor, &c).
-    """
+    """Runs the task, returning the result, reporting errors via exceptions."""
 
     response: AwaitTaskProgress
     channel: grpc.aio.Channel = function_executor.channel()
@@ -238,8 +241,13 @@ async def _run_task_rpcs(
             # indication that it's making forward progress.
 
             if response == grpc.aio.EOF:
-                raise Exception(
-                    "Function Executor didn't return function/task alloc response"
+                # Protocol error: we should get a task_result before
+                # we see the RPC complete.
+                raise grpc.aio.AioRpcError(
+                    grpc.StatusCode.CANCELLED,
+                    None,
+                    None,
+                    "Function Executor didn't return function/task alloc response",
                 )
     finally:
         # Cancel the outstanding RPC to ensure any resources in use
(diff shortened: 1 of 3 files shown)


indexify/src/indexify/executor/function_executor_controller/run_task.py
```diff
@@ -89,13 +94,32 @@ async def run_task_on_function_executor(
     # let the AioRpcError to be raised here.
     timeout_sec = task_info.allocation.task.timeout_ms / 1000.0
     try:
+        last_response = None
         channel: grpc.aio.Channel = function_executor.channel()
-        response: RunTaskResponse = await FunctionExecutorStub(channel).run_task(
-            request, timeout=timeout_sec
-        )
+        fe = FunctionExecutorStub(channel)
```
eabatalov, line 99: There's already a term Function Executor widely used in the code base, let's use something more precise here like fe_client or fe_stub. There's a `function_executor` variable in this function already :)

indexify/src/indexify/executor/function_executor_controller/run_task.py
```diff
@@ -89,13 +94,32 @@ async def run_task_on_function_executor(
...
+        await fe.start_task(request, timeout=timeout_sec)
+        try:
+            async for response in fe.await_task(
+                AwaitTaskRequest(task_id=request.task_id), timeout=timeout_sec
+            ):
+                last_response = response
+        finally:
+            await fe.delete_task(
+                DeleteTaskRequest(task_id=request.task_id), timeout=timeout_sec
```
eabatalov, line 108: We shouldn't use function timeout for non function RPCs here and in fe.start_task

indexify/src/indexify/executor/function_executor_controller/run_task.py
```diff
@@ -89,13 +94,32 @@ async def run_task_on_function_executor(
...
+            )
+
+        if last_response and last_response.WhichOneof("response") == "task_result":
+            result = last_response.task_result
+        else:
+            result = RunTaskResponse(
```
eabatalov, line 114: When do we hit this else branch? Please explain why this is a function error in a comment in the code.

eabatalov: I assume we'll get some task_result if function succeeded or if it failed cleanely without tainting FE. This is why I'm asking what is expected to happen on FE side when this else branch is hit.

indexify/src/indexify/executor/function_executor_controller/run_task.py
```diff
@@ -89,13 +94,32 @@ async def run_task_on_function_executor(
...
+        last_response = None
         channel: grpc.aio.Channel = function_executor.channel()
-        response: RunTaskResponse = await FunctionExecutorStub(channel).run_task(
-            request, timeout=timeout_sec
-        )
+        fe = FunctionExecutorStub(channel)
+        await fe.start_task(request, timeout=timeout_sec)
+        try:
+            async for response in fe.await_task(
+                AwaitTaskRequest(task_id=request.task_id), timeout=timeout_sec
```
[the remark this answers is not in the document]
eabatalov: I guess I'm wrong after reading docs but please confirm that.

indexify/src/indexify/executor/function_executor_controller/run_task.py
```diff
@@ -89,13 +94,32 @@ async def run_task_on_function_executor(
...
+            ):
+                last_response = response
+        finally:
+            await fe.delete_task(
```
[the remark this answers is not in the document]
earhart: Done

indexify/src/indexify/executor/function_executor_controller/run_task.py
earhart: For now, I think this is fine; we do need some sort of timeout. Longer term, I'm definitely open to better ideas.

indexify/src/indexify/executor/function_executor_controller/run_task.py
```diff
@@ -89,13 +94,32 @@ async def run_task_on_function_executor(
...
-        response: RunTaskResponse = await FunctionExecutorStub(channel).run_task(
-            request, timeout=timeout_sec
-        )
+        fe = FunctionExecutorStub(channel)
+        await fe.start_task(request, timeout=timeout_sec)
+        try:
+            async for response in fe.await_task(
+                AwaitTaskRequest(task_id=request.task_id), timeout=timeout_sec
+            ):
+                last_response = response
```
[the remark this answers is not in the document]
earhart: If the FE crashes, the stream should close, or we'll time out; if the FE doesn't disconnect, it might legitimately still be working. I've reworked this logic a bit; hopefully this is more clear?

indexify/src/indexify/executor/function_executor_controller/run_task.py
```diff
@@ -89,13 +94,32 @@ async def run_task_on_function_executor(
     # let the AioRpcError to be raised here.
     timeout_sec = task_info.allocation.task.timeout_ms / 1000.0
     try:
+        last_response = None
```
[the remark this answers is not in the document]
earhart: I looked into that, but once you pull this logic out of the function, there's not much left; it doesn't seem worth having things separate...

indexify/src/indexify/executor/function_executor_controller/run_task.py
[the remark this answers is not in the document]
eabatalov: Thank you for explaining, this should work.

indexify/src/indexify/executor/function_executor_controller/run_task.py
```diff
@@ -90,18 +97,72 @@ async def run_task_on_function_executor(
...
+
+        # Use a single timeout across all three GRPC calls, restarting when await_task() returns responses
+        current_rpc = None
+
+        try:
+            # Create task with timeout
+            logger.info("CreateTask(%s)", task.task_id)
+            current_rpc = fe_stub.create_task(CreateTaskRequest(task=task))
+            await asyncio.wait_for(current_rpc, timeout=timeout_sec)
+            logger.info("CreateTask(%s) => Complete", task.task_id)
```
eabatalov, line 113: task_id is already in the logger context.

Also this looks like debug logging, can we remove it?

indexify/src/indexify/executor/function_executor_controller/run_task.py
```diff
@@ -90,18 +97,72 @@ async def run_task_on_function_executor(
...
+            # Cancel any outstanding RPCs to clean up our client-side resources
+            if current_rpc is not None:
+                current_rpc.cancel()
+
+        if last_response and last_response.WhichOneof("response") == "task_result":
+            result = last_response.task_result
+        else:
+            result = TaskResult(
+                outcome_code=TaskOutcomeCode.TASK_OUTCOME_CODE_FAILURE,
+                failure_reason=TaskFailureReason.TASK_FAILURE_REASON_FUNCTION_ERROR,
```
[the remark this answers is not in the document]
eabatalov: Please just raise an Exception() here with a proper error message, it'll be caught down below in `except Exception as e:` block with all the correct handling.

indexify/src/indexify/executor/function_executor_controller/run_task.py
```diff
@@ -90,18 +97,72 @@ async def run_task_on_function_executor(
...
+                response = await asyncio.wait_for(
+                    current_rpc.read(), timeout=timeout_sec
+                )
+                if response == grpc.aio.EOF:
+                    current_rpc = None
+                    break
+                last_response = response
+
+            # Delete task with timeout
+            current_rpc = fe_stub.delete_task(DeleteTaskRequest(task_id=task.task_id))
```
eabatalov, line 130: Delete task failure should result in FE termination.

indexify/src/indexify/executor/function_executor_controller/run_task.py
```diff
@@ -90,18 +97,72 @@ async def run_task_on_function_executor(
...
+            await asyncio.wait_for(current_rpc, timeout=timeout_sec)
+
+        finally:
+            # Cancel any outstanding RPCs to clean up our client-side resources
+            if current_rpc is not None:
+                current_rpc.cancel()
+
+        if last_response and last_response.WhichOneof("response") == "task_result":
```
eabatalov, line 138: Can process  the last response at the point where we get it so we don't need to create a local var last_response with so big scope.

indexify/src/indexify/executor/function_executor_controller/run_task.py
```diff
@@ -90,18 +97,72 @@ async def run_task_on_function_executor(
...
+        fe_stub = FunctionExecutorStub(channel)
+
+        # Use a single timeout across all three GRPC calls, restarting when await_task() returns responses
+        current_rpc = None
+
+        try:
+            # Create task with timeout
+            logger.info("CreateTask(%s)", task.task_id)
+            current_rpc = fe_stub.create_task(CreateTaskRequest(task=task))
+            await asyncio.wait_for(current_rpc, timeout=timeout_sec)
```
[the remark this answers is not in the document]
eabatalov: Also execution_start_time shouldn't include create_task and delete_task calls because we're not running function code there. This is not an issue right now but if we change anything in these RPCs to make them run longer then we won't know that we're overcharging customers.

indexify/src/indexify/executor/function_executor_controller/run_task.py
```diff
@@ -90,18 +97,72 @@ async def run_task_on_function_executor(
     # let the AioRpcError to be raised here.
     timeout_sec = task_info.allocation.task.timeout_ms / 1000.0
     try:
+        last_response: Optional[AwaitTaskProgress] = None
```
eabatalov, line 100: Is possible let's extract this code with the 3 RPCs to a separate function while relying on this outer try catch block to catch any exceptions going out of it. This should help to add more structure to this large function cause the new extracted function will have to return some well structured value.

indexify/src/indexify/executor/function_executor_controller/run_task.py
```diff
@@ -90,18 +97,72 @@ async def run_task_on_function_executor(
...
+            FunctionExecutorTerminationReason.FUNCTION_EXECUTOR_TERMINATION_REASON_FUNCTION_TIMEOUT
+        )
+        task_info.output = TaskOutput.function_timeout(
+            allocation=task_info.allocation,
+            timeout_sec=timeout_sec,
+            execution_start_time=execution_start_time,
+            execution_end_time=time.monotonic(),
+        )
     except grpc.aio.AioRpcError as e:
         if e.code() == grpc.StatusCode.DEADLINE_EXCEEDED:
```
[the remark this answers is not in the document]
earhart: Done

indexify/src/indexify/executor/function_executor_controller/run_task.py
earhart: Done (thanks -- it was definitely debug code :-)

indexify/src/indexify/executor/function_executor_controller/run_task.py
earhart: TBH I think any of these failing should result in FE termination.

indexify/src/indexify/executor/function_executor_controller/run_task.py
earhart: Not really, but I broke it out into a separate function so this should be okay.

indexify/src/indexify/executor/function_executor_controller/run_task.py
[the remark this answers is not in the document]
earhart: Oh, and as for using the GRPC timeouts -- that results in a different exception, so it'd complicate the timeout handling logic; it's simpler to just have one mechanism for timeouts.

indexify/src/indexify/executor/function_executor_controller/run_task.py
```diff
@@ -171,38 +178,82 @@ async def run_task_on_function_executor(
     )
 
 
-def _task_output_from_function_executor_response(
+async def _run_task(
+    task: Task, function_executor: FunctionExecutor, timeout_sec: float
```
eabatalov, line 182: Nit. Only FE  channel, not whole FE needs to be passed here.

indexify/src/indexify/executor/function_executor_controller/run_task.py
```diff
@@ -171,38 +178,82 @@ async def run_task_on_function_executor(
...
+):
```
eabatalov, line 183: Please add type annotations everywhere including here.

indexify/src/indexify/executor/function_executor_controller/run_task.py
```diff
@@ -83,45 +93,42 @@ async def run_task_on_function_executor(
...
     # let the AioRpcError to be raised here.
     timeout_sec = task_info.allocation.task.timeout_ms / 1000.0
     try:
-        channel: grpc.aio.Channel = function_executor.channel()
-        execution_start_time = time.monotonic()
-        response: RunTaskResponse = await FunctionExecutorStub(channel).run_task(
-            request, timeout=timeout_sec
-        )
-        task_info.output = _task_output_from_function_executor_response(
+        task_result = await _run_task(task, function_executor, timeout_sec)
```
eabatalov, line 104: Type annotation. Please add them everywhere to help ppl reading code. You can see them they are put in most places in existing code base.

indexify/src/indexify/executor/function_executor_controller/run_task.py
```diff
@@ -171,38 +178,82 @@ async def run_task_on_function_executor(
...
+    fe_stub = FunctionExecutorStub(channel)
+
+    # Use a single timeout across all three GRPC calls, restarting when await_task() returns responses
+    current_rpc = None
+
+    try:
+        # Create task with timeout
+        current_rpc = fe_stub.create_task(CreateTaskRequest(task=task))
+        await asyncio.wait_for(current_rpc, timeout=_CREATE_TASK_TIMEOUT_SECS)
+        current_rpc = None
```
eabatalov, line 197: current_rpc = None is not needed because cancel() is idempotent, see https://grpc.github.io/grpc/python/grpc_asyncio.html#grpc.aio.RpcContext.cancel and https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.cancel

indexify/src/indexify/executor/function_executor_controller/run_task.py
```diff
@@ -171,38 +178,82 @@ async def run_task_on_function_executor(
...
+        # Await task with timeout resets on each response
+        current_rpc = fe_stub.await_task(AwaitTaskRequest(task_id=task.task_id))
+
+        while True:
+            # Wait for next response with fresh timeout each time
+            response = await asyncio.wait_for(current_rpc.read(), timeout=timeout_sec)
+            if response == grpc.aio.EOF:
+                current_rpc = None
+                break
+            last_response = response
```
eabatalov, line 208: Deadline extension happens for any message sent over this channel. Can we make it more explicit when it happens and make this code not misbehaving if FE starts sending a new message type on this channel? This will make changes of FE behaviors easier. We don't deploy FE and Executor at the same time except for very very big breaking changes.

Also `last_response` variable is not needed and it obfuscates the logic here imo. We can check if we got the result in the response and save result then call delete task and if it didn't raise then just return result.

indexify/src/indexify/executor/function_executor_controller/run_task.py
```diff
@@ -83,45 +93,42 @@ async def run_task_on_function_executor(
...
-            response=response,
+            result=task_result,
             execution_start_time=execution_start_time,
             execution_end_time=time.monotonic(),
             logger=logger,
         )
+    except asyncio.TimeoutError:
+        # The task is still running in FE, we only cancelled the client-side RPC.
+        function_executor_termination_reason = (
+            FunctionExecutorTerminationReason.FUNCTION_EXECUTOR_TERMINATION_REASON_FUNCTION_TIMEOUT
```
eabatalov, line 116: fyi with the current approach when we treat all grey failures as platform errors, this is not quite correct. "function timeout" is only applicable for the case when await_task RPC timed out.

indexify/src/indexify/executor/function_executor_controller/run_task.py
earhart: Done

indexify/src/indexify/executor/function_executor_controller/run_task.py
earhart: Done

indexify/src/indexify/executor/function_executor_controller/run_task.py
```diff
@@ -171,38 +178,82 @@ async def run_task_on_function_executor(
     )
 
 
-def _task_output_from_function_executor_response(
+async def _run_task(
```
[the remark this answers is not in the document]
earhart: Done

indexify/src/indexify/executor/function_executor_controller/run_task.py
earhart: Done

indexify/src/indexify/executor/function_executor_controller/run_task.py
earhart: Done

indexify/src/indexify/executor/function_executor_controller/run_task.py
```diff
@@ -171,38 +178,82 @@ async def run_task_on_function_executor(
...
+    task: Task, function_executor: FunctionExecutor, timeout_sec: float
+):
+    """Runs the task, returning the result, reporting errors via exceptions."""
```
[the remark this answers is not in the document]
earhart: Done

indexify/src/indexify/executor/function_executor_controller/run_task.py
```diff
@@ -171,38 +178,82 @@ async def run_task_on_function_executor(
...
+
+    last_response: Optional[AwaitTaskProgress] = None
+    channel: grpc.aio.Channel = function_executor.channel()
+    fe_stub = FunctionExecutorStub(channel)
+
+    # Use a single timeout across all three GRPC calls, restarting when await_task() returns responses
```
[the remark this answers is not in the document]
earhart: Done

indexify/src/indexify/executor/function_executor_controller/run_task.py
```diff
@@ -171,38 +178,82 @@ async def run_task_on_function_executor(
...
+    current_rpc = None
+
+    try:
+        # Create task with timeout
+        current_rpc = fe_stub.create_task(CreateTaskRequest(task=task))
+        await asyncio.wait_for(current_rpc, timeout=_CREATE_TASK_TIMEOUT_SECS)
```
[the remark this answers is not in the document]
earhart: Done

indexify/src/indexify/executor/function_executor_controller/run_task.py
earhart: Done

indexify/src/indexify/executor/function_executor_controller/run_task.py
earhart: Done

indexify/src/indexify/executor/function_executor_controller/run_task.py
```diff
@@ -171,38 +178,82 @@ async def run_task_on_function_executor(
...
+        current_rpc = fe_stub.delete_task(DeleteTaskRequest(task_id=task.task_id))
+        await asyncio.wait_for(current_rpc, timeout=_DELETE_TASK_TIMEOUT_SECS)
+
+    finally:
+        # Cancel any outstanding RPCs to clean up our client-side resources
+        if current_rpc is not None:
+            current_rpc.cancel()
+
+    if not last_response or last_response.WhichOneof("response") != "task_result":
+        raise Exception("Expected a final task result")
```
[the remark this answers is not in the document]
earhart: Done

indexify/src/indexify/executor/function_executor_controller/task_output.py
```diff
@@ -95,6 +95,24 @@ def function_timeout(
...
+        execution_start_time: Optional[float],
+        execution_end_time: Optional[float],
+    ) -> "TaskOutput":
+        """Creates a TaskOutput for an unresponsive FE."""
+        # Task stdout, stderr is not available.
+        return TaskOutput(
+            allocation=allocation,
+            outcome_code=TaskOutcomeCode.TASK_OUTCOME_CODE_FAILURE,
+            failure_reason=TaskFailureReason.TASK_FAILURE_REASON_FUNCTION_ERROR,
+            stderr=f"Function executor is not producing health check progress reports.",
```
eabatalov, line 111: This stderr is not correct. function_executor_unresponsive() is called when create_task or delete_task timed out. Also this is a customer visible message so it needs to be phrased in terms of "Function" not FE.

I think we need to use internal error here as it meets the Executor API definition, see https://github.com/tensorlakeai/indexify/blob/main/server/proto/executor_api.proto#L244C5-L244C39

indexify/src/indexify/executor/function_executor_controller/run_task.py
```diff
@@ -78,39 +88,49 @@ async def run_task_on_function_executor(
...
-            # The task is still running in FE, we only cancelled the client-side RPC.
+            # This is either a create_task() RPC timeout or a
+            # delete_task() RPC timeout; either suggests that the FE
+            # is unhealthy.
             function_executor_termination_reason = (
-                FunctionExecutorTerminationReason.FUNCTION_EXECUTOR_TERMINATION_REASON_FUNCTION_TIMEOUT
+                FunctionExecutorTerminationReason.FUNCTION_EXECUTOR_TERMINATION_REASON_UNHEALTHY
             )
-            task_info.output = TaskOutput.function_timeout(
+            task_info.output = TaskOutput.function_executor_unresponsive(
```
eabatalov, line 132: Looking more into this new failure mode the right logic here would be:

If create_task failed or timedout then fail the task alloc with TASK_FAILURE_REASON_FUNCTION_EXECUTOR_TERMINATED which means the function didn't run.
https://github.com/tensorlakeai/indexify/blob/main/indexify/src/indexify/executor/function_executor_controller/task_output.py#L119

If delete_task failed or timedout running then this is a function failure which you're using in function_executor_unresponsive correctly (TASK_FAILURE_REASON_FUNCTION_ERROR). The stderr there though needs to be adjusted and it'd be better to rename function_executor_unresponsive to function_error() and explain why we treat it as function error here if delete_task failed.

Do we handle failures of these RPCs (not timeouts)?

indexify/src/indexify/executor/function_executor_controller/run_task.py
```diff
@@ -78,39 +88,49 @@ async def run_task_on_function_executor(
...
     metric_function_executor_run_task_rpcs.inc()
     metric_function_executor_run_task_rpcs_in_progress.inc()
-    start_time = time.monotonic()
     # Not None if the Function Executor should be terminated after running the task.
     function_executor_termination_reason: Optional[
         FunctionExecutorTerminationReason
     ] = None
-    execution_start_time: Optional[float] = None
+
+    execution_start_time: Optional[float] = time.monotonic()
```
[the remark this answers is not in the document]
eabatalov: No need to fix that

indexify/src/indexify/executor/function_executor_controller/run_task.py
eabatalov: See the else branch in

```
except grpc.aio.AioRpcError as e:
        if e.code() == grpc.StatusCode.DEADLINE_EXCEEDED:
```

we do internal error there.

indexify/src/indexify/executor/function_executor_controller/run_task.py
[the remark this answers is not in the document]
earhart: Yes - we could maybe elide the DeleteTask, but charging for StartTask is unavoidable, and DeleteTask is very quick.

indexify/src/indexify/executor/function_executor_controller/run_task.py
```diff
@@ -171,38 +193,96 @@ async def run_task_on_function_executor(
...
+    await fe_stub.create_task(
+        CreateTaskRequest(task=task), timeout=_CREATE_TASK_TIMEOUT_SECS
+    )
+
+    # Await task with timeout resets on each response
+    await_rpc = fe_stub.await_task(AwaitTaskRequest(task_id=task.task_id))
+
+    while True:
+        # Wait for next response with fresh timeout each time
+        response = await asyncio.wait_for(await_rpc.read(), timeout=timeout_sec)
```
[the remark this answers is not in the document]
earhart: Sounds good - changed to cancel the RPC on timeout.

indexify/src/indexify/executor/function_executor_controller/run_task.py
```diff
@@ -122,26 +122,35 @@ async def run_task_on_function_executor(
...
-                allocation=task_info.allocation,
-                execution_start_time=execution_start_time,
-                execution_end_time=time.monotonic(),
-            )
+            logger.error("task management RPC execution deadline exceeded", exc_info=e)
         else:
-            metric_function_executor_run_task_rpc_errors.inc()
+            # This is a status from an unsuccessful RPC; this
+            # shouldn't happen, but we handle it.
             logger.error("task execution failed", exc_info=e)
```
eabatalov, line 146: "task management RPC failed" ?

indexify/src/indexify/executor/function_executor_controller/task_output.py
```diff
@@ -95,6 +95,24 @@ def function_timeout(
...
+        execution_start_time: Optional[float],
+        execution_end_time: Optional[float],
+    ) -> "TaskOutput":
+        """Creates a TaskOutput for an unresponsive FE."""
+        # Task stdout, stderr is not available.
+        return TaskOutput(
+            allocation=allocation,
+            outcome_code=TaskOutcomeCode.TASK_OUTCOME_CODE_FAILURE,
+            failure_reason=TaskFailureReason.TASK_FAILURE_REASON_FUNCTION_ERROR,
+            stderr=f"Function executor is not responding to task lifecycle RPCs.",
```
eabatalov, line 111: This message should be in our log, not in customer log.
Customer message needs to use terms from our docs.
Let's leave it empty as we're going to do live log streaming soon which will help customers to investigate such grey failures on their side. If we didn't have this feature coming up soon I'd put something like
"function is unhealthy".

eabatalov (approved):
:)

PR #1675, opened 2025-08-08
Process a task alloc outcome only if the task is running the alloc

eabatalov:
We have two paths where we handle task alloc outcomes:
* Task Alloc output ingestion
* FE termination

Depending on exact cause Executor can send notifications to Server to activate both paths per alloc or only one path. E.g. when Executor gets shutdown only FE termination path gets activates on Server. Another example is when a task alloc function timesout. In this case Executor sends task outcome to Server but also send FE termination update with the alloc id being blamed for FE termination.

The current idempotency check "if Task Status is Running" to not process an alloc for the second time on one of the paths fails with the following scenario:

1. First alloc (attempt 0) causes FE termination by timing out in the alloc function.
2. Server gets ingest alloc (attempt 0) output notification.
3. Server creates new alloc (attempt 1) for the same FE because it doesn't know yet that it's terminated. The alloc Task status is set to Running.
4. Server recieves notification that the FE was terminated (due to the first alloc attempt 0) and sees that the task is in Running state. Server creates a new alloc because of this and increments retry counter. Now there are two allocs for the same task and the state machine is somewhat inconsistent for the task. E.g. some allocs end up with "Undefined" state but running the task doesn't fail so far.

To fix this the idempotency is ensured by doing the following sequence on each alloc outcome processing path:

1. Check if the task is running the alloc that we're processing.
2. Once alloc is processed set running alloc id in the task to None.

This ensures that we won't process alloc outcome on either of the paths more than once even when we get an outcome for alloc that is not the current alloc that the task is running.

To implement this task's Running status now includes id of the running allocation. And then when we set the task status to pending or completed this effectively sets the running allocation id to None. Storing alloc id in Task looks okay because we persist both Task and Allocation objects in state store.

This breaks state store backward compatibility but it's okay as with the upcoming S3 direct changes state store will need to get wiped anyway so we're free to implement this fix in any way we see fit. There might be a better way to implement this, please let me know.

I also updated some stale comments and refactored existing code a bit on the code paths that I touched.

This fix is not surfacing in tests right now. The test that I uncommented in this PR is passing without this fix. But with S3 direct changes the timings/ordering of events change and this test starts failing consistently.

### DIFF
--- a/indexify/src/indexify/proto/executor_api.proto
+++ b/indexify/src/indexify/proto/executor_api.proto
@@ -240,9 +240,10 @@ enum TaskOutcomeCode {
 enum TaskFailureReason {
     TASK_FAILURE_REASON_UNKNOWN = 0;
     // Internal error on Executor aka platform error.
-    // Includes grey failures when we can't determine the exact cause.
     TASK_FAILURE_REASON_INTERNAL_ERROR = 1;
     // Clear function code failure typically by raising an exception from the function code.
+    // Also a grey failure where we can't determine the exact cause. We attribute these to
+    // functions to prevent service abuse but not billed intenionally failing functions.
     TASK_FAILURE_REASON_FUNCTION_ERROR = 2;
     // Function code run time exceeded its configured timeout.
     TASK_FAILURE_REASON_FUNCTION_TIMEOUT = 3;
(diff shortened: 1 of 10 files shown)


diptanu (approved):
Looks good.

PR #1707, opened 2025-08-19
Move processed_state_changes to specific payloads.

calavera:
## Context

Most of the time, the `processed_state_changes` vector is empty. By moving it to the variants that can actually receive processed changes, we make the cases where we handle those changes more explicit.

## What

Remove `processed_state_changes` from the `StateMachineUpdateRequest`, and put the vector in the variants that can actually be accompanied by already processed state changes.

## Testing

All unit and integration tests should still pass.

## Contribution Checklist


diptanu:
This makes sense, the only component which handles state changes are the scheduler.

PR #1709, opened 2025-08-19
Fix linting issues.

calavera:
## Context

This PR address the accumulated linting issues in the server's codebase. It also adds a step in the CI workflow to check linting issues before marking the job as successful.

## What

Besides addressing a bunch of minor linting issues in the codebase, there are a couple of big and interesting changes that I'd like to bring to your attention:

1. Usage of `From<T> for RequestPayload` makes it easier to centralize how `RequestPayload` objects are constructed. If there are any changes in the definitions of any variants, there should only be 1 place that you need to update to make it work again.
2. Removal of double heap allocation for structs inside `Vec` and `HashMap` in `in_memory_state.rs`. There is no need to box structs when you put them inside `Vec` or `HashMap` because they will be automatically allocated in the heap for you. Removing the explicit boxing makes the code a little more efficient and also easier to work with. 

## Testing

All tests and lints should pass.

## Contribution Checklist

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -21,11 +21,26 @@ jobs:
     name: Lint Indexify Server
     runs-on: ubuntu-latest
     steps:
+      - name: Install protobuf package
+        run: sudo apt update && sudo apt install -y protobuf-compiler
       - uses: actions/checkout@v4
-      - name: Lint indexify-server
+      - name: Copy rust-toolchain
+        run: cp server/rust-toolchain.toml .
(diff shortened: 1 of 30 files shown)


.github/workflows/tests.yaml
```diff
@@ -41,19 +41,19 @@ jobs:
         with:
           cache-directories: |
             target
-      - name: Setup Rust
```
calavera, line 44: this `Setup Rust` is duplicated, see the step above.

PR #1717, opened 2025-08-21
sse stream fixes

diptanu:
Looks like we were not sending the request id in SSE streams. Added it back and simplified the code for streaming


diptanu:
This tests will not work until a new release of tensorlake is out with the SDK changes.

Issue #1711, opened 2025-08-20
Improve concurrency control in state store


diptanu:
In a lot of places in the scheduler and other places of the code we read objects from database, and then update and write them back. We are using RocksDB concurrency controls to make sure only one writer wins. But often writes are sequential to the state store but the updates don't take into account of the update that just happened. 

An example - Scheduler can make a Task go back to Pending when a container to which the task was allocated fails. Meanwhile the executor can also report the allocation failing for whatever reason, and then the code path that handles the allocation failure would also try to put the task back to pending, and along the way create a new allocation. 

We are checking state of the task in many places to prevent bad updates. The right way to solve this is to attach a vector clock to every domain object - like Task, allocation, Compute Graph, etc. When the state store, receives an update to a domain object it needs to check whether the clock of the updated object matches what's in the database. If the clock of what's stored is greater than the new update sent to the state store, that means the copy updated was not the latest. The state store should throw a retriable exception to the component writing the update, and ask it to redo its calculations.

The clock needs to be kept updated by the state store, and it should be the sole authority to update it. In replicated state machines, the underlying raft subsystem usually generates the clock. In our case we can have an atomic long managed by the state store. (we have something like this already).

PR #1724, opened 2025-08-22
Add vector clocks to the structs that are stored in the state store.

calavera:
## Context

This is a precondition to address #1711. Instead of making all the changes at once. I thought it'd be better to introduce vector clocks first, and then start using them as necessary.

## What

Implement a simple vector clock using AtomicU64 and add it to the structures that are serialized into the state store.

The clock is not used for anything at the moment, allowing us to introduce CAS semantics gradually over futures changes.

## Testing

I'm adding tests for the new builders, and updating existent tests.

## Contribution Checklist

### DIFF
--- a/server/src/data_model/mod.rs
+++ b/server/src/data_model/mod.rs
@@ -1,3 +1,4 @@
+pub mod clocks;
 pub mod filter;
 pub mod test_objects;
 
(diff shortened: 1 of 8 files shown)


miguelhrocha (approved):

server/src/data_model/mod.rs
```diff
@@ -443,18 +461,19 @@ impl ComputeGraph {
...
+            .created_at(self.created_at)
+            .version(self.version.clone())
+            .code(self.code.clone())
+            .start_fn(self.start_fn.clone())
+            .nodes(self.nodes.clone())
+            .edges(self.edges.clone())
+            .runtime_information(self.runtime_information.clone())
+            .state(self.state.clone())
+            .build()
+            .expect("ComputeGraphVersionBuilder should build successfully")
```
diptanu, line 476: @calavera We don't use expect or unwrap in non-test code. In some cases during startup we might be using unwraps. The thought is that Its better to surface up error messages to the APIs or in logs than getting the server to crash once the server has started up and can handle requests.

PR #1727, opened 2025-08-26
Simplify builders

calavera:
## Context

I noticed that the data models don't make use of the derive builder conventions to generate structs. Since it gives you some nice ergonomics, I think it's better to work with the library, instead of against it.

## What

This change removes the usage of `#[builder(build_fn(skip))]` to take full advantage of the derive builder crate.

I've put individual changes in different commits, so it's easier to review.

## Testing

Tests have been updated to ensure the structures are initialized correctly.

## Contribution Checklist


diptanu (approved):
LGTM. Thanks, I didn't know that you can have a method to populate a field. I wrote those build methods because I wanted to control the id generation logic.

PR #1728, opened 2025-08-27
Data Applications Serverside Impl

diptanu:
Updated proto to support dynamic expansion of graphs. 

Removed task_cache as we need to re-write it to support persistence of cache keys in database. It can be simplified as well. We will bring it back after this PR lands. I don't want to re-write that in this PR. 

Removed the old unversioned routes 

Removed the logging endpoints since we are going to start using telemetry service internally. 

---

> [!NOTE]
> Introduce v3 executor API with dynamic execution plan and request model, replace tasks with function runs, update state/store and routes to applications, and remove legacy task cache/logging/routes.
> 
> - **Protocol/API**:
>   - Add v3 `ExecutorAPI` (allocation/result v3, `ExecutionPlanUpdates`, `FunctionCall`, `ReduceOp`); extend `DataPayload` (id, content_type, metadata, RAW).
>   - Rename task/invocation enums to allocation/request enums; add `invoke_function` streaming RPC.
> - **Execution Controller**:
>   - Replace task-centric flow with allocation/request model; create function runs, propagate outputs, handle request errors; remove task cache.
>   - Update metrics and naming (allocation vs task; request vs invocation).
> - **State Store & In‑Memory**:
>   - Remove `Tasks/FnOutputs/ReductionTasks`; add function runs and request contexts; new keys and GC handling; simplify migrations.
> - **HTTP/Routing**:
>   - Rename compute graph to `applications`; new v1 routes for requests/function runs; remove old unversioned and logging endpoints.
> - **Downloads**:
>   - Stream request/function outputs via new paths; honor `DataPayload` metadata offsets.
> - **Executors/Scheduling**:
>   - FE selection/creation works with function runs; desired state carries function run allocations.
> - **Tests/Deps**:
>   - Update tests to new request/function-run model; bump deps and clean up unused code.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a6de3a0943cb9685072f4f199c6008bfccc4367e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


server/proto/executor_api_v3.proto
```diff
@@ -0,0 +1,347 @@
...
+    // This allows resolving datapayloads of resolved functions calls when it's upstream is called again.
+    // The ID is unique within the scope of the request.
+    optional string id = 1;
+    optional FunctionRef target = 2;
+    repeated FunctionArg args = 3;
+
+    // This metadata will contain opaque metadata which will allow the SDK 
+    // to connect the datapayloads of resolved functions calls when it's finally called again.
+    // The SDK can store additional opaque metadata for it's own use.
+    optional bytes call_metadata = 5;
```
eabatalov, line 292: Let's change this to DataPayload so SDK can put as much information needed to implement certain feature also the amount of metadata stored depends on the size of user generated data, like e.g. number of futures in a list which a function call argument.

server/proto/executor_api_v3.proto
```diff
@@ -0,0 +1,349 @@
...
+    optional string uri = 1;
+    optional DataPayloadEncoding encoding = 2;
+    // Not set and ignored by Server right now.
+    optional uint64 encoding_version = 3;
+
+    optional string content_type = 10;
+
+    // size of the metadata (optional)
+    // DataPayload metadata is stored at the head of the binary 
+    // blob. This metadata is used optionally to store additional
```
eabatalov, line 30: At the head of the Data Payload. Because many data payloads can be stored per BLOB. We don't want to limit one data payload per BLOB in the protocol to keep it flexible.

calavera (commented):
I left you some minor comments. This is looking really good.

PR #1744, opened 2025-09-02
Replace the direct usage of TransactionDB.

calavera:
## Context

This PR continues abstracting the DB logic behind the Driver interface.

## What

Instead of passing TransactionDB objects, we're passing RocksDBDriver. This doesn't change much what we have, but it allow us to start putting more logic behing the driver trait.

Repace all usage of `cf_db` with `column_family` to be more explicit of what it is, and also remove the call from within the enum implementation. This makes the code more consistent across the board.

Stop calling `column_family` from within iterators. That call uses a RwLock to access column families. If the lock is held for writing, all read operation wait for it. Instead of calling it for every iteration within loops, we can call it before looping, reducing the lock contention in specific areas.

## Testing

Added new tests to verify more behavior changes.

## Contribution Checklist

### DIFF
--- a/server/Cargo.toml
+++ b/server/Cargo.toml
@@ -27,7 +27,7 @@ nanoid = "0.4.0"
 object_store = { version="0.12.3", features = ["aws"] }
 uuid = { version = "1.18.0", features = ["v4"] }
 tower-http = { version = "0.6.6", default-features = false, features = ["cors", "trace"] }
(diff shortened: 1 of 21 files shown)


diptanu (approved):
Feels like this is moving things around mostly and putting them behind some abstractions? LGTM

PR #1768, opened 2025-10-04
feat: app navigation update

ilbum:
## Context

Because different namespaces can have different application names, resetting the routing to the application level will ensure proper namespace switches.

## What

When the user switches application namespace it doesn't try to maintain the same application in the URL, to prevent it looking for requests that do not exist.

## Testing

Ran locally while connecting to dev environment.

## Contribution Checklist

### DIFF
--- a/server/ui/src/routes/root.tsx
+++ b/server/ui/src/routes/root.tsx
@@ -23,6 +23,7 @@ import {
   Outlet,
   redirect,
   useLocation,
+  useNavigate,
   useParams,
 } from 'react-router-dom'
 import theme from '../theme'
@@ -53,12 +54,13 @@ export async function loader({ params }: LoaderFunctionArgs) {
 
 function Dashboard() {
   const location = useLocation()
+  const navigate = useNavigate()
   const { namespace = 'default' } = useParams<{ namespace: string }>()
   const { namespaces } = useLoaderData() as RootLoaderData
 
   const handleNamespaceChange = (event: SelectChangeEvent) => {
     const newNamespace = event.target.value
-    window.location.href = window.location.origin + `/ui/${newNamespace}`
+    navigate(`/${newNamespace}`)
   }
 
   const navItems: NavItem[] = [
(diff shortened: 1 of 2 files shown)


server/ui/src/routes/root.tsx
```diff
@@ -58,8 +58,7 @@ function Dashboard() {
 
   const handleNamespaceChange = (event: SelectChangeEvent) => {
     const newNamespace = event.target.value
-    const newPath = location.pathname.replace(/^\/[^/]+/, `/${newNamespace}`)
-    window.location.href = window.location.origin + '/ui' + newPath
+    window.location.href = window.location.origin + `/ui/${newNamespace}`
```
RaeesBhatti, line 61: We could use `navigate` to keep the SPA in memory instead of reload the page but the current state will also function correctly.
```diff
diff --git a/server/ui/src/routes/root.tsx b/server/ui/src/routes/root.tsx
index 69f54862..73131940 100644
--- a/server/ui/src/routes/root.tsx
+++ b/server/ui/src/routes/root.tsx
@@ -23,6 +23,7 @@ import {
   Outlet,
   redirect,
   useLocation,
+  useNavigate,
   useParams,
 } from 'react-router-dom'
 import theme from '../theme'
@@ -53,12 +54,13 @@ export async function loader({ params }: LoaderFunctionArgs) {
 
 function Dashboard() {
   const location = useLocation()
+  const navigate = useNavigate()
   const { namespace = 'default' } = useParams<{ namespace: string }>()
   const { namespaces } = useLoaderData() as RootLoaderData
 
   const handleNamespaceChange = (event: SelectChangeEvent) => {
     const newNamespace = event.target.value
-    window.location.href = window.location.origin + `/ui/${newNamespace}`
+    navigate(`/ui/${newNamespace}`)
   }
 
   const navItems: NavItem[] = [
```

RaeesBhatti (approved):
I suggested a slightly different approach but this works too. Approved!

PR #1769, opened 2025-10-06
Blocking Function Calls Server Impl 

diptanu:
Added support for blocking function calls in the server. This is not ready to be merged but it's ready to be reviewed - protocol changes and the implementation. The implementation is not 100% complete yet - 
1. We need to make the function calls idempotent i.e not start a new request if the input parameters and source function run id is same.
2. Cancel previous runs if something about the invariants changes 
3. Cancel runs if the parents are cancelled 

Assumption on how the data plane state machine will work - 
1. User code calls a tensorlake function, the FE blocks and sends the request to the executor. The executor creates data playloads, and fires up a invoke_function RPC call to the server with the datapayloads and the name of the function.
2. Server starts the request and returns a request id
3. Executor needs to fire a report executor state API call with the new request ID in the new `request_watches` attribute. 
4. The server tracks these watch requests, and if the request ID changes state then it will send the state and output of these requests back to the executor in the desired executor state message. This helps us to reduce some work on the dataplane to not run a new loop for managing watches 
5. The executor needs to include these request watch ids in the executor hash. If the hashes don't change server will not update it's internal state for executors. 

This design adds very little complexity in the state machine of the server and on the executor side there is no extra loop for managing watches - we can reuse most of the existing infrastructure for getting updates from server. The major change is going to be in FE and in the executor firing up these requests and also unblocking the FEs once the requests are completed.


server/proto/executor_api.proto
```diff
@@ -340,18 +342,35 @@ message FunctionCallRequest {
...
+message FunctionCallResponse {
+    optional string request_id = 1;
 }
 
-message FunctionCallResponse {
-    oneof event {
-        string update = 1;
-        FunctionCallResult result = 2;
-    }
+enum RequestFailureReason {
```
eabatalov, line 349: Can we avoid adding another enum with failure reasons? We have one for FE init and one for alloc failures already.

server/proto/executor_api.proto
```diff
@@ -340,18 +342,35 @@ message FunctionCallRequest {
...
+    REQUEST_FAILURE_REASON_UNKNOWN = 0;
+    REQUEST_FAILURE_REASON_INTERNAL_ERROR = 1;
+    REQUEST_FAILURE_REASON_FUNCTION_ERROR = 2;
+    REQUEST_FAILURE_REASON_REQUEST_ERROR = 3;
+    REQUEST_FAILURE_REASON_NEXT_FUNCTION_NOT_FOUND = 4;
+    REQUEST_FAILURE_REASON_CONSTRAINT_UNSATISFIABLE = 5;
+}
+
+enum RequestOutcome {
```
eabatalov, line 358: We're not starting a new request (with a new ID and empty request state and etc) with a blocking function call, because of this we should not use "Request" term for something other then a Request.

Also same question about outcomes here. We have 2 enum with outcomes already.

